PR #2, opened 2021-04-18
Enforcing type check

MrKuzio:
I've done it with regex across the whole repo, and I changed the `==` into `===`. 
This change will shut up my linter warnings.


cyruzzo:
I think that changing == to ===  it's not something that you can do without thinking of consequences.

I'm pretty sure that I have pieces of codes where I consider 1 and "1" to be the same thing... If we now  switch to "type safe comparison" suddendly those values are different.

I would accept the other "minor commits" but since it's a single PR I cannot pull it as it is.

PR #3, opened 2021-04-18
continuing beta styling

Palad1NPT:
- Top Menu styles and animations.
- Scene button to middle of the accordion to fix problems on smaller screens.
- Setting Top Menus on top of combat tracker.
- Combat tracker bottom buttons adjust to parent size.
- Send to players button.

WARNING
Minor changes to the js files, basically some classes added and some inline styles removed to avoid excess usage of important.

### DIFF
--- a/abovevtt.css
+++ b/abovevtt.css
@@ -11,6 +11,9 @@ body > button,
 #mega_importer button,
 #prewiz td button,
 #wizard_popup button,
+.top_menu button,
+.monster_frame > button,
+.stp-btn,
 .sidebar__controls button {
     background: #3d3d3d;
     color: #fff;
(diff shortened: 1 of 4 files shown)


abovevtt.css
```diff
@@ -52,6 +68,9 @@ body > button:hover,
     line-height: 19px;
 }
 
+button#scene_selector_toggle {
```
MrKuzio, line 71: The type `button` doesn't increase the specificity of the selector. 
It might be redundant in this case.

Palad1NPT: It is, it was copied from autogenerated selector by Chrome DevTools.
Not a blocker tho', so I might change it on a future PR.

PR #11, opened 2021-04-20
Css refactoring/combat area

MrKuzio:
A few things are happening in this PR.

1. Removed declaration for `nextid` because it's never used
2. Moved the css for the `lightgreen` background colour into the css file
3. The css selector is more specific: now the `lightgreen` background is only applied when the attribute `data-current` is set. So I removed the js code that was overriding the background colour to `''` as it is now redundant.
4. Added the `alt=""` attribute to the img element. It is required for accessibility.


MrKuzio:
Closed PR and opened with rebase from new version.

PR #13, opened 2021-04-20
hp aura and dead token X

Palad1NPT:
hp aura and dead token X

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -166,6 +166,22 @@ class Token {
 
 			this.options.hp = old.find(".hp").val();
 			this.options.max_hp = old.find(".max_hp").val();
+
+			// HEALTH AURA
+			if (this.options.max_hp > 0) {
+
+				if (this.options.max_hp > 0 && parseInt(this.options.hp) === 0) {
+					old.prepend(`<div class="dead" style="--size: ${parseInt(this.options.size) / 10}px;"></div>`);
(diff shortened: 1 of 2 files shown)


Token.js
```diff
@@ -166,6 +166,22 @@ class Token {
 
 			this.options.hp = old.find(".hp").val();
 			this.options.max_hp = old.find(".max_hp").val();
+
+			// HEALTH AURA
+			if (this.options.max_hp > 0) {
+
+				if (this.options.max_hp > 0 && parseInt(this.options.hp) === 0) {
```
MrKuzio, line 173: `this.options.max_hp > 0` is already determined on line 171.
I think we can remove it from line 173.  :)

Token.js
```diff
@@ -166,6 +166,22 @@ class Token {
 
 			this.options.hp = old.find(".hp").val();
 			this.options.max_hp = old.find(".max_hp").val();
+
+			// HEALTH AURA
+			if (this.options.max_hp > 0) {
+
+				if (this.options.max_hp > 0 && parseInt(this.options.hp) === 0) {
+					old.prepend(`<div class="dead" style="--size: ${parseInt(this.options.size) / 10}px;"></div>`);
```
MrKuzio, line 174: Good one with the CSS custom properties! :D

PR #16, opened 2021-04-21
player tab restyle & add party button

Palad1NPT:
Fix Free Maps section buttons hovering color
Players tab styling
Add Party button


cyruzzo (changes requested):
Just change the check inside token_button

Token.js
```diff
@@ -791,6 +791,11 @@ function token_button(e) {
 		options.color = $(e.target).attr('data-color');
 	}
 
+	if (tokenIndex && tokenTotal) {
```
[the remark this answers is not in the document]
Palad1NPT: you're absolutely right, old habits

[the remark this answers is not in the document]
Palad1NPT: changed!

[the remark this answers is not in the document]
nickinnon: Another approach:
`tokenIndex.length` and make sure `tokenIndex` is instantiated as an empty array so it is never null or undefined.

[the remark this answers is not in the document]
nickinnon: So if you have a single item in index 0, the length will be 1 -> truthy

Main.js
```diff
@@ -591,14 +591,27 @@ function open_player_sheet(sheet_url) {
...
+
 
 			var playerdata = {
 				id: sheet_url,
 				hp: current_hp,
 				max_hp: max_hp,
 				ac: $(event.target).contents().find(".ddbc-armor-class-box__value").html(),
 				pp: pp.html(),
-				conditions: conditions
+				conditions: conditions,
```
nickinnon, line 611: NIT
Can change this to just `conditions` as you did with `abilities` below to keep it consistent.

PlayerPanel.js
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
 
 
 	});
-
-
-
+	const addPartyButtonContainer = $("<div class='add-party-container'></div>");
+	const addPartyButton = $("<button id='add-party'>ADD PARTY</button>");
+	addPartyButton.on('click', () => {
+		window.pcs.forEach(function (p, i) {
```
nickinnon, line 40: NIT

May be worth givng `p` a better var name to reduce cognitive load. I.E. Is this player, party? Don't make us think.

PlayerPanel.js
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
...
 
 	});
-
-
-
+	const addPartyButtonContainer = $("<div class='add-party-container'></div>");
+	const addPartyButton = $("<button id='add-party'>ADD PARTY</button>");
+	addPartyButton.on('click', () => {
+		window.pcs.forEach(function (p, i) {
+			token_button({ target: $(`[data-name='${p.name}']`) }, i, window.pcs.length);
```
nickinnon, line 41: Couple gotcha issues here.

What happens if two  items of the same name are in the List? E.G. Players Bob and Bob? Looks like the selector pulls from the name, so we would have a duplicate here. I don't know what the behavior is. To solve this, we should add a key here (assuming this doesn't impact the UI) so the selection is unique. We can do that by appending the index to the name, or preferably using another UUID (if something like that exists on this model).

Additionally, what happens when an upstream bug causes name to be undefined?

There is a trick here to avoid a nasty `undefined` or error if `p` or `p.name` is undefined.

```
window.pcs.forEach(function ({ name = '' }, i) {
    token_button({ target: $(`[data-name='${name}']`) }, i, window.pcs.length);
    // . . .
 ```

File: Token.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (Palad1NPT) [reply Token.js]: tokenIndex is numeric not an array, this has been fixed anyway

File: PlayerPanel.js [hunk 5]
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
...
-
-		row_inside.append(stats);
-
-		conditions = $("<div style='float:left;'/>");
-		conditions.append("<div><b>Conditions</b></div>");
-		conditions.append("<div class='conditions'></div>");
-
-		row_inside.append(conditions);
 
+		let playerData;
```
nickinnon, line 54: Initialize this to an empty variably of the desired primitive. 

Personal preference. I'm not the maintainer, but there isn't a style guide.

PlayerPanel.js
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
...
-			stats.find(".hp").text(data.hp + "/" + data.max_hp);
-			stats.find(".ac").text(data.ac);
-			stats.find(".pp").text(data.pp);
-
-			for (var i = 0; i < data.conditions.length; i++)
-				conditions.find(".conditions").append($("<div/>").text(data.conditions[i]));
+			playerData = window.PLAYER_STATS[pc.sheet];
 		}
+		
+		newPlayerTemplate = `
```
nickinnon, line 59: Is this defined somewhere? This is a global. Use let/const if that isn't intended.

PlayerPanel.js
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
...
+				</div>
+				<div class="player-card-content">
+					<div class="player-token" style="box-shadow: ${
+						playerData ? `${token_health_aura(
+							Math.round((playerData.hp / playerData.max_hp) * 100)
+						)} 0px 0px 11px 3px` : 'none'
+					};">
+						<img width="70" height="70" src="${pc.image}" style="border: 2px solid ${color}" />
+						${
+							playerData ? `
```
nickinnon, line 83: Another way is to short circut this with just `&&` so you don't have an empty ternary. This is a style thing, so don't worry about it. Just bringing it up as a common practice thing

Main.js
Palad1NPT: yep you're right, ahahah but I didn't read it only added my properties.

PlayerPanel.js
```diff
@@ -34,87 +34,121 @@ function update_pclist() {
...
-			stats.find(".ac").text(data.ac);
-			stats.find(".pp").text(data.pp);
-
-			for (var i = 0; i < data.conditions.length; i++)
-				conditions.find(".conditions").append($("<div/>").text(data.conditions[i]));
+			playerData = window.PLAYER_STATS[pc.sheet];
 		}
+		
+		newPlayerTemplate = `
+			<div class="player-card">
```
nickinnon, line 60: Just in general here, a lot of these string literal expressions, you'd want to assign to a temp variable, above building the string, then place the just the var name into the HTML.

This will clean things up a bit and separate the data/UI layers.

Another advantage would be that using string literals like this actually saves the whitespace, which isn't something we necessarily want en client. I forget if this keeps the indentation without context that we're already multiple columns into the file.

Token.js
```diff
@@ -717,7 +681,7 @@ class Token {
 }
 
 
-function token_button(e) {
+function token_button(e, tokenIndex = null, tokenTotal = null) {
```
nickinnon, line 684: I would set these to empty arrays, then check the length on the array rather than expecting null as a valid primitive. Usually leads to issues/bugs.

nickinnon (commented):
Some notes. Not a maintainer. TIOLI

Token.js
nickinnon: Correction, since these are Number prims of an index, I would suggest setting them to 0 then using Number.isNumber(var). 

Good rule of thumb is never accept null or mix your privatives loosely typed languages.

Palad1NPT: that would invalidate the reason why we use it (create a circular placement of tokens when using the add party button), 0 is a valid index and in this case a valid total.

PlayerPanel.js
MrKuzio: It would be great to build the UI with Vue or React, for this template management (my choice would be Vue).
By doing so we will keep the presentation and the logic separate, following the MVC principles.

But I am not the maintainer and I understand this would add another layer of complexity to a so far simple tool. 
It might require a better discussion than comments on a PR, imho.

SceneData.js
```diff
@@ -5298,3 +5298,87 @@ var PRESET = {
...
+        "Attack rolls against the creature have advantage."
+    ],
+    "Unconscious": [
+        "An unconscious creature is incapacitated (see the condition), can’t move or speak, and is unaware of its surroundings",
+        "The creature drops whatever it’s holding and falls prone.",
+        "The creature automatically fails Strength and Dexterity saving throws.",
+        "Attack rolls against the creature have advantage.",
+        "Any attack that hits the creature is a critical hit if the attacker is within 5 feet of the creature."
+    ]
+}
```
MrKuzio, line 5384: I am a bit worried about hard-coding these descriptions in our codebase. 
It would be better if we pulled from an official source, even if it's a one-off async request and we store it globally. But I think it would allow us to avoid having obsolete description in case of errata or updates in the future.

Token.js
```diff
@@ -1074,3 +1043,39 @@ function deselect_all_tokens() {
...
+			return r + r + g + g + b + b;
+		});
+
+		const pHex = (n) => parseInt(n, 16);
+
+		var result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+		return result ? `rgb(${pHex(result[1])} ${pHex(result[2])} ${pHex(result[3])} / 80%)` : null;
+	}
+	return hexToRGB(percentToHEX(hpPercentage));
+}
```
MrKuzio, line 1081: QUESTION

Why are we converting HEX to RGB? 
I css both are accepted values `rgb(255, 255, 255)` === `#ffffff`.

Or am I missing something? 🤔 

And if you need colours in percentage we could even use the HSL colour notation: 
- H hue
- S Saturation
- L Lightness
https://www.w3schools.com/css/css_colors_hsl.asp

Palad1NPT: wouldn't we need also a function to convert it from HEX to HSL?

MrKuzio: HEX, HSL, and RGB are all valid color declarations in css. 
So, why do we need to convert them in one way or another?

MrKuzio: Search for `CSS Color Values` in the page: https://www.w3schools.com/csS/css_colors.asp

PlayerPanel.js
Palad1NPT: good catch, didn't thought of that use case, fixed.

PlayerPanel.js
Palad1NPT: fixed, thanks!

SceneData.js
nickinnon: Suggestion:

May be out of scope for this PR. We should be able to pull directly from the combat conditions apendix and use that raw. 

https://www.dndbeyond.com/sources/basic-rules/appendix-a-conditions

Palad1NPT: the character sheet has the conditions descriptions, but it needs a button click per condition to reveal the data.

Token.js
Palad1NPT: because I need to set opacity on the color

SceneData.js
cyruzzo: Hardcoding is ok from a tech point of view.. I don't think they'll change often (if ever)... What I'm not sure about is if it's legally ok to copy that data inside the plugin .

Palad1NPT: conditions are included in the SRD and are licensed under OGL

PlayerPanel.js
```diff
@@ -34,87 +34,136 @@ function update_pclist() {
...
+							</div>
+						` : `
+							<div class="player-no-attributes">
+								Please load the character sheet.
+							</div>
+						`
+					}
+				</div>
+				${
+					playerData ? `
```
cyruzzo, line 125: In 0.0.41 I fixed this by changing to "playerData && playerData.abilities".
I backported that to main. Base your branch on main and I'll pull.

Palad1NPT: done!

PR #21, opened 2021-04-22
backporting fixes

cyruzzo:


CombatTracker.js
```diff
@@ -1,15 +1,16 @@
...
 	toggle.click(function(){
 		if($("#combat_tracker_inside").is(":visible")){
 			$("#combat_tracker_inside").hide();
-			$("#combat_tracker").css("height","20px");
+			$("#combat_tracker").css("height","20px"); // IMPORTANT
 		}
 		else{
 			$("#combat_tracker_inside").show();
-			$("#combat_tracker").css("height","450px");
+			$("#combat_tracker").css("height","450px"); // IMPORTANT
```
Palad1NPT, line 13: we should really start using classes and avoid inline css

Palad1NPT (approved):

CombatTracker.js
cyruzzo: I accept all css changes and re-organizations as long as they don't break things

PR #24, opened 2021-04-23
DRAW/FOG Polygon

Palad1NPT:


cyruzzo:
On the DM view. When you do a "reveal polygon" ther resulting area is not fully cleared. There's still some opacity. On player view that dosn't seem to be an issue.

PR #25, opened 2021-04-24
[bugfix] bugfix different version data and select/measure layer problem

Palad1NPT:
[bugfix] bugfix different version data and select/measure layer problem

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -868,7 +868,11 @@ function setup_draw_buttons() {
 
 
 		var target = $("#fog_overlay");
-		target.css("z-index", "50");
+		if (!e.currentTarget.id || (e.currentTarget.id !== "measure-button" && e.currentTarget.id !== "select-button")) {
+			target.css("z-index", "50");
+		} else {
+			target.css("z-index", "20");
+		}
 
 		if ($(e.target).attr('id') == "measure-button") {
 			target = $("#VTT");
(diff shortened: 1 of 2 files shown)


Fog.js
```diff
@@ -868,7 +868,11 @@ function setup_draw_buttons() {
 
 
 		var target = $("#fog_overlay");
-		target.css("z-index", "50");
+		if (!e.currentTarget.id || (e.currentTarget.id !== "measure-button" && e.currentTarget.id !== "select-button")) {
+			target.css("z-index", "50");
+		} else {
+			target.css("z-index", "20");
+		}
```
MrKuzio, line 875: What are you trying to do here? 🤔 
Can you run me through the logic, please?

Palad1NPT: yes, to be able to draw over tokens, I set z-indexes to bring the drawing layer on top of the tokens layer.
for the measure and select this doesn't apply as these tools interact directly with the tokens.

cyruzzo: I think it's better that we definitely change the behaviour for "measure" to stop it interacting with tokens.
So this "exception" for change of z-index should only be for "select".

I think in the future we should create some ui to display token movement directly inside the draggable options of the token.. so everytime you move a token you alway get that measurament without having to go to the measure tool.

Palad1NPT: cool, going to make it only for the select then!

Palad1NPT: done

PR #37, opened 2021-05-01
Hidden Monster First-Time-Tip

SnailDice:
Hidden monster Tip that appears on first use only by adding a file to local storage when clicking x, that prevents it from showing next time.


Palad1NPT:
I think this should be a task in itself
We should make this for all the features on a reusable way using jquery dialog, and also a way for people to turn it off in case they're data is cleared or wiped.

PR #43, opened 2021-05-04
SnailDice new scrapped maps

Palad1NPT:
SnailDice new scrapped maps


cyruzzo:
uhmm,.. I didn't test those. But it seem to me that the thumbs are not thumbs but full image.

Palad1NPT:
correct

PR #45, opened 2021-05-05
feature: change ADDSCENE to + and make it always on top, open scenes …

roi007leaf:
…on press


Palad1NPT:
Hi @roi007leaf welcome to the contributors :) so... a couple of things...

First, you have a merge conflict that you need to solve.
Also, you've misunderstood the positioning of the button. It should stay at the bottom right of the scenes panel.

roi007leaf:
@Palad1NPT oh shoot.. it said scene page.. not scenes selector widget or something.. are you sure that this implementation isn't better in this case? instead of opening the scenes widget every time the dm could just press the floating button @cyruzzo  your thoughts?

cyruzzo:
This PR is causing issues with the Main.js file so I need to close it.

PR #47, opened 2021-05-05
Measure and Token Drag Cosmetic Enhancements

brookesi:
Made the waypoint lines thinner and slightly transparent, added markers for each waypoint segment and fixed a bug for maps with an offset where snapping to grid wasn't accurate.

NOTE: This needs further work so the line tracks the centre of the token properly similar to Roll20 but I just don't have time to fix that. Most of the work is done in Fog.js WaypointManager class and the mouse handlers in Token.js


brookesi:
Also note that this does not include the 'visible to all players' message broker implementation either

PR #71, opened 2021-06-18
added hover tooltip to display token names

Kudolpf:


Kudolpf:
Will include a name editor in the context menu before PR

PR #87, opened 2021-07-25
Add ability to offset player map

vlaminck:
# What
This is an investigation into the recent issues brought up in Discord about DM maps and player maps not being aligned. It adds new options to the "Manual Grid Data" menu to allow offsetting the player map.  

# How
I added `player_map_offset_x` and `player_map_offset_y` to the scene data. These are defaulted to 0, and are filled in from the new inputs that I added to the "Manual Grid Data" form. 

These offsets are then used in the `reset_canvas` function in `Fog.js`. What that does is offset the canvas layers that sit on top of the player map. When a DM hides/reveals fog or draws, everything works as expected, but because the canvas is offset from the map image, it all lines up properly.

# Concerns
This appears to fix the offset issue, but I haven't done extensive testing on every part of the app. It's possible that this could introduce bugs in other areas of the app. 

As you can see in the following screenshot, this allows the maps to be aligned, but there is still a problem with scale. Considering these images are poorly scanned, I don’t know if this is a problem worth solving. If there are other official maps that could benefit from a solution like this, we can continue exploring it, but if it’s just a few poorly scanned maps, I don’t know if this is worth exploring further. 

# Screenshot
screenshot

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -367,6 +367,27 @@ function circle_intersection(x0, y0, r0, x1, y1, r1) {
 }
 
 function reset_canvas() {
+
+	if (!window.DM) {
+		 // only offset the player map
+		let offsetX = "0px";
+		if (CURRENT_SCENE_DATA.player_map_offset_x != undefined) {
+			offsetX = CURRENT_SCENE_DATA.player_map_offset_x + "px";
+		}
+		let offsetY = "0px";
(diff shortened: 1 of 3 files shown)


vlaminck:
Closing this PR as it was mostly for investigation purposes, and ultimately, not a problem that needs to be solved with code.

PR #91, opened 2021-08-04
Allow rotation of selected tokens

vlaminck:
# What
This adds the ability to rotate selected tokens. Here's a breakdown of everything that was added:
1. When you select tokens using the select tool, a bounding box will be drawn around the selected tokens. On the top side of that box is an eye symbol that can be dragged around. 
2. When dragging the eye symbol, all selected tokens will rotate to face wherever you drag that eye symbol to.
3. Holding shift while dragging the eye symbol will rotate tokens at 45º angles.
4. Rotations happen in real time for the DM, and are sent to players when the DM releases the eye token. This is to help with performance.

# Limitations
This is a DM-first feature because it is only available while using the select tool. My understanding is that the select tool will be going away, and players will be able to select multiple tokens at that time as well. When that happens, this will become available to players. I can help with that transition if needed, but including rotation in that migration should be pretty straight forward.

# Demo 
https://i.imgur.com/cou10qv.gif

# Demo While Holding Shift
https://i.imgur.com/hSTwktm.gif

# Demo With Player
https://i.imgur.com/BHNg1nZ.gif

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -995,6 +995,7 @@ function drawing_mouseup(e) {
 
 		redraw_canvas();
 		console.log("READY");
+		draw_selected_token_bounding_box();
 	}
 	if (e.data.shape == "measure") {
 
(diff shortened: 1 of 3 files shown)


cyruzzo:
Seems great. I'll test this next week.

PR #107, opened 2021-09-03
Brush tool, line thickness

Jupi-Taru2:
Add brush tool, gets the mouse position every 75ms to keep the number of points fairly low while still providing good tracking.
Add line-width slider for draw tools, works with the brush, lines, and any 'border' shape.
Made functions for drawing each shape to cut down on repeated code in drawing_mousemove, redraw_drawings, redraw_canvas.
Changed drawing_mousemove so the shape looks like what it will be while drawing, rather than always being red and transparent.
("Filled" and "Border" shapes are 90% opacity while drawing).
Adding fog now creates a grey shape as it is being drawn. Revealing still creates a red shape. This makes it less likely for DMs to mix them up as they are using them.
Right clicking while drawing shapes/fog will now cancel the draw.
Middle-clicking will no longer end a drawing, so users can pan while they draw.

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -871,7 +871,7 @@ function drawing_mouseup(e) {
 	}
 	
 	// ignore if right mouse button for drawing or fog, cancel is done in drawing_contextmenu
-	if((window.DRAWFUNCTION == "draw" || window.DRAWFUNCTION == "1" || window.DRAWFUNCTION == "0") && e.button == 2)
+	if((window.DRAWFUNCTION == "draw" || window.DRAWFUNCTION == "1" || window.DRAWFUNCTION == "0") && e.which !== 1)
 	{
 		return;
 	}
--- a/Fog.js
+++ b/Fog.js
@@ -720,7 +720,7 @@ function drawing_mousedown(e) {
 			drawStroke,
 			lineWidth
 		);
-		drawClosingArea(window.BEGIN_MOUSEX[0], window.BEGIN_MOUSEY[0], !isNaN(window.DRAWFUNCTION));
+		drawClosingArea(ctx,window.BEGIN_MOUSEX[0], window.BEGIN_MOUSEY[0], !isNaN(window.DRAWFUNCTION));
 	} else {
 		window.BEGIN_MOUSEX = (e.pageX - 200) * (1.0 / window.ZOOM);
 		window.BEGIN_MOUSEY = (e.pageY - 200) * (1.0 / window.ZOOM);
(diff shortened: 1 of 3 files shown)


cyruzzo:
I still need to look and test this. My first thought is that the "brush tool" may write up too much data.
When the game starts the DM sends the full current scene data to the players through the gamelog  in a single message (with all fog and drawings..).. and we may reach an upper limit to that..

Jupi-Taru2:
Yeah I tried to keep the number of points down, it takes 3 at each end of the brush stroke and then while the mouse is moving it takes a point every 75ms. I found that unless it's a very large or complex drawing it doesn't take more than 50 points per drawing. It can be reduced by increasing how long it waits between polling (line 829), but it's tradeoff between size and fidelity. 100ms still works well, and 200ms is still mostly functional and really few points.

LimpingNinja:
I've been testing this on my side with players and mainly what I use it for is drawing routes, pointing arrows, circling points, etc. I've written a few messages and drew a few vulgar images as well and none of the players have had issue, including one who is unable to play in foundry due to an older system and bandwidth constraints.

cyruzzo:
Thanks for the additional feedback. I'll test this later in the cycle.
Bandwidth is not the issue. The only potential for a "fuck up" is if the Scene Data grows bigger that the message size supported by the DDB message broker. You probably wonder what's the maximum message size... and. uhmm.. I hadn't tested :)

LimpingNinja:
I'm not sure you would see anyone run into the limit in actual play (as opposed to tested), the message broker should essentially be only limited by the data stored in the passed object. JSON I think has a uvalue of about 4mb and javascript objects start to get wonky at 10x above that, but you could probably just set a length limit to the array lengths stored for various drawing types in the data. But just as a test example, two characters on my side and a remote player logged into two characters just to see if there was wonky behavior, I did this act of scribbling for 5 minutes and swapped back and forth between scenes over a couple logins and browser switches just to see what I could get to fall over. Eventually I got tired of scribbling and did not reach an upper limit and (interestingly) got no slowdown which I have seen before on other VTT drawing systems. If you want I can automate that and see where it falls over, but personally I think if someone does more scribbling than that on their scene they may have bigger problems going on ;) 

AsYouDo

cyruzzo:
ok, that screenshoot from a player view is pretty convincing. will merge before next beta.

Jupi-Taru2:
I actually did a stress test today and it did fail to send (granted it's a ridiculous amount of brush strokes). The issue is from the number of siginficant digits in the points leads to very large JSON data
(e.g. {"x":140.22252705380274,"y":91.44947416552353}). I shouldn't lose any perceptible accuracy if I round to 2 digits after the decimal, and it would significantly reduce the size of the JSON data.
image

cyruzzo:
oh yes, please round it :) I think that even rounding to integer will not loose many detail

Jupi-Taru2:
As you said, rounding to an integer didn't hurt accuracy at all, and now the JSON data is much smaller! I was easily able to send this mess of scribbles.
image

PR #118, opened 2021-09-14
Combat Tracker Text Bigger

Jupi-Taru2:
Made the text in the combat tracker bigger, it's pretty unreadable in large resolutions.


cyruzzo:
Did you test with monsters having 3 digits HP ? (I still didn't)

Jupi-Taru2:
It looks fine in Chrome
image

PR #127, opened 2021-09-20
fixing jitsi modal width overflow

hypergig:
Currently, if you open the jitsi modal (gallery mode) the right side runs way off the screen, in some cases (depending on your display size) pushing all the participants out of view. This change addresses that by setting the width of the modal relative to the space and not statically based on `window.width`.

## Before
Screen Shot 2021-09-20 at 9 54 03 AM

## After
Screen Shot 2021-09-20 at 9 56 25 AM

### DIFF
--- a/Jitsi.js
+++ b/Jitsi.js
@@ -78,7 +78,7 @@ function jitsi_tile_listener(event) {
 }
 
 function jitsi_modal() {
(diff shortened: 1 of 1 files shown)


hypergig:
folding into https://github.com/cyruzzo/AboveVTT/pull/128

PR #128, opened 2021-09-20
jitsi improvements

hypergig:
* added share audio feature (for sharing your computer's audio, a just a tab, etc)
* added share desktop feature (for sharing something quickly from your desktop, a tab, etc )
* added select background feature, for hiding the giant mess behind you
* moved deprecated `TOOLBAR_BUTTONS` options to `configOverwrite` object
* on video layout toggle, force the tiling state rather than toggling it
* fixed event listener workaround
* fixed bug where jitsi was being bottomed when not being asked to 
* general code cleanup
* fixed width overflow
  * width before
Screen Shot 2021-09-20 at 9 54 03 AM
  *  width after
Screen Shot 2021-09-20 at 9 56 25 AM

### DIFF
--- a/Jitsi.js
+++ b/Jitsi.js
@@ -67,21 +67,17 @@ function init_jitsi() {
 			VERTICAL_FILMSTRIP: false
 		},
 	};
-	api = new JitsiMeetExternalAPI(domain, options);
-	window.jitsiAPI = api;
+	window.jitsiAPI = new JitsiMeetExternalAPI(domain, options);
+	jitsi_bottom()
 
-	/* You can not execute any commands right after you instantiate the jitsi client
-	and there is really no way to wait for the client to be ready. A workaround is to
-	create a one time use handler just for startup, to render the pane and set the tile
-	layout, attached to the subjectChange event, which by chance is almost the last thing
-	the client does before being considered ready. After which, commands can be executed.
-	*/
-	api.addEventListener('subjectChange', jitsi_startup)
+	//You must wait for the local participant to join the conference before attempting to execute any commands against the api 
+	window.jitsiAPI.addEventListener('videoConferenceJoined', jitsi_startup)
 }
 
+// This only gets used once, the very first time after you start video
 function jitsi_startup() {
-	window.jitsiAPI.removeListener('subjectChange', jitsi_startup);
-	jitsi_bottom()
+	window.jitsiAPI.removeListener('videoConferenceJoined', jitsi_startup);
+	jitsi_bottom();
 }
 
 function jitsi_modal() {
--- a/Jitsi.js
+++ b/Jitsi.js
@@ -87,7 +87,7 @@ function jitsi_startup() {
 function jitsi_modal() {
 	$("#meet").css("position", "fixed").css("top", "100px").css("height", "80%").css("left", "50px").css("width", "75%");
 	window.tile_desired = true;
-	window.jitsiAPI.executeCommand(`setTileView`, false);
+	window.jitsiAPI.executeCommand(`setTileView`, true);
 	$("span", $("#jitsi_switch")).text("fullscreen_exit");
 	$("#jitsi_switch").attr("data-name","Exit fullscreen (v)");
 }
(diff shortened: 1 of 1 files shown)


cyruzzo:
Before this.. I was forcing normal view when jitsi is in the bottom.. and forcing tile view when it's magnified.
Right now it's not focicng it (and it may be ok).. but it's still toggling it everytime you go to magnified to bottom.

jordan-da:
> Before this.. I was forcing normal view when jitsi is in the bottom.. and forcing tile view when it's magnified.
> Right now it's not focicng it (and it may be ok).. but it's still toggling it everytime you go to magnified to bottom.

Ah ok. Yeah I was actually trying to figure what your preference was, I asked in discord but it got lost in other chatter.

Personally, I prefer the tile layout when in large modal view as well as bottom view. Maybe most especially bottom view, as when you are on speaker layout and on the bottom, you lose a great deal of real estate on the edges.

So my proposal is, if you agree, to use tile layout for both modal and bottom views. Which I think looks better all around. Also, the user can always switch the jitsi layout themselves in either view as their is a toggle tile layout button present in the iframe. So we are merely providing a suggestion.

And if you get any volume of complaints over this, we simply change it the default.

cyruzzo:
Uhmm.. Tile view at the bottom create very small images.
Jitsi is not really adapting to fit the space nicely.

But also the other layout sucked when used at the bottom. So we can try with both "tile view".

image

hypergig:
> Uhmm.. Tile view at the bottom create very small images.
> Jitsi is not really adapting to fit the space nicely.
> 
> But also the other layout sucked when used at the bottom. So we can try with both "tile view".
> 
> image

oy.. that is very very odd.. that is not the way it renders for me.. i get them all side by side.. let me look into it :/

jordan-da:
Ok this was my fault, I never tested with more than two participants. Clearly that looks awful. I wish they distributed the tiles across all available real estate, instead of trying to jam them like that.

I will disable tile layout when in bottom. My apologies.

hypergig:
@cyruzzo fixed the default layouts for both views

hypergig:
> But also the other layout sucked when used at the bottom. So we can try with both "tile view".

It does suck, but the film strip is better than that cluster of tiny boxes jammed in the middle.

I am looking into killing off the iframe and moving to using their low level api to render individual, adjustable, windows for each participant. That would also give @Jupi-Taru a natural place to support hiding self view (we would just hide the pane). And a place to add individual volume controls (and mute) for each participant.

Jupi-Taru:
I tried running this as it is right now, and it made the jitsi iframe go to the top of the page and stay there. I think you should put

	$("#meet").css("position", "fixed").css("top", "").css("height", "120px").css("left", "50px").css("width", "75%").css("bottom", "10px");
back into the bottom of init_jitsi

hypergig:
> I tried running this as it is right now, and it made the jitsi iframe go to the top of the page and stay there. I think you should put
> 
> ```
> $("#meet").css("position", "fixed").css("top", "").css("height", "120px").css("left", "50px").css("width", "75%").css("bottom", "10px");
> ```
> 
> back into the bottom of init_jitsi

@Jupi-Taru what were your repro steps to get it to move to the top?

cyruzzo:
so, is this ready for me to test it and merge it ?

hypergig:
> so, is this ready for me to test it and merge it ?

@cyruzzo yep! 

I was unable to reproduce the issue @Jupi-Taru mentioned before and in discord he mentioned that it may have been a camera issue on his end that aggravate it, but if it crops up during testing please let me know and I will address it right away. 

Thanks!

cyruzzo:
Uhmm.. I still get the "tile view" in the bottom like this...

image

cyruzzo:
ah ok.. the fix you talked about on discord for the grid was based on the ability to inject css into the iframe.. which we cannot do.
ok.. I'll merge this as it is

hypergig:
yep.

Just to over communicate, the default layout while on bottom is the speaker layout, and while in modal it will be the tile layout.

It will have to stay this way until jitsi makes some changes on their end _or_ we chose to build a custom UX for AV (lots of effort). This is the best we can for now :/

My next pr adds support for popping the jitsi pane out into a completely new browser window, so at least there is that option for those that wish to have a large AV window _and_ the map at the same time.

PR #129, opened 2021-09-21
Add Jisti Show/Hide Self Button

Jupi-Taru:


cyruzzo:
this add another "host permission" in the meet.jit.si .
I'm not sure I want to add that.
I don't event know what chromes do when you add new permissions.

If anything I think it should be added as an "optional permission" and trigger a request for it.

Jupi-Taru:
OK, it took a lot of trial and error but I figured out how to inject the show/hide button using optional permissions. It injects the code from background.js if it has permissions (webNavigation and meet.jit.si/aboveVTT*) and there's a button to give those permissions in the new options page.

cyruzzo:
Closing this. Sorry for wasting time :(

I really don't want to ask for permissions even optional ones. The firefox review process is already a PITA lately.

hypergig:
it seems like you're able to hide yourself you now natively also

PR #132, opened 2021-09-28
Message broker confirmations

Jupi-Taru:
Added a unique connection_id to the message broker (MB).
Added onlineUserList to the MB, an array of connection IDs of each other online user.
Added sent_messages to the MB, and array of the IDs of all sent messages.
Added received_messages to the MB, and array of the IDs of all received messages.
Added message_confirmations, a dictionary where each key is the ID of a messages sent and each value is and arrays of the connection ids of all users who have send confirmations to that message.
Added unconfirmed_messages, a dictionary where each key is the ID of a message sent but not yet confirmed by everyone in the online user list

Added the ability to send and receive message bundles, which are messages where the data is an array of messages.

When handling messages, the message ID will be checked against the received_messages list, and if it is there it will not be handled (but a confirmation will still be sent for it).

Every 10 seconds MB send an "AmOnline" message to everyone else with their connection id. The receivers add them to their online users list.

When a user sends a message, the id is added to sent_messages, a new entry with an empty array is added to message_confirmations, and (if a confirmation is required) the message is added to unconfirmed_messages.

MB sends a confirmation for every message received (except confirmations, AmOnline, and message bundle messages). These confirmations include the original message id and the MB's connection id. When a confirmation is received, the connection id from the confirmation is added to the array of the message in message_confirmations.

Every second the MB will loop through the messages in unconfirmed messages. If messages have received confirmations, remove them from the unconfirmed messages list. If any unconfirmed messages are older than 1000ms and younger than 10,000ms and have not been confirmed by all online users, resend them in a message bundle. If the message has been confirmed by everyone on the online user list, remove them from the unconfirmed messages list. If the message age is older than 10,000ms, remove the message from the unconfirmed messages list.

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1488,6 +1488,13 @@ function init_ui() {
 
 
 	init_journal($("#message-broker-client").attr("data-gameId"));
+	setTimeout(function() {
+		window.MB.sendAmOnline(10000);
+	}, 1000);
+	
+	setTimeout(function() {
+		window.MB.resendUnconfirmedMessages(1000,10000);
+	}, 1000);
 }
 
 function init_buttons() {
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -76,7 +76,6 @@ function addVideo(stream,streamerid) {
 
 class MessageBroker {
 
-
 	loadWS(token, callback = null) {
 
 		if (callback)
@@ -93,6 +92,7 @@ class MessageBroker {
 		var url = this.url;
 		var userid = this.userid;
 		var gameid = this.gameid;
+		var connection_id = this.connection_id;
 
 		console.log("STARTING MB WITH TOKEN=" + token);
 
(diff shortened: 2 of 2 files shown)


Jupi-Taru:
I wonder if I should leave the timeout for offline player to stop getting messages at 10 seconds, or increase it to 20 or 30?

cyruzzo:
I made some tests (simulating lost messages by losing the websocket from the console) and it seems to recover nicely.

Some initial thought:
- the received_messages is an array that "grows to the infinite". Also. since it's an array and you're using "in_array" the code is continuosly scanning the full array size

- the "Msg Bundle" could potentially create a message that is too big and fail (if it contains a "scene" message for example.. or if it contains notes). Maybe divide the array of messages in group of 5s or 10s and send multiple messages. I don't think it's unlikely for this to happen (simply a user that diconnects just before big moments could cause a bloat in the "unsent messages" that make the whole thing unusable).

Jupi-Taru:
I made the received_messages object a Map for quicker reference, and added a check to delete any messages before the last 1000.

As for message size, I made a general solution. I tested with longer and longer random strings through the chat box, and found that the message fails if it's >42,000 characters. So I put a check in SendMessage to test if the message JSON is bigger than 30,000, and if so to break it into chunks of 20,000 characters that get sent one at a time and reassembled on the other side when all the chunks are received. This pairs nicely with the confirmation/resend scheme as it ensures all the chunks will be delivered successfully.

cyruzzo:
ok, that's great. this should help improve stability by preventing both message loss and also duplicate messages (that I think are happening on the message broker).
I'll do some tests and merge it soon.

cyruzzo:
uhmm.. the chunk size should be much higher.. on a test file someone sent me I had scenes of 130k and they were working.

cyruzzo:
is the "resend" process still working ? I tried to simulate loosing a message (so the sender did not receive a "confirm") and no "resend" is triggered now.

Jupi-Taru:
It should be, but it currently gives up after 10 seconds of not getting a response. I'll double check that I didn't break it somehow

cyruzzo:
tried again after a restart,   I probably broke something by stress testing it with a big scene
btw. I'm using a maxmessage size of 100k and a chunk size of 80k

Jupi-Taru:
Well, if the message is too big to be successfully delivered, then the message broker will give up on that message because it will never get a confirmation.
When you say 100k are you talking about byte size or JSON character length?

Jupi-Taru:
You can change the message timeout by setting the last parameter here to higher than 10,000(ms)

https://github.com/Jupi-Taru/AboveVTT/blob/Message-broker-confirmations/Main.js#L1496

Jupi-Taru:
Changed the chunking to a loop, so now we should avoid those pesky max call stack size issues.

Jupi-Taru:
I added an window.onunload event that sends a "goodbye" message. This removes them from the online user list, so the message broker will stop looking for message confirmations from the disconnected user. It also disconnects their dice stream, as it turns out RTCPeerConnection.onconnectionstatechange is not supported in Firefox, so the streams may be left open.

cyruzzo:
Uhmm... The dicestreaming feature should arlready be disabled (and the button hidden) on firefox.

Jupi-Taru:
Oh, ok, I guess that's not an issue. This will still delete the dice stream right away, rather than after 10-20s timeout, so that might be useful. I can remove it if you'd like, though.

cyruzzo:
image

Chunks (like msgbundl) should not require individual confirmation. Instead, there should be a confirmation only for the message contained into the the chunk (if needed) like it's for msgbundle.

Otherwise you can very possibly end in a unrecoverable loop that eat your memory.

suppose you have
unconfirmed_message {'idmessage1': {very big object} }  (may as well be a series of smaller messages that together go over chunk size)

now "resend unconfirmed message run..."
it generate a new MsgBundle thas is over max length
now this gets split into two chunks  (chunk1, chunk2) ,
we now have 
unconfirmed_message {'idmessage1': {very big object}, 'chunk1.1': {chunk1} , 'chunk1.2': {chunk2} }

bla bla.. this  may very easily grow on a situation on which it can't recover.
I had this happen randomly one time.

Jupi-Taru:
Added in Cyruzzo's bug fixes
 Fixed issue with IDs. Changed senddelay. Late init
 https://github.com/cyruzzo/AboveVTT/commit/f24a163863c892e993e8533994384b209aea0e96

PR #133, opened 2021-10-06
Load player stats from json

Jupi-Taru:
Using the DnDBeyond character API at https://character-service.dndbeyond.com/character/v4/character/ and the the DnDBeyond javascript middleware at https://media.dndbeyond.com/character-tools/vendors~characterTools.bundle.f8b53c07d1796f1d29cb.min.js, as well as code for how to combine the two from https://github.com/TeaWithLucas/DNDBeyond-DM-Screen, fetch the player character info without having to load the player sheet web page. The update runs every 10 seconds.

Tokens will only be updated if any stats have changed, otherwise there would be a flood of token update messages.

Also, DMs will no longer automatically open all player sheets or keep them open in the background. The flag window.KEEP_PLAYER_SHEET_LOADED can turn that function back on if we want it to be an option in the future

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1551,8 +1551,9 @@ function init_ui() {
 				retriveRules();
 				loadModules(initalModules);
 				window.character_update_task = setInterval(get_pclist_player_data, 10000);
-            }
-		},100);
+			}
+		}, 100);
+	}
 }
 
 function init_buttons() {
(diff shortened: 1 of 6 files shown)


Main.js
```diff
@@ -1488,6 +1537,23 @@ function init_ui() {
 
 
 	init_journal($("#message-broker-client").attr("data-gameId"));
+	
+	if (window.DM) {
+		// use DDB character tools to update character info every 10 seconds
+		loadScript("https://media.dndbeyond.com/character-tools/vendors~characterTools.bundle.f8b53c07d1796f1d29cb.min.js",
```
hypergig, line 1543: just to be clear, the only reason why you defer this load is because you need to wait to see if the current user is the DM, right?

hypergig (approved):
this looks pretty solid to me

Main.js
Jupi-Taru: Yes, I didn't want to load that large and busy javascript file unless it's needed by the DM

PR #140, opened 2021-10-16
Use number input instead of text

tehaleph:
The type of inputs for entering HP has been changed to a "number".

It may sound strange, but it is not always easy to enter the resulting HP and it seems to me that it will be more useful to be able to manually change HP.


tehaleph:
Strange, I didn't get this error. In any case, I close the PR, 'cause this thing with calculation in the HP solves my problem completely, thanks for the hint!

PR #154, opened 2021-11-28
Fixed draw perf by throttling mousemove events

Etus12:


Etus12:
Video: https://media.discordapp.net/attachments/832199739398946817/914615874751701022/fps_abovevtt.mp4

PR #162, opened 2021-12-06
Various UI-Tweaks and additions

cyelis1224:
adjusted button placement and scaling and removed the scroll bars as well as added a color picker to the drawing tool.

Please see attached video

https://user-images.githubusercontent.com/82621530/144784954-f76a957f-e447-479f-b5ae-21a68984ae4c.mp4

.

### DIFF
--- a/AOETemplates.js
+++ b/AOETemplates.js
@@ -1,4 +1,3 @@
-
 // Lines are not supported yet
 const AOE_TEMPLATES = {
     'dark-circle': 'https://abovevtt-assets.s3.eu-central-1.amazonaws.com/aoe/SpellToken_DarkCircle.png',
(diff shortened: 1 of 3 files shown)


cyruzzo:
I like having the color picker there. Can we have it so that after you use the select picker that color is actually selected instead of changing the last one in the palette ? Cause that's a bit confusing.

Can you make a PR with just the color picker functionality ? It's better to keep different features in different PRs so it's easier to manage them.

I'm not sure about the red to blue change... but I gues it makes more sense since a lot of DDB stuff is blue. I'll probably require some feedback from the users about that.

This also interacts badly with the campaign page before joining abovevtt. Some uf the buttons become large and out of position:

image

PR #163, opened 2021-12-06
Add player token customization

vlaminck:
## Trello Card

https://trello.com/c/XCPdCfm2

## Features

This adds custom token images for player tokens. Player tokens can be customized by right-clicking the token image in the Players pane. 

### Players Pane Updates

**Token Context Menu**
If the player token has not been placed, the context menu has four options: "Place Token", "Place Hidden token", "Copy Url", and "Customize". Once a token has been placed, the place token options are replaced with "Locate Placed Token".

**Drag and Drop Token Placement**
Player tokens can now be placed on a scene, by dragging the player token image onto a scene and dropping it where you want it placed. If a token has already been placed on a scene, the token will not be placed, but instead will highlight the already placed token.

avtt-player-token-context-menu

### Player Token Customization Modal
Just like the monsters modal, player tokens now support multiple images. Right-click the token in the players pane, and select "Customize" will open the customization modal. 

**Image list**
If there are no images, the modal displays the character portrait, and has a form at the bottom to replace the current image.

**Defaulting back to the portrait**
If there are one or more custom images, the form also includes a "Remove All Custom Images" button which will reset back to using the character portrait.

**Drag and Drop**
Each custom image can be drag and dropped onto the scene as long as there isn't already a player token in the scene. If there is, the already placed token will be highlighted.

avtt-player-token-customize

### Changing the image of an already placed token

Right-clicking a placed token provides an option to "Change Image" which will open the customization modal. This modal acts exactly like the modal accessed via the Players Pane with one notable exception.

**Set For This Token Only**
At the bottom of the modal is a new button titled "Set For This Token Only" which allows the image to be set on the token without saving it for future use.

avtt-player-token-placed-image-change

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1512,6 +1512,7 @@ function init_ui() {
 
 	token_menu();
 	load_custom_monster_image_mapping();
+	register_player_token_customization_context_menu();
 
 
 	window.WaypointManager=new WaypointManagerClass();
--- a/MonsterPanel.js
+++ b/MonsterPanel.js
@@ -160,10 +160,15 @@ function register_custom_monster_image_context_menu() {
 			remove: { 
 				name: "Remove",
 				callback: function(itemKey, opt, originalEvent) {
+					let selectedItem = $(opt.$trigger[0]);
+					let imgIndex = parseInt(selectedItem.attr("data-custom-img-index"));
+					if (imgIndex < 0) {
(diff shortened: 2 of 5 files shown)


cyruzzo:
can you rebase / merge main ? :(

vlaminck:
> can you rebase / merge main ? :(

Done!

PR #165, opened 2021-12-07
Rework locked and add dmLocked

cyelis1224:
Changed locked to only affect player control and added dmLocked to also lock tokens for DM.

see video:

https://user-images.githubusercontent.com/82621530/145080530-b2336c59-852f-46c6-b3d3-6955e4eddbd8.mp4

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -142,8 +142,8 @@ class Token {
 		this.move(this.options.top, newLeft)
 	}
 	move(top, left) {
-		if (this.options.locked || !window.DM && this.options.restrictPlayerMove) return; // don't allow moving if the token is locked
-		if (!window.DM && this.options.restrictPlayerMove) return; // don't allow moving if the token is locked
+		if (window.DM && this.options.locked) return; // don't allow moving if the token is locked
+		if (!window.DM && (this.options.restrictPlayerMove || this.options.locked)) return; // don't allow moving if the token is locked
 		this.update_from_page();
 		this.options.top = top;
 		this.options.left = left;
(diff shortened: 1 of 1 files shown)


cyelis1224:
fixed name scheme and locking issues.

Token.js
```diff
@@ -108,6 +108,7 @@ class Token {
 	}
 	rotate(newRotation) {
 		if (this.options.locked) return; // don't allow rotation if the token is locked
+		if (this.options.restrictPlayerMove) return; // don't allow rotation if the token is locked
```
cyruzzo, line 111: I think this should be if (!window.DM && this.options.restrictPlayerMove ) 

so this block only applies to non DM

Token.js
```diff
@@ -142,6 +143,7 @@ class Token {
 	}
 	move(top, left) {
 		if (this.options.locked) return; // don't allow moving if the token is locked
+		if (this.options.restrictPlayerMove) return; // don't allow moving if the token is locked
```
cyruzzo, line 146: same as before... you cannot just stop movement if the option is set..   you need to check if window.DM

Token.js
```diff
@@ -664,13 +666,22 @@ class Token {
 				old.find("img").removeClass("token-round");
 			}
 			
-			if(this.options.locked){
+			if(this.options.locked && !this.options.restrictPlayerMove){
```
cyruzzo, line 669: also here
if (this.options.locked  ||   (!window.DM && this.options.restrictPlayerMove)

Token.js
```diff
@@ -1036,6 +1068,10 @@ class Token {
 				tok.draggable("disable");
 				tok.removeClass("ui-state-disabled");
 			}
+			if(this.options.restrictPlayerMove){
```
cyruzzo, line 1071: this should only apply if !window.DM ?

cyruzzo (commented):
didn't check througly.. there are probably more places where the conditions need to be checked

cyelis1224:
I'm pretty sure this is good to go now, but that's up to you. Thanks again for your help!

PR #178, opened 2021-12-15
update settings panel UI with example tokens

vlaminck:
This is a UI update to the settings tab. 
1. It updates the style of the import / export section
2. It converts the checkboxes to a toggle that mimics the DDB content sharing settings. 
3. It adds descriptive hover text to all the settings and import / export buttons
4. It adds example tokens that get updated as settings are changed to better show what all the settings do.

avtt-settings-panel-ui-again

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -251,7 +251,8 @@ class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 
 		this.persist();
 
-
+		// some things can't be done correctly until after the scene finishes loading
+		redraw_settings_panel_token_examples();
 
 	}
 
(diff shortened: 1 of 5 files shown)


hypergig:
I love the improved preview.. however, I think I preferred the check  boxes for binary options.. a drop box for just 2 options feels weird to me. But that is just my opinion. Looks awesome!

vlaminck:
> I love the improved preview.. however, I think I preferred the check boxes for binary options.. a drop box for just 2 options feels weird to me. But that is just my opinion. Looks awesome!

I agree. I was hoping the drop downs would help to provide more information about the setting, but it really didn't. I've updated the PR to use toggle elements, and have added hover text to help describe each setting.

PR #181, opened 2021-12-19
WIP: Monsters Panel Rebuild

vlaminck:
This is a rebuild of the monsters tab

# Features

## Buttons
I replaced the encounter drag handle with a configuration gear. Clicking it opens the same token customization modal that the drag handle did. The plus button still adds a token at the middle of the screen, and right clicking it still places a hidden token at the center of the screen.
Screen Shot 2021-12-18 at 8 00 19 PM

## Loading Screen
I took the cool loading screen that DDB uses on the encounters tab, and I cover the monsters tab while we load it in, and rip it apart. 
avtt-monsters-tab-loading-screen

## Drag and Drop
I added the ability to drag a row's image to the scene and drop a token wherever you want. I limited it to the image because clicking the row opens a modal that interferes with jquery draggable in a way that I wasn't able to solve. I hope to revisit this in the future. Holding shift while dropping will set the token as hidden.
avtt-monsters-tab-drag-and-drop

## DDB Dice Rolling
I created a new `EncounterHandler.js` which fetches all encounters associated with the current campaign. If there is an encounter named `AboveVTT` it will be used as a sort of "backing encounter" for the monsters tab. If there isn't an encounter named `AboveVTT`, it will create one. With this backing encounter, we're able to use the DDB dice for monsters and have the dice show up in the Gamelog. If, for some reason, we aren't able to create a backing encounter, it will fallback to the old logic of swapping the dice roll buttons via `scan_monster` function. 
avtt-monster-tab-ddb-dice


hypergig:
very nice

vlaminck:
There is a bug with users who are not subscribers. I'm working on a fix now.

vlaminck:
This work has been included in #185

PR #182, opened 2021-12-20
Add support for non-youtube animated maps

Etus12:
~~WIP, do not merge~~

Done, can be merged following code-review - see latest comment below.

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -110,11 +110,12 @@ function getPlayerIDFromSheet(sheet_url)
 
 window.YTTIMEOUT = null;
 
-function image_load_cb() {
-	alert("Map could not be loaded");
+function map_load_error_cb() {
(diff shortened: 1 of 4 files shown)


Etus12:
https://user-images.githubusercontent.com/5877139/146792483-ba194780-1f3f-46ac-b62d-a85421e49519.mov

Etus12:
This commit adds support for video hosted animated maps.
This uses the browser's "video" capability, which means video files
other than mp4 could be used depending on local codec support.

1. Handles import/export backwards compatibility (previous maps are assumed
   to have is_video as false).
2. Added a new checkbox to the scene editor called "Is Video?" for
   both the player and DM maps. Logic is as follows:
	- If the URL is a youtube link, we handle the map as a YouTube video.
	- If "Is Video?" is checked, we handle the map as a hosted video file.
	- If "Is Video?" is unchecked, we handle the map as an image.

PR #185, opened 2021-12-28
DDB Dice Everywhere (DM only)

vlaminck:
The primary goal of this PR is to use DndBeyond's dice everywhere that rolls can be made. 

### Generic Dice buttons
The dice buttons that DDB has on the lower left corner of the screen have been added, and our dice buttons in the gamelog synchronize with their dice buttons.
avtt-ddb-dice-generic-buttons

### Send to (Default)
The gamelog now supports defaulting rolls to either self or everyone.
avtt-ddb-dice-send-to-default

### Monsters Panel
Dice can be rolled directly inside of a monster stat block within the monsters sidebar panel.
avtt-ddb-dice-monsters-panel

### Placed Token Stat Blocks
Dice can be rolled directly inside of a monster stat block for placed tokens.
avtt-ddb-dice-placed-monster

### Send To Gamelog
Tooltips found in stat blocks are now shown on the screen, and right-clicking them will send them to the gamelog
avtt-ddb-dice-tooltip-to-gamelog

### Encounters
I tried several different ways of doing this, and the only way I could make it work everywhere was to load the game on the `/encounters/:id` page instead of the `/campaign/:id` page. I was really hoping to be able to load an iframe above the game, but it isn't possible to pass clicks through an iframe. So the only way to use the dice rolling is to do it on one of the encounters pages. I chose `/encounters/:id`.

This had its own complications because the encounters page does not use jQuery. I had to change our manifest and Load.js files to inject all the dependencies that we use. I also made some changes to ensure that all of our injected scripts load synchronously instead of asynchronously. This fixed a few issues with race conditions between files that depend on other files.

Once all of our dependencies were properly loaded, I was able to get the game to work on the encounters page with a few minor tweaks. These tweaks are made after checking `is_encounters_page()` which is a convenience function I added to simplify any logic that needed to be separated.

Running the game on the encounters page allows us to 
1. expose the generic dice buttons in the lower left
2. support the `Send To (Default)` feature in the gamelog
3. use DDB dice for the buttons that we inject into the gamelog

### Monsters Panel
The monsters panel now loads the `/encounters/:id/edit` page. This allows us to roll DDB dice from stat blocks found within the panel itself. Because it's an iframe, those rolls are contained within the monsters panel instead of rolling on the full screen. I'm unable to stretch that out across the whole screen without blocking the rest of the UI. Unfortunately, there isn't a way to pass clicks through iframes. I thought about using the edit page for the main screen, but it doesn't have a gamelog on it. While getting this to work, I was able to add drag and drop for monsters as well as update the UI to have a settings cog. I believe there are more improvements that we can make here, but I will circle back to that after this has been merged.

### Placed Monster Stat Blocks
Monsters that have been placed on the scene can now use DDB dice as well. This is done by loading `/combat-tracker/:id` in an iframe instead of the details page for that monster. I used the combat tracker because it has a nicer stat block with a "close" button. I was also able to copy tooltips into the root page when you hover over them. Right-clicking these tooltips will send them to the gamelog which I indicate in the title text for the tooltip. This will be immensely helpful for DMs, and the title text will help with discoverability of the feature.

### EncounterHandler.js
Most of this is coordinated within `EncounterHandler.js`. The `EncounterHandler` class is responsible for making DDB API calls, and keeping the encounter state up to date. If the DM does not have an encounter named `AboveVTT` in the current campaign, the EncounterHandler creates one. 

Any time a monster is added to the scene, the EncounterHandler updates the backing encounter with any new monsters. It only does this if there isn't already a monster of that type on the encounter. 

Any time the backing encounter is updated, the combat-tracker iframe is reloaded to pull in the new stat blocks. This is done by creating a new iframe below everything else and then swapping the old one after it finishes loading. If the user tries to view the new stat block during this time, a loading indicator is shown over the stat block until it finishes loading. 

### Fallbacks And Future Proofing
If, for any reason, the EncounterHandler is unable to create the backing encounter, the old way will be used.
If the DM is not a subscriber, DDB doesn't allow them to use dice on encounters pages. In this case, we continue to use the old way. DDB uses feature flags to determine this, and the `EncounterHandler` is fetching that feature flag. So once they enable that feature for non-subscribers, we will automatically allow non-subscribers to use their dice.


cyruzzo:
I really really  like this.

But just a couple of things worth discussing

- if we keep the "dice roller button" in the bottom - left, we could as well remove the other dice roller under the chat log

- in the new "monster charater sheet" we lose the ability to "send to gamelog" the image of the monsters. I don't have stats but that's a feature that some people like. We also lose the description of the monster (but that can still be opened in the detail page so it's less important)

- when you roll from the monster page it actually ignores the "send to self / everyone" that you can set in the top right of the gamelog.

- DDB dices are slow. Some people like to use them, as they add to the experience. Some people still hate how slow they are. 

This is not a list of things to fix...  maybe we can gather some people and talk about this changes

vlaminck:
> * if we keep the "dice roller button" in the bottom - left, we could as well remove the other dice roller under the chat log
I was wondering about this, too. Once I got them synchronized with our buttons, I kind of like ours better. The positioning makes more sense to me, but their button is consistently in the lower left so I just left it there. I could easily leave their buttons behind everything else instead of pulling it to the front, and it should still work the same.

> * in the new "monster charater sheet" we lose the ability to "send to gamelog" the image of the monsters. I don't have stats but that's a feature that some people like. We also lose the description of the monster (but that can still be opened in the detail page so it's less important)
This is something I've been thinking about as well. They don't have the image there, but we know the monster details page, and I think I could fetch the image and inject it so we don't lose that feature.

> * when you roll from the monster page it actually ignores the "send to self / everyone" that you can set in the top right of the gamelog.
I thought I had that working, but I'll look into it again.

> * DDB dices are slow. Some people like to use them, as they add to the experience. Some people still hate how slow they are.
I could see this being an option in the settings sidebar pane. I don't think that would be too hard to do.

vlaminck:
I just found a bug in the EncounterHandler. I would hold off on merging this until I get a chance to fix it.

vlaminck:
1. I've fixed the bug that I found in the EncounterHandler.
2. I've fixed the bug where the monster stat block wouldn't synchronize with the gamelog. 
3. I've added support for DDB dice on the character sheet. Because of this, I am closing this PR and opening a new one #194

PR #187, opened 2021-12-31
Convert rpgDiceRoll results to use DDB format

vlaminck:
This attempts to parse rpgDiceRolls to DDB format before sending them over the websocket. If the parsing fails for any reason or if the expression is too complex, it falls back to the old way.
avtt-rpgDiceRoll-to-ddb-format

I've also started injecting a "Send to Everyone" button into gamelog messages that were sent "TO: Self". This works for our rolls as well as for DDB dice rolls. The following gif shows an  rpgDiceRoll followed by two DDB rolls from the mobile app. The first is "to self" and the second is "to everyone" to show that it doesn't inject the button for those rolls.
avtt-gamelog-send-to-everyone

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1384,25 +1384,31 @@ function init_ui() {
 		$(".dice-roller > div img[data-count]").each(function() {
 			rollExpression.push($(this).attr("data-count") + $(this).attr("alt"));
 		});
-		const roll = new rpgDiceRoller.DiceRoll(rollExpression.join("+"));
-		const text = roll.output;
-		const uuid = new Date().getTime();
-		const data = {
-			player: window.PLAYER_NAME,
-			img: window.PLAYER_IMG,
-			text: text,
-			dmonly: window.DM || false,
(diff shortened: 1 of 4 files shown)


cyruzzo:
I like it to have things uniform.
But we lose the "send to players" button.
If I merge this all DMs rolls will be automatically visible to players.

vlaminck:
I played around with injecting a "send to everyone" button for all "to self" DDB rolls, but never finished it. I could start working on that again because I agree that it would be very helpful.

As for sending to everyone by default, that will be configurable if we merge #185. I didn't want to dig too deep on that until we had a chance to discuss that PR.

vlaminck:
OK, I've fixed the rolls going to everyone by default. DMs roll to self by default, and players roll to everyone by default.

I also inject a "Send To Everyone" button which works for our messages as well as DDB messages.

cyruzzo:
great

cyruzzo (approved):

PR #188, opened 2021-12-31
Fix chat log links being converted to images.

jleldridge:
~The fix is generally pretty simple. After validating that a string is a URL before injecting it to chat, there was no handling for if  the string was a non-image url. I've added an additional bit of logic to determine whether the link is to an image or not, and generate an `<a>` tag instead of an `` tag if it's not an image type.~

~I've included the most common image types in my additional logic, but if another needs to be accounted for, it's easy enough to add.~

I've gone with an alternative approach to the original. If a link is detected, we try to load an image tag with that link, if the image load succeeds, we create an image tag, otherwise we create an `<a>` tag.

Note: I'm using string interpolation to construct the `<a>` tag, I've not seen that much elsewhere in the code, so if there's some reason not to use that JS feature let me know and I'll remove it.

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1350,7 +1350,11 @@ function init_ui() {
 
 
 			if(validateUrl(text)){
-				text="<img width=200 class='magnify' href=" + parse_img(text) + " src='" + parse_img(text) + "' alt='Chat Image'>";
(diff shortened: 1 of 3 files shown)


cyruzzo:
nice solution.

PR #190, opened 2022-01-02
Image link async unfurling

jleldridge:
This PR further improves image handling in chat by initially posting the image link to the chat, then replacing it with the image if the image can be loaded. If the image can't be loaded, or the link doesn't point to an image, the link stays in place.

This ensures that the chat messages will maintain their order even if messages are being posted very quickly or the load time for an image is very long. This should also be more efficient since we're loading the image once and not loading a separate image in memory before creating the image tag (which would cause us to have to load the image twice).

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -26,27 +26,6 @@ function checkImage(url, callback){
 	image.src = parse_img(url);
 }
 
-function sendLinkOrImage(text, dmonly, whisper, isImg) {
-	let originalText = text;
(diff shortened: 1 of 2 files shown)


cyruzzo:
this is now conflicting with some other stuff I merged. Can you fix the conflict ?

cyruzzo:
(no problem. I fixed that and merged)

PR #194, opened 2022-01-08
Use DDB Dice wherever possible and play on encounters/characters pages

vlaminck:
The primary goal of this PR is to use DndBeyond's dice everywhere that rolls can be made. 

# DM Features
### Generic Dice buttons
The dice buttons that DDB has on the lower left corner of the screen have been added, and our dice buttons in the gamelog synchronize with their dice buttons.
avtt-ddb-dice-generic-buttons

### Send to (Default)
The gamelog now supports defaulting rolls to either self or everyone.
avtt-ddb-dice-send-to-default

### Monsters Panel
Dice can be rolled directly inside of a monster stat block within the monsters sidebar panel.
avtt-ddb-dice-monsters-panel

### Placed Token Stat Blocks
Dice can be rolled directly inside of a monster stat block for placed tokens.
avtt-ddb-dice-placed-monster

### Send To Gamelog
Tooltips found in stat blocks are now shown on the screen, and right-clicking them will send them to the gamelog
avtt-ddb-dice-tooltip-to-gamelog

### Encounters
I tried several different ways of doing this, and the only way I could make it work everywhere was to load the game on the `/encounters/:id` page instead of the `/campaign/:id` page. I was really hoping to be able to load an iframe above the game, but it isn't possible to pass clicks through an iframe. So the only way to use the dice rolling is to do it on one of the encounters pages. I chose `/encounters/:id`.

This had its own complications because the encounters page does not use jQuery. I had to change our manifest and Load.js files to inject all the dependencies that we use. I also made some changes to ensure that all of our injected scripts load synchronously instead of asynchronously. This fixed a few issues with race conditions between files that depend on other files.

Once all of our dependencies were properly loaded, I was able to get the game to work on the encounters page with a few minor tweaks. These tweaks are made after checking `is_encounters_page()` which is a convenience function I added to simplify any logic that needed to be separated.

Running the game on the encounters page allows us to 
1. expose the generic dice buttons in the lower left
2. support the `Send To (Default)` feature in the gamelog
3. use DDB dice for the buttons that we inject into the gamelog

### Monsters Panel
The monsters panel now loads the `/encounters/:id/edit` page. This allows us to roll DDB dice from stat blocks found within the panel itself. Because it's an iframe, those rolls are contained within the monsters panel instead of rolling on the full screen. I'm unable to stretch that out across the whole screen without blocking the rest of the UI. Unfortunately, there isn't a way to pass clicks through iframes. I thought about using the edit page for the main screen, but it doesn't have a gamelog on it. While getting this to work, I was able to add drag and drop for monsters as well as update the UI to have a settings cog. I believe there are more improvements that we can make here, but I will circle back to that after this has been merged.

### Placed Monster Stat Blocks
Monsters that have been placed on the scene can now use DDB dice as well. This is done by loading `/combat-tracker/:id` in an iframe instead of the details page for that monster. I used the combat tracker because it has a nicer stat block with a "close" button. I was also able to copy tooltips into the root page when you hover over them. Right-clicking these tooltips will send them to the gamelog which I indicate in the title text for the tooltip. This will be immensely helpful for DMs, and the title text will help with discoverability of the feature.

### EncounterHandler.js
Most of this is coordinated within `EncounterHandler.js`. The `EncounterHandler` class is responsible for making DDB API calls, and keeping the encounter state up to date. If the DM does not have an encounter named `AboveVTT` in the current campaign, the EncounterHandler creates one. 

Any time a monster is added to the scene, the EncounterHandler updates the backing encounter with any new monsters. It only does this if there isn't already a monster of that type on the encounter. 

Any time the backing encounter is updated, the combat-tracker iframe is reloaded to pull in the new stat blocks. This is done by creating a new iframe below everything else and then swapping the old one after it finishes loading. If the user tries to view the new stat block during this time, a loading indicator is shown over the stat block until it finishes loading. 

### Fallbacks And Future Proofing
If, for any reason, the EncounterHandler is unable to create the backing encounter, the old way will be used.
If the DM is not a subscriber, DDB doesn't allow them to use dice on encounters pages. In this case, we continue to use the old way. DDB uses feature flags to determine this, and the `EncounterHandler` is fetching that feature flag. So once they enable that feature for non-subscribers, we will automatically allow non-subscribers to use their dice. 

# PC Features
### Playing From Character Page
PCs now play on the characters page instead of in an iframe. This allows us to use DDB dice everywhere even when the sheet is "closed". It also allows the Gamelog to use the Send To (Default) feature as well.
avtt-character-ddb-dice

### Thin Character Sheet
Because we're no longer loading character sheets in an iframe, we can't use the native resizing logic that DDB has built. We instead restructure the elements on the screen to be a thin character sheet that is vertically scrolled. This actually makes the thinner version considerably more useable.
avtt-character-sheet-thin

### Sidebar Character Details
Using the character sheet means that clicking on an item to see details now loads over the gamelog. This makes using our tabs a little trickier, but I've done my best to make it act as close to how DDB does it.
avtt-character-sidebar-details

# Opt Out ability
I've also added the ability to opt out of these features from within the Instructions area. This will allow us to push a new build with all of these changes and have a fail-safe if there are any bugs that users experience. 

Screen Shot 2022-01-15 at 10 09 00 AM
Screen Shot 2022-01-15 at 10 09 07 AM

And if they continue to play without the new feature, I move the toggle where it can serve as a reminder that they should switch back at some point.
Screen Shot 2022-01-15 at 10 09 16 AM


cyruzzo:
hmmm. for some reason I cannot open the "apprentice wizard" stats.
It get stuck like this:
image

vlaminck:
> hmmm. for some reason I cannot open the "apprentice wizard" stats.
> It get stuck like this:
> image

This should be fixed now.

PR #196, opened 2022-01-12
add changelog file

vlaminck:
I'm not sure if this is the best place for this, but I'm trying a few options to see what feels right


vlaminck:
Video tags don't render in github 🤦

PR #199, opened 2022-01-15
Make the monster tokens hidden to the players by default.

Kokolokoli:
This PR is intended as a first step into the project. It changes and sets the default token settings inside the init_settings() function.

### DIFF
--- a/Settings.js
+++ b/Settings.js
@@ -157,13 +157,12 @@ function init_settings(){
 				toggle.click();
 			}
 			if (setting.name == 'hidden') toggle.click();
-			//window.TOKEN_SETTINGS[setting.name] = (settings.name == 'hidden'); // Should be able to remove this line since the click does the exact same thing
 		}
 		persist_token_settings(window.TOKEN_SETTINGS);
 		redraw_settings_panel_token_examples();
 	});
 	body.append(resetToDefaults);
-	resetToDefaults.click(); // Sets the default settings instead of all false.
+	resetToDefaults.click(); // Sets the default token settings
 
 
 	const experimental_features = [
--- a/Settings.js
+++ b/Settings.js
@@ -156,12 +156,14 @@ function init_settings(){
 			if (toggle.hasClass("rc-switch-checked")) {
 				toggle.click();
 			}
-			window.TOKEN_SETTINGS[setting.name] = false;
+			if (setting.name == 'hidden') toggle.click();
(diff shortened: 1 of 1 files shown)


Kokolokoli:
Didn't do a commit for it since it didn't require changes, but I did test in Google Chrome as well

vlaminck (commented):
I don't think I understand what this accomplishes. It seems like it would prevent the "Reset Token Settings to Defaults" button from working properly. It seems like it would just make the button toggle the `hidden` feature instead of actually resetting all the settings to their default values.

Kokolokoli:
Since the "Reset Token Settings to Defaults" button calls the click() of each toggle, and the click of each toggle sets the corresponding setting to the new value (See Settings.js line 125, which uses the function defined in SidebarPanel.js line 224).

Reset Token Settings essentially sets everything to false, except for the "hidden" setting where it sets it right back to true.

I removed the window.TOKEN_SETTINGS[name] = false for each toggle in the loop. This is what used to set the initial values when the init_settings was called, but then it needlessly set them a second time whenever clicking the button (since toggle.click already does that). 

Instead, I simply click the "Reset Token Settings to Defaults" after building it and appending it.

Settings.js
```diff
@@ -156,12 +156,13 @@ function init_settings(){
...
 				toggle.click();
 			}
-			window.TOKEN_SETTINGS[setting.name] = false;
+			if (setting.name == 'hidden') toggle.click();
 		}
 		persist_token_settings(window.TOKEN_SETTINGS);
 		redraw_settings_panel_token_examples();
 	});
 	body.append(resetToDefaults);
+	resetToDefaults.click(); // Sets the default token settings
```
vlaminck, line 165: This line in particular just resets all the defaults on page load.

vlaminck:
Other than that line that resets the defaults when the page loads, your logic is fine. I'm curious why we would want to start defaulting `hidden` to `true`, though. This is a departure from current expectations.

Settings.js
Kokolokoli: Yes. Sorry I misunderstood that last part and corrected my above comment. However, the token settings still don't persist in the main branch (everything is false on load).

Kokolokoli:
Extremely good remark.

I'm a big dummy, that's why. I completely misunderstood a perfectly written message. I was supposed to make "Restrict Player Movement" true by default, not "hidden".

https://discord.com/channels/815028457851191326/832199739398946817/931271695560507423

Will close this PR for now.

PR #200, opened 2022-01-15
Add 'Restrict player movement' as a default token setting

Kokolokoli:
This PR is intended as a first step into the project. (See https://discord.com/channels/815028457851191326/832199739398946817/931271695560507423)

It adds the restrictPlayerMove option to the token settings options in the Settings tab.

The "Reset Token Settings to Defaults" button calls the click() of each toggle that is turned on, and the click of each toggle sets the corresponding setting to the new value (See Settings.js line 125, which uses the function defined in SidebarPanel.js line 224).

Before, it simply set everything to false.
Now, Reset Token Settings essentially does the same thing except for the "restrictPlayerMove" setting where it sets it right back to true. I removed the window.TOKEN_SETTINGS[name] = false for each toggle in the loop. It needlessly set them a second time whenever clicking the button (since toggle.click already does that).

### DIFF
--- a/Settings.js
+++ b/Settings.js
@@ -157,13 +157,12 @@ function init_settings(){
 				toggle.click();
 			}
 			if (setting.name == 'hidden') toggle.click();
-			//window.TOKEN_SETTINGS[setting.name] = (settings.name == 'hidden'); // Should be able to remove this line since the click does the exact same thing
 		}
 		persist_token_settings(window.TOKEN_SETTINGS);
 		redraw_settings_panel_token_examples();
 	});
 	body.append(resetToDefaults);
-	resetToDefaults.click(); // Sets the default settings instead of all false.
+	resetToDefaults.click(); // Sets the default token settings
 
 
 	const experimental_features = [
--- a/Settings.js
+++ b/Settings.js
@@ -128,7 +128,6 @@ function init_settings(){
 	for(let i = 0; i < token_settings.length; i++) {
 		let setting = token_settings[i];
 		let currentValue = window.TOKEN_SETTINGS[setting.name];
-		console.log('CURRENT VALUE: ', currentValue);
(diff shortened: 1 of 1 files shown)


Kokolokoli:
@vlaminck Regarding the line that sets the default values on page load, I removed it after reexamining the request. Thank you for the remark on the previous PR, it made me notice my mistake!

Settings.js
```diff
@@ -156,14 +163,13 @@ function init_settings(){
 			if (toggle.hasClass("rc-switch-checked")) {
 				toggle.click();
 			}
-			window.TOKEN_SETTINGS[setting.name] = false;
+			if (setting.name == 'restrictPlayerMove') toggle.click();
```
vlaminck, line 166: I don't think we want to default this to `true` because that changes the current default behavior of allowing players to move any token. So I would remove this line.

You're right that we don't need the line that you removed. I think I added that to ensure that every option had a value in the settings object. Otherwise, it's not guaranteed to be there. That's not actually a big deal so removing that line is fine, but I like the idea that if we have a setting we also have a value for it. Thinking through it now, that logic belongs in some sort of "on load" migration process. It's also unnecessary so no need to do it at all. This is a long winded way of saying it's ok to remove that line 😆

Kokolokoli: Alrighty
If it can relieve any concerns, all the TOKEN_SETTINGS are undefined on startup and in practice it acts just like if they were ```false```... but you probably already know that ahaha

Settings.js
```diff
@@ -122,6 +128,7 @@ function init_settings(){
 	for(let i = 0; i < token_settings.length; i++) {
 		let setting = token_settings[i];
 		let currentValue = window.TOKEN_SETTINGS[setting.name];
+		console.log('CURRENT VALUE: ', currentValue);
```
Kokolokoli, line 131: @vlaminck  I will remove the console.log as well, but what do you think of setting the settings to ```false``` here instead?  At least for the time being, so they all have a value from the get go.

vlaminck (approved):

PR #212, opened 2022-02-20
UI update 70

cyelis1224:
Recreation of the UI in the same color scheme and style of the changes made by Mikedave. technically probably not going to be in 70, but thats when i built it so... Let me know what you think?


cyelis1224:
previous commit fixed an issue that caused incompatibility with the CT-Scaler this should be fine now.

cyruzzo:
this has probably become obsolete

PR #235, opened 2022-03-10
re-introduce the guest dm feature

vlaminck:
This works because the `.joindm` click will initialize the EncounterHandler with an encounter on the current user's account. Everything I tested seems to work just fine, but it was mentioned that there could be some cloud issues with this feature.


cyruzzo:
we will re-consider when everyone is in the cloud

Issue #224, opened 2022-03-09
Homebrew monster images aren't loading in stat blocks


vlaminck:
DDB gives us 2 urls for monster images. A large size, and a thumbnail size. We default to the large one, and fall back to the thumbnail size if the large one doesn't exist. However, for homebrew monsters, the large image url exists, but isn't accessible. 

This is the fallback code in `EncounterHandler`
```
	let url = stat.data.largeAvatarUrl;
	if (url === undefined || url == null || url.length == 0) {
		url = stat.data.avatarUrl;
	}
```

We might be able to just use `avatarUrl`, but the best solution would be to hit the url to verify it's valid.

PR #243, opened 2022-03-11
#224 - fix for homebrew monsters not showing any image

bain2236:
homebrew now shoes token as image
added .history to ignore for use with local history plugin in vs code

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -679,6 +679,30 @@ function open_monster_stat_block_with_stat(stat, tokenId) {
 	}
 }
 
+
+function fetch_mon_image(url) {
+	console.debug(`fetch_mon_image starting`);
+	get_cobalt_token(function (token) {
+		$.ajax({
+			url: url,
+			beforeSend: function (xhr) {
+				xhr.setRequestHeader('Authorization', 'Bearer ' + token);
+			},
+			xhrFields: {
+				withCredentials: true
+			},
+			success: function (responseData) {
+				console.log(`fetch_mon_image succeeded`);
+				return 200
+			},
+			failure: function (errorMessage) {
+				console.warn(`fetch_mon_image failed ${errorMessage}`);
+				return 403
+			}
+		});
+	});
+}
+
 function inject_monster_image(stat) {
 	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length > 0) {
 		// we only need one
@@ -689,7 +713,9 @@ function inject_monster_image(stat) {
 		return;
 	}
 	let url = stat.data.largeAvatarUrl;
-	if (url === undefined || url == null || url.length == 0) {
+	
+	if (url === undefined || url == null || url.length == 0 || fetch_mon_image(url) !== 200) {
+		console.log("using monsters avatar instead of large avatar")
 		url = stat.data.avatarUrl;
 	}
 	if (url !== undefined && url != null && url.length > 0 && window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 .vscode
+.history


vlaminck (commented):
If we want to fetch the image to validate it, I think we'll need to convert `inject_monster_image` to be asynchronous. However, while reviewing this PR, I had a couple of ideas. 

First is to just use `avatarUrl` all the time. It's less data to download and render, but we might find monsters that didn't have an `avatarUrl` or had a different image and then we're right back where we started, but with lower quality images.

The other idea I had was to do something like this with the `img` tag:

    

to be clear, I have no idea if this will work, but the theory is to try and load `largeAvatarUrl`, and if there's an error, swap the `src` to be `avatarUrl`. It also replaces the `onerror` function in case that image doesn't load either. If this works, it might be worth creating a separate function that does the src swapping and maybe hides the image if both fail, but that's more of a nice to have than it is necessary.

EncounterHandler.js
```diff
@@ -679,6 +679,30 @@ function open_monster_stat_block_with_stat(stat, tokenId) {
 	}
 }
 
+
+function fetch_mon_image(url) {
+	console.debug(`fetch_mon_image starting`);
+	get_cobalt_token(function (token) {
```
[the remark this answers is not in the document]
bain2236: Aye makes total sense. Just nabbed a snippet from elsewhere. Removed the cobalt token stuffs

EncounterHandler.js
```diff
@@ -689,7 +689,13 @@ function inject_monster_image(stat) {
 		return;
 	}
 	let url = stat.data.largeAvatarUrl;
-	if (url === undefined || url == null || url.length == 0) {
+	// attempt to get the avatar to
+	let largeAvatarReponse = $.ajax({
+		async: false,
```
vlaminck, line 694: I'm a little worried that making this synchronous will cause performance issues. If the request takes a long time or times out, the code after this won't execute until the network request completes.

EncounterHandler.js
```diff
@@ -689,12 +689,10 @@ function inject_monster_image(stat) {
 		return;
 	}
 	let url = stat.data.largeAvatarUrl;
-	if (url === undefined || url == null || url.length == 0) {
```
vlaminck, line 692: It's nice that we can drop this. I would also drop the `let url` above, and remove `url !== undefined && url != null && url.length > 0 && ` below. Then you'll need to change the img `src` to reference `stat.data.largeAvatarUrl` as well. 

Sorry to be nitpicky on this, but I think that would clean up this code nicely.

EncounterHandler.js
```diff
@@ -689,12 +689,10 @@ function inject_monster_image(stat) {
...
-		url = stat.data.avatarUrl;
-	}
+
 	if (url !== undefined && url != null && url.length > 0 && window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
-		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");				
-		let image = `<img style="width:100%" class="injected-image" src="${url}" alt="${stat.data.name}" class="monster-image">`;
+		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
+		let image = `<img style="width:100%" class="injected-image" src="${url}" alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'"></img>`;
```
vlaminck, line 695: The last thing I think you need here is to add `; onerror=''` so it ends up like this

    onerror="this.src='${stat.data.avatarUrl}';onerror='';"

Otherwise we could end up in an infinite loop if `stat.data.avatarUrl` also errors out

EncounterHandler.js
bain2236: no need to apologise at all, I completely agree. I do think initially I added the `stat.data.largeAvatarUrl` as the source then I clocked I still had the let url so removed it

EncounterHandler.js
```diff
@@ -688,13 +688,10 @@ function inject_monster_image(stat) {
...
-	if (url === undefined || url == null || url.length == 0) {
-		url = stat.data.avatarUrl;
-	}
-	if (url !== undefined && url != null && url.length > 0 && window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
-		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");				
-		let image = `<img style="width:100%" class="injected-image" src="${url}" alt="${stat.data.name}" class="monster-image">`;
+	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
+		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
+		let image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
+			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'";onerror=''></img>`;
```
vlaminck, line 694: Very close. The `onerror=''` needs to be inside the `"` like this:

    onerror="this.src='${stat.data.avatarUrl}';onerror='';"


That way, when the `onerror` function executes, it changes the src and then changes the original onerror function.

PR #245, opened 2022-03-11
Fix/238 zoom to fit

### LINKED ISSUE #238: The zoom to fit button should fit the map in the window
Rather than just setting the zoom to 1.0, we should try to align the map in the center of the view at whatever zoom makes it fit

bain2236:
fix #238 
big thanks to hypergig for figuring out the min work in the `reset_zoom` function
Added reset_zoom call when the scene loads so it auto fits the map to the viewport
Added localstorage for zoom/scroll positioning per scene. You can now zoom on a scene, change scene and back and your zoom/positioning will be restored 💪 

I've not tested anything when it comes to cloud storage / changing window size etc but happy to get that going within this PR
I tried testing it as a player but the localstorage was shared and I couldn't log in to DNDB incognito


Main.js
```diff
@@ -86,7 +86,72 @@ function decrease_zoom() {
 }
 
 function reset_zoom () {
-	change_zoom(60.0 / window.CURRENT_SCENE_DATA.hpps);
+	console.group("reset_zoom")
```
bain2236, line 89: This weird indentation isn't visible in vscode so assuming github is doing something stupid?!

PR #248, opened 2022-03-11
add functions for saving custom tokens in the cloud

vlaminck:
I tested the local functions, but I haven't tested the cloud functions because I don't know how to run the backend locally. Here is a backend PR for getting and setting the tokens in the cloud https://github.com/cyruzzo/AboveVTT-Backend/pull/2

`get_user_cloud_key` returns a url encoded and base64 encoded value which should be safe to use for making requests to the cloud. I'm not sure if that will be a problem for use with whatever DB the cloud uses, though.

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -210,6 +210,17 @@ class EncounterHandler {
 
 	}
 
+	get_or_fetch_campaign_info(callback) {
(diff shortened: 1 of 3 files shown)


cyruzzo:
cleaning up... reopen if needed

PR #249, opened 2022-03-11
persist player token customizations in the cloud

### LINKED ISSUE #295: Migrate player token customizations to the cloud

vlaminck:
player token customizations are specific to a campaign so I just added them to the campaignData that is already persisted in the cloud. I tested with two accounts across firefox and chrome and seems to be working just fine

Closes #295


vlaminck:
As discussed, this is not the place for player token customizations. I'll have a different PR for that soon

PR #284, opened 2022-03-16
Draw outside game space

### LINKED ISSUE #246: Allow selection and fog to extend outside map
It's really hard to select tokens on the edge of the map, and it's impossible to draw/clear fog all the way to the edge of the map. Being able to draw/select outside of the map would be a great QoL improvement.

Azmoria:
Fixes  #246  Allows edge/outside gamespace selection, drawing, measure etc.

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -651,13 +651,7 @@ function redraw_drawings() {
 function stop_drawing() {
 	$("#reveal").css("background-color", "");
 	window.MOUSEDOWN = false;
-	var target = $("#fog_overlay");
-	target.css('cursor', '');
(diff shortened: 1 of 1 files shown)


Azmoria:
Example video: https://youtu.be/xHOWdwKEwCU

cyruzzo:
this is causing an issue... maybe with z-index or something.

try this:
- join as a player
- zoom out a lot pressing the zoom button

at a certain point you won't be able to press the zoom button anymore as they get covered by VTTWrapper

Azmoria:
I think that catches everything in the player view.

PR #285, opened 2022-03-16
#276 : resolved homebrew monster not cascading onerror down to MB

### LINKED ISSUE #276: sending homebrew monster image to gamelog is coming back with a broken image link in the game log. it is also showing the "avatar" picture and not the large picture
sending homebrew monster image to gamelog is coming back with a broken image link in the game log. it is also showing the "avatar" picture and not the large picture in the character sheet window

This is after 0.71.4 . Probably the fix to #224 only fixed the displaying part in the character sheet ?

bain2236:
fix #276 
This is a homebrew monster.
It appears that the `onerror` does not propogate down into the MB. Tried a few different ways to get it to go down but this seemed like the most sensible solution
image

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -690,8 +690,12 @@ function inject_monster_image(stat) {
 	}
 	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
 		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
-		let image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
+		const image = stat.data.isHomebrew ?
+			`<img style="width:100%" class="injected-image" src="${stat.data.avatarUrl}"
+			alt="${stat.data.name}" class="monster-image" ;onerror=''></img>`
+		 	: `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
 			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'";onerror=''></img>`;
+
 		content.find(".mon-stat-block").after(image);
 		let button = $("<button class='ddbeb-button monster-details-link'>SEND IMAGE TO GAMELOG</button>");
 		button.css({ "float": "right" });


EncounterHandler.js
```diff
@@ -690,8 +690,12 @@ function inject_monster_image(stat) {
 	}
 	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
 		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
-		let image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
+		const image = stat.data.isHomebrew ?
+			`<img style="width:100%" class="injected-image" src="${stat.data.avatarUrl}"
+			alt="${stat.data.name}" class="monster-image" ;onerror=''></img>`
```
cyruzzo, line 695: uhmm.. isn't that ";" wrong ?

EncounterHandler.js
```diff
@@ -690,8 +690,12 @@ function inject_monster_image(stat) {
...
+		 	: `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
 			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'";onerror=''></img>`;
```
cyruzzo, line 697: uhmm.  this also has two "onerror"

also.. on the receiving side everything is passed through DomPurify.. so any javscript is removed when sent through the gamelog

bain2236: Aye, so if the first fails it rolls over to the avatar images and if that fails it lands on the empty string to avoid infinite looping.

This doesn't pass the onerror down, but it does pass the correct src down when it's a homebrew rather than relying on the onerror.

Homebrew monsters 403 when you try to get the large image.

I do think the comment about the semicolons is correct though, I'll take a look in the morning

vlaminck: Yeah, this part 

    onerror="this.src='${stat.data.avatarUrl}'";onerror=''

Should be 

    onerror="this.src='${stat.data.avatarUrl}';onerror='';"

The original onerror function needs to set the src and replace itself with an empty onerror function.

bain2236: Cheers, should now be sorted

EncounterHandler.js
```diff
@@ -690,16 +690,18 @@ function inject_monster_image(stat) {
 	}
 	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
 		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
-		let image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
-			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'";onerror=''></img>`;
+		const image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
+			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}';onerror='';"></img>`;
```
vlaminck, line 694: I'm sorry for all the back and forth on this one, but I just learned that `onerror` is deprecated 🤦 
I think you should use something like `$(image).on("error", function(event) { ... });` instead. That will probably be a lot easier to read anyway.

bain2236: Was a bit fiddly, I couuldn't just do `$(image)[0].src = stat.data.avatarUrl` so just went with removing the previous image and loading a new one

EncounterHandler.js
```diff
@@ -690,16 +690,23 @@ function inject_monster_image(stat) {
 	}
 	if (window.EncounterHandler.combat_body.find(".encounter-details-content-section__content .injected-image").length == 0) {
 		let content = window.EncounterHandler.combat_body.find(".encounter-details-content-section__content");
-		let image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
-			alt="${stat.data.name}" class="monster-image" onerror="this.src='${stat.data.avatarUrl}'";onerror=''></img>`;
+		const image = `<img style="width:100%" class="injected-image" src="${stat.data.largeAvatarUrl}"
+			alt="${stat.data.name}" class="monster-image"></img>`;
+		$(image).on("error", () => {
```
vlaminck, line 695: If you want, you can update this to be `(event) => {` which would allow you to get the image with `$(event.target)` rather than searching the entire iframe document for it. Not a big deal, but slightly more efficient.

EncounterHandler.js
vlaminck: ahhh. Makes sense

PR #311, opened 2022-03-20
#298 : made dead X a hidden element, shows when dead

### LINKED ISSUE #298: Token Dead X sometimes disappears
This is hard to reproduce, but I think it can be solved by always having the dead element on tokens and then hide/show it instead of adding and removing it.

bain2236:
fixes #298 
added trim to text box input as I saw it mentioned that placing whitespace can have adverse effects
made dead X a hidden element, shows when dead
did some cleanup of related code as kept seeing duplicated fragments

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -138,9 +138,7 @@ class Token {
 		// this just rotates locally to help with performance.
 		// draggable.stop will call place_sync_persist to finalize the rotation. 
 		// If we ever want this to send to all players in real time, simply comment out the rest of this function and call place_sync_persist() instead.
-		let scale = (((this.options.size - 15) * 100) / this.options.size) / 100;
-		if ( !(this.options.max_hp) > 0 || (this.options.disableaura))
-			scale=1;
+		let scale = this.get_token_scale()
 		
 		var selector = "div[data-id='" + this.options.id + "']";
 		var tokenElement = $("#tokens").find(selector);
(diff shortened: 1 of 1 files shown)


Token.js
```diff
@@ -643,37 +671,13 @@ class Token {
...
-			let scale = (((this.options.size - 15) * 100) / this.options.size) / 100;
-			if (!(this.options.max_hp) > 0 || (this.options.disableaura))
-				scale = 1;
+			let scale = this.get_token_scale()
 			var rotation = 0;
 			if (this.options.rotation != undefined) {
 				rotation = this.options.rotation;
 			}
 			old.find("img").css("transform", "scale(" + scale + ") rotate("+rotation+"deg)");
-
```
bain2236, line 655: this block is now called via the 2 new functions at the end of place, so it will be applied to both new and old tokens

Token.js
```diff
@@ -848,27 +851,7 @@ class Token {
 				tok.find(".elev").show();
 			}
 
-			// HEALTH AURA / DEAD CROSS
```
bain2236, line 851: Same again on this one. it's called at the end of place via the 2 new functions

Token.js
```diff
@@ -268,8 +266,59 @@ class Token {
...
+		}
+	}
+
+	/**
+	 * updates the colour of the health aura if enabled
+	 * @param token jquery selected div with the class token
+	 */
+	update_health_aura(token){
+		console.group("update_health_aura")
+		if ((!this.options.disableaura) && this.options.max_hp > 0) {
```
vlaminck, line 297: You don't need the extra parentheses around `!this.options.disableaura`. Doesn't hurt to have them, but doesn't actually do anything either.

Token.js
```diff
@@ -268,8 +266,59 @@ class Token {
...
+	}
+
+	/**
+	 * returns different scales of the token based on options such as aura disabled
+	 * @returns scale of token
+	 */
+	get_token_scale(){
+		let scale = (((this.options.size - 15) * 100) / this.options.size) / 100;
+			
+		if (!(this.options.max_hp) > 0 || (this.options.disableaura))
```
vlaminck, line 315: You can clean up some of the parentheses here, too. Looks like they were there before, but as long as you're refactoring, we might as well clean it up a bit.

And as long as you're cleaning things up, I would reverse the logic in this function. Do this check first and return 1 early. That way we don't have to do all that math above if we're not even going to use it.

Token.js
```diff
@@ -287,7 +336,7 @@ class Token {
 			delete this.options.hidden;
 
 		if ( ( (!(this.options.monster > 0)) || window.DM || (!window.DM && this.options.hidestat)) && !this.options.disablestat && old.find(".hp").length > 0) {
-			if (old.find(".hp").val().startsWith("+") || old.find(".hp").val().startsWith("-")) {
+			if (old.find(".hp").val().trim().startsWith("+") || old.find(".hp").val().trim().startsWith("-")) {
```
vlaminck, line 339: good call on the `trim()` here

Token.js
vlaminck: The only thing that concerns me in this PR is that you're not taking this part 

    const pData = window.PLAYER_STATS[this.options.id] || this.options;

I'm not exactly sure what the difference is between `window.PLAYER_STATS` and `token.options` at this point in the code, but it looks like it's still used in other places so you should probably check it in your `update_dead_cross` and `update_health_aura` functions. If I'm wrong, and the token is already synchronized with these values, then using the token values is just fine. I'm not familiar with how player stats get sync'd, but I noticed that this one was slightly different from the others, and wnated to point it out just in case.

vlaminck (commented):
The only thing that stands out to me on this is the `window.PLAYER_STATS`. Maybe @cyruzzo or someone else who knows more about that part of the app can comment on it.

But other than that, this looks great! 👍

vlaminck (changes requested):
I think this needs to check `window.PLAYER_STATS`
Sorry for the double comment, I want to test the automation in the Bugs Project.

Token.js
bain2236: aye thought that would be the case but just snipped it from somewhere else. So used to having my hands held with a linter

vlaminck (approved):
This looks great!

PR #320, opened 2022-03-22
Stop selection box when clicking out of context and add back hold shift to add tokens to selection.

### LINKED ISSUE #319: Clicking outside of the token context menu starts drawing a selection box
We have this issue since forever and it's quite annoying :)

You get kinda "trapped" into that selection.

Azmoria:
Should fix #319 

• Stop selection box from showing up when clicking out of a context menu
• When shift is held add tokens to selection instead of new selection.


Azmoria:
Example video of the 2 fixes: https://www.youtube.com/watch?v=y_QXZ8Ofjcw

PR #334, opened 2022-03-24
Update MessageBroker.js

### LINKED ISSUE #333: Token delete not syncing in player view
When deleting a token it isn't deleted on the player view until refresh. If the player moves the token in the mean time it adds it back to the DM view.

Azmoria:
Fixes #333 This should allow the token delete to sync again. The problem seems to be the delete function was stopping if they weren't deleteable by players. This sets the deleteable by players flag to true for delete messages sent.

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -392,6 +392,7 @@ class MessageBroker {
 			if(msg.eventType=="custom/myVTT/delete_token"){
 				let tokenid=msg.data.id;
 				if(tokenid in window.TOKEN_OBJECTS)
+					window.TOKEN_OBJECTS[tokenid].options.deleteableByPlayers = true;
 					window.TOKEN_OBJECTS[tokenid].delete(false,false);
 			}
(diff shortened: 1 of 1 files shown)


MessageBroker.js
```diff
@@ -392,6 +392,7 @@ class MessageBroker {
 			if(msg.eventType=="custom/myVTT/delete_token"){
 				let tokenid=msg.data.id;
 				if(tokenid in window.TOKEN_OBJECTS)
+					window.TOKEN_OBJECTS[tokenid].options.deleteableByPlayers = true;
```
vlaminck, line 395: The early return at the beginning of Token.delete was introduced when the `deletableByPlayers` feature was introduced. I think we should just roll back that bit of logic instead of working around it here. We'll need to double check everywhere that calls `delete` to make sure we don't accidentally allow players to delete them. If that gets too messy, and this this is actually the cleanest option, then what you have here is good.

cyruzzo:
really hacky.. but quick

PR #339, opened 2022-03-25
adding a token to the journal clears the input

bain2236:
Minor QOL improvement, clicking add clears the url so you don't have to manually remove the url

### DIFF
--- a/SidebarPanel.js
+++ b/SidebarPanel.js
@@ -214,6 +214,7 @@ class SidebarPanel {
           return;
         }	
         imageUrlEntered(imageUrl);
+        $(this).parent().find("input").val("")
       }
     });
 


vlaminck:
I, personally, don’t want this because adding token images with sequential URLs is insanely fast and easy. This change would make that process more tedious. It makes sense, and I agree that this is probably the right thing to do, but I don’t want it 😞

Most content providers that

vlaminck:
I just realized this only clears it when you click the "add" button 🤦 
Ignore me. I completely agree that this should be merged 👍

vlaminck (approved):

bain2236:
is that you just hitting enter? I didn't know that was a thing

Issue #303, opened 2022-03-18
Improve Combat Tracker UI


vlaminck:
This needs to be discussed as a group. 

Do we want the UI to be much different or do we want to just give it a light facelift?

vlaminck:
Is there anything else we want to do here? I feel like this was completed with #346

vlaminck:
Related PRs 
* #455
* #443
* #438
* #416
* #414
* #412 

I think this issue can be closed. Most of these are simple bug fixes at this point.

PR #346, opened 2022-03-26
UI Update - Combat tracker and Frame moving/resizing

Azmoria:
I had to make some changes to both my PR's to get these to play well together. I switched the original PRs to draft. 

If you would like I will split them back out into the draft PRs. TBH I just didn't want to put in the effort to split the changes if you're ok with this PR but I'm happy to do it if you like.

This does the below and takes care of at least a part of #303 

• Combat tracker update
>      • Updated look - Icon buttons, new sizing and slightly different layout.
>      • Added title bar
>      • Moveable and resizable by DM's and players
>      • Opens the next monster stat block in the monster frame when next is clicked

• Dm's monster and player frames
>      • Resizable and moveable
>      • Removed player toggle size button (can be resized so not needed and wasn't functional with resizing)
>      • Adjusted close buttons to match each other
>      • Title bar for monster frame

• All moveable frames are contained within the window.
• Resizable frames have a min width and height to keep them from becoming too small to enlarge without glitchyness
• Moveable items will be at the top of the stack when clicked
     • Note: Monster and player frames you must click the titlebar or the resize handles to focus it. Clicking in the iframe will not bring it to the front.
     
 Video Example: https://www.youtube.com/watch?v=gSXeCtJxqbs


Azmoria:
Update: 
• Reload button
• No title for consistency - if I can figure out how to consistently load the names of monsters/players maybe but I don't think titles are necessary at the moment.

image

cyruzzo:
great

PR #351, opened 2022-03-26
splash and console now tell you if you're in an unpacked version

bain2236:
version comes directly form manifest file

can't even push the right buttons in github.
draft because no FF support yet.


cyruzzo:
Like the idea.. but probably not worth  the effort

PR #359, opened 2022-03-29
Fixing the '24 minutes ago' bug

reidb:
'player connected' messages were using hard coded date and timestamps.  Changed them to use a similar method for adding date and time as later in the same function.


cyruzzo:
you shouldn't mix different stuff on a single PR (the fonts). That's probably because your sending me PR from your main. You should create a "feature branch" on your repo and send me a PR from that

reidb:
Fair point, I will do future PRs from a branch.  Reverted manifest commit, should be just the bug fix left now.

cyruzzo (approved):

PR #369, opened 2022-03-31
jitsi external lib update

hypergig:
just your run of the mill external lib update


cyruzzo:
stale.. killing this.. we will update when jitsi breaks.. or if firefox reviewers complains..

PR #374, opened 2022-04-01
mon image shows even when homebrew

bain2236:
found a weird bug with urls in which the large avatar url is saved onto the monster but it might be wrong. the last 2 sections of the url have to be 1000 otherwise it fails

shown it working
https://discord.com/channels/815028457851191326/815028804103307354/959469570110939156

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -808,6 +808,14 @@ function inject_monster_image(stat) {
 		console.groupEnd();
 		return;
 	}
+	// weirdly if either of the url numbers in these 2 positions aren't 1000 it throws a 403 error...
+	let splitUrl = stat.data.largeAvatarUrl.split("/")
+	if ([splitUrl.length -2] !== 1000 || [splitUrl.length -3] !== 1000){
+		splitUrl[splitUrl.length -2] = 1000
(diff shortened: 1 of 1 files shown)


cyruzzo:
I'm a bit scared that this could interfere with something else.. 
Are we sure that there always need to be that "1000" ?

bain2236:
So it's a weird one I'll give you that.
I went to the monster detail page (homebrew) opened the image in a new tab and it was 1000/1000. I dumped out the data in the stat object and it was 1000/1000. Once I changed it to be 1000/1000 it worked 🤷🏻‍♂️
Tried it on a few homebrew monsters and it worked. Happy to do another round of testing though.
Even if it does fail it will fall back to the small avatar image

bain2236:
So did some further testing and recorded it
https://user-images.githubusercontent.com/9283242/161733898-4017ebfa-a76c-4595-af0b-7da56b2a3072.mp4


https://user-images.githubusercontent.com/9283242/161733936-6f83ff88-5cf3-4b57-a24a-34c9f65d2495.mp4

https://user-images.githubusercontent.com/9283242/161734614-5e1cd30b-ee06-4548-9394-6d271d611fce.mp4

PR #378, opened 2022-04-02
Stop scroll when switching scenes and improve performance of scene panel

### LINKED ISSUE #358: The "scenes panel" should not close when you press DM or PLAYERS button
Currently, If you have to switch the scene for both DM and PLAYERS ..  you have to re-open the scenes panel after the first click.
It's probably better to keep the scenes panel open as switching both is the most common use case.

### LINKED ISSUE #384: perf issues in the scenes menu
My last game hard crashed the tab 3 times while in the scenes menu. Wasn't able to troubleshoot it in game but it seems to happen when you first open the scenes menu, but only sometimes. There have also been reports of issues that may also have been related to the scenes menu on [discord](https://discord.com/channels/815028457851191326/815028457851191329/960278786488819743).

Azmoria:
Closes #358 

I believe this also Fixes #384 

I believe the actual request once figured out on discord was to not have the scenes panel scroll up when selecting a new scene. So you can easily switch both players and DM.

https://user-images.githubusercontent.com/65363489/161395246-7b13a1ef-f946-434b-a021-0f047a4568a1.mp4

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -1086,7 +1086,6 @@ class MessageBroker {
 		}
 
 		if(window.CLOUD && window.DM){
-			refresh_scenes();
 			$("#combat_area").empty();
 			ct_load();
 		}
--- a/ScenesPanel.js
+++ b/ScenesPanel.js
@@ -852,56 +852,55 @@ function refresh_scenes() {
 		controls.append(delete_button);
 		newobj.append(controls);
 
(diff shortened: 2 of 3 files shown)


Azmoria:
Performance video example

https://www.youtube.com/watch?v=PPlDuNLCC2g

bain2236:
No issues for me performance wise and the scene panel does stay open when changing players/dm to a different scene

PR #383, opened 2022-04-02
inspiration now comes from player sheet

### LINKED ISSUE #347: temp hp aura not updating
aura updates for the player but doesn't appear to sync with DM.
Sometimes a token has their temp hp added to their current hp and sometimes not. Not been able to find out why there's an inconsistency

https://user-images.githubusercontent.com/9283242/160240358-2f411fc1-f231-4c88-83d6-5c3675ac060f.mp4

### LINKED ISSUE #367: synchronize "inspiration" marker on player tokens
When a character has inspiration, the token should automatically get the inspiration marker applied. Similarly, the token marker should be removed when the character removes inspiration.

bain2236:
temp hp has an identifier (dual aura when aura enabled. or hp bar colour change if not)
Believe this resolves #347 
resolves #367 

https://user-images.githubusercontent.com/9283242/161400060-741dbac1-f914-4a03-b654-008afb4ff260.mp4

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -805,6 +805,7 @@ class MessageBroker {
 	}
 
 	sendTokenUpdateFromPlayerData(data) {
+		console.group("sendTokenUpdateFromPlayerData")
 		if (data.id in window.TOKEN_OBJECTS) {
 			var cur = window.TOKEN_OBJECTS[data.id];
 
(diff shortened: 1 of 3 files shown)


bain2236:
I've poked around at this to see if it resolves issues of temp hp not updating hp aura's correctly. It appears that occasionally a players token hp is it's hp + the temp hp, but other times not which causes the colour of the aura to not be displayed consistently. I've removed the "fix <issue number>" from this PR as it was never really intended to fix said issue. The dual colour temp hp aura is just a bonus ontop of the inspiration showing as I was updating all the relevant code sections to be able to visibly display temp hp on a token

vlaminck:
It looks like you have a few other PRs merged into this one. Am I seeing that correctly? There are changes to 19 files, and a lot of them don't seem to have anything to do with your solution.

bain2236:
😢 will resolve tomorrow

bain2236:
Git issues resolved

vlaminck (approved):
very nice!

PR #386, opened 2022-04-04
Fix for slow dice

### LINKED ISSUE #385: Dice roll slowly
The dice rolled anywhere not in an iframe roll very slowly. This is caused by a timeout in the message broker

Azmoria:
Fixes #385 

Reduced the timeout to 1 and it fixes the slow dice.

https://user-images.githubusercontent.com/65363489/161475576-7c35b97a-68f1-468f-89f5-107b85c92635.mp4

Note: since the changes they made to encounters and the character sheet my dice roll dummy fast like this on all of dndbeyond.

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -610,7 +610,7 @@ class MessageBroker {
 				    ((window.DM && msg.data.context.entityType === "user") || window.PLAYER_ID == msg.data.context.entityId)) {
 					console.log("Hooking requestAnimationFrame for dice roll");
 					this.origRequestAnimFrame = window.requestAnimationFrame;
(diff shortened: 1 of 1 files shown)


bain2236:
Tested this out and couldn't find any issues with rolling dice. I also have turbo fast dice now, which isn't a bad thing. Even bumping the quality of dice + shadows + hardware acceleration they're rolling really fast still. Tested a few rolls with 30 dice and it was really responsive. Minor slow down in the first 1 second of clicking roll but that's dndb creating the dice and putting strain then it's smooth as butter

cyruzzo:
I've done things differently by just removing the code that messes with requestAnimationFrame.. big thanks for troublshooting things anyway :)

PR #390, opened 2022-04-05
Feature/302 update rpgdiceroller to be more dndb

bain2236:
More could be done here but I didn't do it.. because I think the only thing that's left is some monster stat block stuff? and as players currently can't own a monster (as in right click a token and view sheet) I didn't do it (there is a PR lurking around for that though which I don't mind sorting out much as I've sorted this stuff.

- rolling via rpg dice roller in extra tab now looks like dndbeyond
- reworked extra tab buttons to add in the correct info for action / type
- chat box rolling now looks like dndb
- added help message to chat box
- added /help command to open dice notation documentation
- added custom action/types to chat box rolling

chat box rolling :
https://user-images.githubusercontent.com/9283242/162044133-42c5f310-3bd3-4c4e-8ba6-d770f5ba471a.mp4

Monster dice rolling
https://user-images.githubusercontent.com/9283242/162044321-9f218dc7-50ce-4bd0-ac82-e8497a3950b5.mp4


bain2236:
Just came across a section of the rpg dice roller documentation that won't work with these current changes so I'll need to make another change. Will get around to it sometime this week hopefully
specifically the group rolling interacts badly with the adding custom actions/types via the chat

bain2236:
updated and tested a different way to get action/types working via command line. opting for use of `-a=/--action=` and `-t=/--type=` respectively rather than using a json esque structure. rolls without actions/types still default to display custom:custom

bain2236:
Pulling this down as I'm making updates to it

PR #394, opened 2022-04-09
Tokens Panel Unification

### LINKED ISSUE #287: Drag and Drop for monsters no longer works
I think the issue is related to dragging from an iframe onto the main document.

### LINKED ISSUE #305: Add number and letter tokens to the builtin token list
natemoonlife has been kind enough to make some for us. We should include them in the app.

https://discord.com/channels/815028457851191326/851179618991472699/931627791449993266

### LINKED ISSUE #353: As a DM, I want to add monsters from an encounter that I've created
1. This would require us to display a list of encounters. (probably filtered to the current campaign, but I could see people wanting to run an encounter from a different campaign as well)
2. Clicking a listed encounter would need to load a list of monsters in that encounter similar to other tokens in the sidebar.(I'm sure people would also want to see the details of the encounter)
3. Each monster in the encounter would need to support drag and drop and plus button clicks like other tokens in the sidebar.

### LINKED ISSUE #392: Unify token tabs
Right now we have a player tokens tab, a monster tokens tab, and a custom tokens tab. It would be nice if they were combined into a singular unified tab "Tokens" tab.

vlaminck:
Closes #392 
Closes #305
Closes #287 
Closes #353 

# Features
1. Added `window.ajaxQueue.addDDBRequest` for conveniently queuing authenticated requests to DDB
a. Updated EncounterHandler to use this
2. Migrated `tokendata` to `mytokens`
a. This greatly simplifies the data structure from nested objects to a flat list
b. The keys in the object drop the `data-` prefix since that was a hold over from the old `token_button` function which no longer exists
c. builtin tokens have also been migrated to this new structure
3. Adds numbers and letters provided by natemoonlife to the builtin tokens
4. Adds a reusable folder system that displays a list of `SidebarListItem` objects.
a. These objects are essentially "ViewModels" that represent some backing data object. 
b. In this PR, they are various types of tokens, but this can easily be extended for use with scenes, audio, or anything that we want to list in a folder structure.
5. Creates a new "Tokens" Panel that unifies the "Players", "Monsters", and "Custom" tokens panels
a. This is only for the DM. Players still get the old "Players" panel
6. Adds search for all tokens

# Demos
## 3. Adds numbers and letters provided by natemoonlife to the builtin tokens
avtt-builtin-numbers-letters

## 4. Adds a reusable folder system that displays a list of `SidebarListItem` objects.
https://user-images.githubusercontent.com/584771/162590300-fdf1303d-b74d-42e2-9f7d-2b8e38c6df49.mov

### Player tokens
avtt-tokens-panel-pcs

### My Tokens (formerly "custom tokens")
avtt-tokens-panel-mytokens

### Builtin Tokens
avtt-tokens-panel-builtin

### Monster Tokens
avtt-tokens-panel-monsters

## 6. Adds search for all tokens
avtt-tokens-panel-search

### Monster Tokens Filtering
Monster Tokens Filtering


vlaminck:
This will benefit from https://github.com/cyruzzo/AboveVTT/pull/408
The buttons that we inject into the monster stat blocks will be able to call `DiceRoller`

Azmoria:
From what I can tell everything seems to work great with the token menu. Is there a limit to the tokens placed when using `place tokens` on a folder? Worried about breaking peoples scenes if they try and it adds far to many tokens. This might be a non-issue though.

Absolutely love the encounters, being able to see them add the tokens it's great. So is searching everything. I haven't tested all the cleaned up stuff but from what I've played around with so far I haven't had any issues.

Only issues for me: 

1) It doesn't populate anything until I search or filter in the monsters folder. After backing out of the search or filter it populates the stats for the tokens on the map. 

2) If I have 3 ogres on the map it populates 3 ogres. Do we want that or should it just be one of each type?

Not major issues though. Awesome job this is a huge improvement imo.

https://user-images.githubusercontent.com/65363489/166119686-be62744c-2b4b-4109-8c91-0d923d693d6b.mp4

vlaminck:
> From what I can tell everything seems to work great with the token menu. Is there a limit to the tokens placed when using `place tokens` on a folder? Worried about breaking peoples scenes if they try and it adds far to many tokens. This might be a non-issue though.

I don't limit the number of tokens, but I prompt if you try to add more than 10 at a time (was 20, but I just changed it to 10). I wasn't sure if we should put a hard limit on it or not, but we easily could.







> 1. It doesn't populate anything until I search or filter in the monsters folder. After backing out of the search or filter it populates the stats for the tokens on the map.
>
> 2. If I have 3 ogres on the map it populates 3 ogres. Do we want that or should it just be one of each type?

I thought that populating the list with the monsters that are currently on the scene might be nice, but I guess that really only matters if you're going to add more of the same monsters... I just thought it might remove the need for searching the same monsters over and over again. I can remove that if we think it's not worth doing or if it's at all confusing.  
The blank list on first load is probably a race condition. I'll figure that out quick.  
The duplicates is also not intended in the monsters folder. It is intended in the Encounters folders, though. 






> Edit: I just realized based on your examples above this might not be the intended behavior at all. It looks like it should be loading the whole monster list. Unless you have a load of tokens to test this.

The example I posted in the description was made before I started populating the list with monsters on the scene.





> Edit 2: I'd love to see encounter groups have an add button separate from the whole encounter. Would be sweet for fights with waves/monsters showing up after x number of rounds. 

This is something I'd like as well, but I wasn't sure about the UI for it. There are some followup UI improvements that I think we could make that aren't necessary, but would be nice to have. This is one of them, and another would be grouping multiple monsters with some sort of "x4" on it if there were 4 monsters of that type. I decided not to do that yet because the way that DDB structures the monster grouping is odd, and I didn't want to figure that out yet.

vlaminck:
I've reverted the monster folder logic where it displays what is currently on the scene. The monster folder now displays the first 10 monsters if there is no search term just like the DDB encounter builder does.

bain2236:
I love this! it's so cool
Few things super minor things though

1. folders are supposed to have the move cursor?
2. making a crazy depth level of folders eventually trims  the name down. maybe an on hover of the name
3. can we get folders to show up with search as well?

Azmoria:
> I thought that populating the list with the monsters that are currently on the scene might be nice, but I guess that really only matters if you're going to add more of the same monsters... I just thought it might remove the need for searching the same monsters over and over again. I can remove that if we think it's not worth doing or if it's at all confusing. The blank list on first load is probably a race condition. I'll figure that out quick. The duplicates is also not intended in the monsters folder. It is intended in the Encounters folders, though.


I don't mind it populating with the monsters on the scene. I see you reverted it and it's fixed the empty folder bit. The empty folder was what was confusing me - makes sense if it's a race condition. As far as the rest of it for sure it can be dealt with after.

I think this is awesome.

vlaminck:
> 1. folders are supposed to have the move cursor?

Good catch! This was held over from the old way. None of the images should have that. Fixed


> 2. making a crazy depth level of folders eventually trims  the name down. maybe an on hover of the name

Good call. Fixed.


> 3. can we get folders to show up with search as well?

Do you mean if a folder is named "goblinoids" and you search "goblin", then the "goblinoids" folder would show up as well?

vlaminck:
> I don't mind it populating with the monsters on the scene. I see you reverted it and it's fixed the empty folder bit. The empty folder was what was confusing me - makes sense if it's a race condition. As far as the rest of it for sure it can be dealt with after.

After you brought it up, I noticed that the behavior is a bit odd so I reverted it. There have been requests for listing all tokens on the scene so I was thinking about adding a new root folder for "On The Scene" (open to naming suggestions here) that lists out ever token on the scene with a "locate" button.

bain2236:
Exactly what I meant on the folder names. Even if goblinoids didn't actually have any goblin tokens in. Sometimes it's easier to search for a folder name. Like I have a folder called clutter which is just some stuff I throw down ontop of the background for spice. Haven't a clue what's actually in it

vlaminck:
> Exactly what I meant on the folder names. Even if goblinoids didn't actually have any goblin tokens in. Sometimes it's easier to search for a folder name. Like I have a folder called clutter which is just some stuff I throw down ontop of the background for spice. Haven't a clue what's actually in it

Good call. Done

bain2236:
Works as expected, cleans up so much, adds so much value

PR #399, opened 2022-04-16
Make our dice rolling look like dndbeyond, includes rolling via chat box and player monster frames (extra tab/monster sheets). Added actions, types and more commands to chatbox rolling. Added "X/each" spell slot trackers to monster sheets

### LINKED ISSUE #302: Update rpgDiceRoller formatting to look more like DDB dice rolls
This was added a while back, but was disabled because it was missing a couple of features. The code still exists in `Main.js` `function send_rpg_dice_to_ddb`. 

These are the reasons it was disabled:
1. It displays the DM image, but it should display the monster image.
2. The "send to everyone" button doesn't always show up. (this might actually be unrelated, and would instead be handled by #272 )
3. "To Hit" information was missing.

bain2236:
- Fixed up RPG dice roller to make it look like dndbeyond dice rolls
- Added custom actions/types via chat box
- Added helper text and /help command to chat box that opens the docs on dice rolling
- Uplifted dice buttons on monster sheets that aren't from encounters (now that players can have access to a monsters sheet)
resolves #302 

Some more user readable notes:

- Rolling via chat now has more commands and those rolls look like DNDBeyond rolling
- Rolling from extra tab now shows as DNDBeyond would if they even allowed rolling from there..
- Buttons on the monster sheet owned by players more closely align with the standard DDB styling and they also have all the information that you would expect in the combat log when rolled
- Spell slots trackers for for "X/each"
- Video.. apologies for the super quiet audio, youtube has clamped it down for whatever reason will get better at this :D https://www.youtube.com/watch?v=w4HftubO1ek


vlaminck:
This will benefit from #408

vlaminck:
This is great! 

Two things I found:
1. the `rollType` should default to "roll" instead of "custom". If you roll DDB dice, this is what they do
2. Firefox displays the hover text below the chat input which is impossible to read. It then seems to push the entire gamelog up for some reason.

Two things I think would be worth changing, but not necessary:
1. Because we will soon be using the DiceRoller introduced in #408 I think we should lose the `/help` command
2. Would it be possible to specify the action and rolltype without the quotes? so `/r 1d4 action:rollType` instead of `/r 1d4 "action:rollType"`. If this is too hard, then it's not worth doing, but it would be a really nice qol improvement now that I've played with it.

vlaminck (changes requested):
This is great! 

Two things I found:
1. the `rollType` should default to "roll" instead of "custom". If you roll DDB dice, this is what they do
2. Firefox displays the hover text below the chat input which is impossible to read. It then seems to push the entire gamelog up for some reason.

Two things I think would be worth changing, but not necessary:
1. Because we will soon be using the DiceRoller introduced in #408 I think we should lose the `/help` command
2. Would it be possible to specify the action and rolltype without the quotes? so `/r 1d4 action:rollType` instead of `/r 1d4 "action:rollType"`. I think this would be a really nice qol improvement now that I've played with it. This regex seems to work by finding everything after the last number `const commandRegex = /([^\d]+$)/;`

bain2236:
Nice nice, think I was just being lazy with updating the action types for cli input and between quotes seemed like the easiest option. I'll download ff and poke around with the issue mentioned in the morning

bain2236:
https://user-images.githubusercontent.com/9283242/164969068-82b2a2ec-2da6-46d3-9e3f-7efb4c32a14b.mp4

Tested in chrome, left, FF right.

bain2236:
User notes:

- Rolling via chat now has more commands and those rolls look like DNDBeyond rolling
- Rolling from extra tab now shows as DNDBeyond would if they even allowed rolling from there..
- Video.. apologies for the super quiet audio, youtube has clamped it down for whatever reason will get better at this :D https://www.youtube.com/watch?v=w4HftubO1ek

vlaminck (approved):

vlaminck:
This has been merged into a feature branch and a new PR has been opened to replace this one https://github.com/cyruzzo/AboveVTT/pull/462

PR #400, opened 2022-04-16
remove campaign id and secret from url

### LINKED ISSUE #297: Remove query params from url
Currently, the join link id is in the url. We probably shouldn't expose that. 

To do this, we will either need to write it to localStorage or be able to fetch it from the cloud. I think fetching it from the cloud would be ideal, but we need to figure out how to do user-specific cloud interactions first. Currently, it is all done using the join link key.

vlaminck:
Instead of setting `cid` and `cs` in the url, they are written to `localStorage`. There are ways to fetch the campaign id from the DDB API, but there's no way to fetch the join link that I know of. If we stop using the join link, we can change this to not write to `localStorage` as well. But at least this info is no longer in the url for everyone to see.

Closes #297


Azmoria (approved):
Works as expected - didn't run into any issues.

cyruzzo:
uhmmm..  not super-sure about having a "single key".
Maybe it's better to store something that maps a campaign id to it's secret.

Something like localStorage.setItem("CampaignSecret1671234","1671234123412341234");

cyruzzo:
closing to cleanup the queue.. reopen if needed

PR #401, opened 2022-04-16
Bugfix - context menu reverting token options, health of tokens not right

### LINKED ISSUE #405: bugs for 0.7.42 pre-release (aka main brain 0.7.41)
Found a few bugs lurking within main before the next release of 0.7.42

- pressing configure on a token resets it's previously configured options
- non monster tokens have a health bar when "disable stat" is off

bain2236:
@vlaminck confirmed that there was issues with a merge see https://discord.com/channels/815028457851191326/930926587392692335/964934545629663283
in regards to the context menu reverting changes.
This has now been resolved

Also sorted out all tokens having hp bars on them. By actually just adding in that feature, when tokens have "disablestat" disabled a DM can add hp/ac/elevation to a token via the context menu / by adding hp to the tokens hp bar
resolved issue where hidestat was also hiding for DM
fixes #405

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -452,7 +452,7 @@ class Token {
 		hpbar.append(hp_input);
 		hpbar.append(divider);
 		hpbar.append(maxhp_input);
-		if (this.options.monster > 0) {
+		if (!this.isPlayer()) {
 			hp_input.change(function(e) {
 				hp_input.val(hp_input.val().trim());
 				self.update_and_sync(e);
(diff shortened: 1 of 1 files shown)


bain2236:
tokens such as helper tokens are appearing to the DM with hp stats etc. so only add hp stats etc when the token has a hp

bain2236:
There is an issue with this when it comes to AOE tokens. It's not a big one, just means that creating an AOE token is applying some additional options that you might not want.. like it having a health bar. Not 100% how this is happening or whether it's pulling what options to use from the "new token" configuration menu thing.
Personally I'd say this minor issue than an infuriating one that this PR resolves (tokens reverting their options)

PR #409, opened 2022-04-19
Add scalable images and custom border color for tokens

### LINKED ISSUE #361: Allow token border color to be customized
DMs should be able to change any token’s border color. Players should be able to change the color of their own token’s border.

### LINKED ISSUE #459: Ping color should match token border color
It would be helpful it ping colors were the same color as the player token color so everyone knew who sent the ping.

Azmoria:
Adds an option to scale the token image separate from the token itself.
The HP Aura, Border, and new temp hp aura snagged from Bain's PR all fit the shape of the token. 

This only updates the token when hitting enter/leaving focus of the input or mouse up on the slider. This would also preps/allows for actual HP bars if we wish to use them see example image at the bottom (not in this PR)

Example video with menu: https://www.youtube.com/watch?v=sgD8cJJSpAg

This allows for more token options and is a QoL improvement for custom token usage.

EDIT: Partially fixes #361 

EDIT 2: Closes #459 see comment below

Not in this PR. These HP bars are something we can do if we choose with a line or two of CSS now: 
https://user-images.githubusercontent.com/65363489/163916570-48f7baa4-fedc-473d-8f41-d746f0bc40cd.png
As a suggestion it may help with color blind users - it may be easier to see a depleting bar then the auras. It might be a QoL option we can add.


Azmoria:
Now also includes custom border colors which partially fixes #361

vlaminck (changes requested):
It looks like conditions and markers also get the border and health aura applied. Is there any way to prevent that?
Screen Shot 2022-04-23 at 1 21 59 PM

Azmoria:
> It looks like conditions and markers also get the border and health aura applied. Is there any way to prevent that? Screen Shot 2022-04-23 at 1 21 59 PM
> 
> It actually helps them stand out more so I don't hate it. If it would be too hard to change, I'm ok with it.

This is fixed. I was a missing `>` in css lol. I think a static drop shadow instead might be better for having them stand out if we prefer.

vlaminck (dismissed):

Azmoria:
Standard images with default (1.0) scaling should now be appropriately within their bounds. As seen in the right part of the comparison image below.

image

vlaminck (dismissed):
Excellent work! Works for me in both Chrome and Firefox

bain2236:
This is great fun. Becaue a giant bandit captain with 2 bandits on it's shoulder with a pink border is what everyone needs in their game right?
Great stuff
image

vlaminck (approved):

Azmoria:
I've added ping color matching border color. I also adjusted the ping animation to be smoother (which results in it being on screen for a shorter period but I think this is a good thing personally). I can split this back out again if necessary. This seemed like the easiest way to include it since the PR is still open.

Closes #459 

Example video 1 (before animation changes):
https://www.youtube.com/watch?v=7bSeH2RpA_A

Animation example:
https://user-images.githubusercontent.com/65363489/166393949-2f3338ff-09d6-4ea2-931c-4de2573aba77.mp4

Azmoria:
Fixed ignore aspect ratio. It wasn't functional with the updated centering. It should now work again.

https://user-images.githubusercontent.com/65363489/166844957-912c75ce-8f2c-4143-988c-4419fccd6a5b.mp4

Azmoria:
In #470 closing this. References above still apply.

PR #414, opened 2022-04-21
Fix: The remove from combat button in the combat tracker should now update the token combat status properly.

### LINKED ISSUE #413: Bug: Combat tracker remove button does not update token combat status
When using the combat tracker button to remove a token from combat it does not update the combat status of the token. As such the context menu does not update.

Azmoria:
The remove from combat button in the combat tracker should now update the token combat status properly. 

fixes #413


vlaminck (approved):

jordan-da (approved):
look cool to me

PR #418, opened 2022-04-21
Fix: Repush the fix for monster roll syncing with default rolls.

### LINKED ISSUE #417: Bug: A merge commit removed the line locking monster sidebar open
A merge commit removed the line locking the monster sidebar open. This broke send to default syncing. 

As a side note - it looks like dndbeyond is planning on adding DM to defaults. It was there for a bit but then the took it down cause it broke stuff.

Azmoria:
This should fix the sync issues with monster rolls again. This locks open the sidebar so the default menu is "clickable"
Fixes #417

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -196,6 +196,9 @@ class EncounterHandler {
 				remove_combat_tracker_loading_indicator();
 		}
 		minimize_monster_window_double_click($("#resizeDragMon"));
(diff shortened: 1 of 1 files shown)


vlaminck (approved):

jordan-da (approved):
I only partly understand what this fixes but still looks ok to me

PR #420, opened 2022-04-22
Fix: Get Super Mega Wizard to appear again when clicking the button.

### LINKED ISSUE #419: Bug: Super mega wizard not working
The super mega wizard is not opening. 

This is due to an error being thrown on a check to see if a token is a player token on an undefined id.

Azmoria:
This fixes #419 

This changes the check if player id function to return false on an undefined id.

This fixes the super mega wizard due to at some point it running that function when creating a new token and running build_hp. Tokens might be the handles you drag around. I'm not sure but this fixes it.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -452,7 +452,29 @@ class Token {
 		hpbar.append(hp_input);
 		hpbar.append(divider);
 		hpbar.append(maxhp_input);
-		if (!this.isPlayer()) {
+		if(self.options.id != undefined) {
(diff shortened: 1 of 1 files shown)


vlaminck (dismissed):

vlaminck (approved):

jordan-da (approved):
🍕

jordan-da (approved):

hypergig (approved):
another approval, this time with my non-work account

PR #422, opened 2022-04-22
Fix: Reload maps on manual grid change so grid updates are displayed properly

### LINKED ISSUE #421: Bug: Maps don't refresh to display grid when manual grid changes are saved
When you click the save button in the manual grid updates the views do not refresh to display the updated grid.

Azmoria:
Fixes #421 
This will reload maps if selected on manual grid change so grid updates are displayed properly. This will only reload the maps if the player or dm is actually on the map that is changed.

It accomplishes this by clicking the selected maps dm or player button if the buttons scene ID matches the ID of the edit menu's scene.

### DIFF
--- a/ScenesPanel.js
+++ b/ScenesPanel.js
@@ -217,13 +217,15 @@ function edit_scene_dialog(scene_id) {
 			console.log('setto ' + n + ' a ' + $(this).val());
 		});
 		if(window.CLOUD){
-			window.ScenesHandler.persist_scene(scene_id,true,true);
-			if($(".player_scenes_button.selected").parent().parent().attr("data-scene-index") == scene_id) {
-				$(".player_scenes_button.selected").click();
-			}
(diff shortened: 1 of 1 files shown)


ScenesPanel.js
```diff
@@ -217,7 +217,15 @@ function edit_scene_dialog(scene_id) {
 			console.log('setto ' + n + ' a ' + $(this).val());
 		});
 		if(window.CLOUD){
-			window.ScenesHandler.persist_scene(scene_id,true,true);
+			window.ScenesHandler.persist_scene(scene_id,true,true);						
+			setTimeout(function() {
```
hypergig, line 221: @Azmoria is it possible to hook into an event instead of an arbitrary timer? I feel like we bump into a few bugs caused by these sorts of race conditions.

Azmoria:
It works without 200 but not consistently for me. 

I agree with both of you - I want to hook into after the persist stuff but having trouble. I'll keep working at it.

PR #424, opened 2022-04-22
Add: Shortcut (Ctrl+right click) To open both the context menu and extended token configuration.

### LINKED ISSUE #423: Request: Add shortcut to extended token options
Add a mouse shortcut to get to extended token options.

Azmoria:
Closes #423 
Right click will open the context menu as usual.
Ctrl + right click will open both the context menu and extended configuration options.

I think this is a happy middle ground for spliting options while giving a bit of QoL boost to those who wild shape, polymorph etc. a lot. I think this will be more and more useful if/as we add more to the configure menu.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -1232,12 +1232,6 @@ class Token {
 				draw_selected_token_bounding_box(); // update rotation bounding box
 			});
 
-			tok.mouseup(function (e) {
-				if (e.which === 3 && (e.ctrlKey || e.metaKey)) {
-				    token_context_menu_expanded([tok.attr("data-id")]);			    
-				}
-			});
-			
 			console.groupEnd()
 		}
 		// HEALTH AURA / DEAD CROSS
@@ -1857,6 +1851,10 @@ function token_menu() {
 		selector: '.VTTToken',
 
 		build: function(element, e) {
+
+			if (e.which === 3 && (e.ctrlKey || e.metaKey) && window.DM) {
+				token_context_menu_expanded([element.attr("data-id")]);			    
+			}
 
 			if ($(element).hasClass("tokenselected") && window.MULTIPLE_TOKEN_SELECTED) {
 				if (!window.DM) {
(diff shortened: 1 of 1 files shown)


vlaminck (changes requested):
I like this approach, but I found a couple of issues with it. 

1. We need to restrict it to `window.DM`. As a player, I can get access to the configure menu, and mess around with the configuration options.
2. This only works if I ctrl+right click while the context menu is already open. I think the contextMenu library is hijacking the right click and calling stopPropagation. Could you check for the `ctrl` key in where we build the contextMenu instead?

Azmoria:
Thanks for the suggestion and catch with the players. This makes much more sense. I didn't realize the contextmenu had an event attached that's very useful.

vlaminck:
> Thanks for the suggestion and catch with the players. This makes much more sense. I didn't realize the contextmenu had an event attached that's very useful.

Yeah, it's in Token.js `token_menu()` and uses this https://swisnl.github.io/jQuery-contextMenu/

Azmoria:
> > Thanks for the suggestion and catch with the players. This makes much more sense. I didn't realize the contextmenu had an event attached that's very useful.
> 
> Yeah, it's in Token.js `token_menu()` and uses this https://swisnl.github.io/jQuery-contextMenu/
> 
> so maybe try it in the `build` function there? I honestly don't know if that will work, but it's worth a try

Yeh I already moved it there. It works for me at least. Let me know.

vlaminck (approved):
Works like a charm!

bain2236:
👍

vlaminck:
After recent discussions in Discord, do we still want to go this route? Or would it be best to just build a singular token configuration window? I think we were leaning toward the singular menu

Azmoria:
> After recent discussions in Discord, do we still want to go this route? Or would it be best to just build a singular token configuration window? I think we were leaning toward the singular menu

I think that was the general consensus and best in tune with the feedback. I can take this down or we can offer it as a temporary qol improvement until we can build better.

vlaminck:
I think this is no longer needed now that @Azmoria and I have started rebuilding the token context menu.

PR #425, opened 2022-04-22
Fix: When editing a scene clicking save should now refresh the scene if it is active. It will no longer require you to manually reload the scene for players or dm's.

### LINKED ISSUE #421: Bug: Maps don't refresh to display grid when manual grid changes are saved

Azmoria:
This will fix #421 

It will refresh the viewed scenes on saving a scene edit such as the grid or maps. 

Thank you so much Mikedave for the digging and help with finding the solution. Put up a new PR to keep it cleaner.

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -50,8 +50,12 @@ function get_canvas_max_area() {
 class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 
 	reload(callback = null) {
-		this.switch_scene(this.current_scene_id, null);
-	}
+        if (window.CLOUD) {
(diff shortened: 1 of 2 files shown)


vlaminck (approved):

jordan-da (approved):
nice

Azmoria:
unfortunately this also runs into the timing issue and doesn't work for maps updating =(. I think the only good solution is a change to the messaging to/from the cloud so we can get a message back telling us to reload the scenes. 

The timeout one could work as a temporary fix though.

Azmoria:
This should now be fixed by #426 

Thanks Mikedave for going through that and fixing it.

PR #428, opened 2022-04-23
Add: In Player view monster stat blocks are now moveable/resizable

### LINKED ISSUE #427: Request: Player monster frames moveable/resizable
A request to have the monster frames moveable and resizeable like the DM frames.

Azmoria:
Closes #427

This makes the player monster stat blocks moveable/resizable as well instead of static and covering most of the player area.

Edit: example video https://www.youtube.com/watch?v=-SjtevefyCY

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -468,8 +468,11 @@ function load_monster_stat(monsterid, token_id=false) {
 	}
 
 	let container = $("<div id='resizeDragMon'/>");
-		
-	
+
+	if(!window.DM && $("#site #resizeDragMon").length>0){
+		$("#resizeDragMon iframe").remove();
+		container = $("#resizeDragMon");
+	}
 	//container.css("width","900px");
 	let iframe = $("<iframe>");
     // UGUALE A COMBAT TRACKER INSIDE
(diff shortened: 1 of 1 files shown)


vlaminck (commented):
Not sure if there's anything we can do about monsters that players don't own.

Also looks like the close button has issues in firefox. That might just be a bug in `main` as I've seen it on other PRs, too.

Screen Shot 2022-04-23 at 3 11 35 PM

vlaminck (changes requested):
Looks like we're not removing some things for players

Screen Shot 2022-04-23 at 3 14 26 PM

Azmoria:
Yeh the styling issues are in main I'm working on them now.

 Should we do anything about the unknown monsters? And the footer I assume? Both these issues are probably also in main.
 

All I did in this PR is put the current iframe into my moveable div. If we need to do something about these we should do it before this beta goes live.

vlaminck:
> Yeh the styling issues are in main I'm working on them now.
> 
> Should we do anything about the unknown monsters? And the footer I assume? Both these issues are probably also in main.
> 
> All I did in this PR is put the current iframe into my moveable div. If we need to do something about these we should do it before this beta goes live.

You are correct. Those are bugs in `main`

vlaminck (approved):

PR #430, opened 2022-04-23
Fix: for drag map lag in player view

### LINKED ISSUE #429: Bug: Dragging Map as a player isn't smooth.
This is caused by a DNDBeyond event listener on scroll that throttles scroll speed.

https://user-images.githubusercontent.com/65363489/164943189-d759777d-5c1d-4d99-a208-374d615acc79.mp4

Azmoria:
Fixes #429

This is setup to listen to the event first then stops the propagation of the throttle scroll event listener.

edit: Better comparison videos. It is more pronounced on my monitor than in the videos below but I can still see a difference in these with regards to smoothness.

Before:
https://user-images.githubusercontent.com/65363489/164948832-db056c26-d2f6-42d0-898e-aeb521201328.mp4

After:
https://user-images.githubusercontent.com/65363489/164948834-b4034f06-ae0f-4bc2-8cef-b365e44eb1ed.mp4

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -800,6 +800,9 @@ function init_splash() {
 	setTimeout(function() {
 		init_loading_overlay_beholder();
 	}, 0)
+	window.addEventListener("scroll", function(event) {
+	   event.stopImmediatePropagation();
+	}, true);
 }
 
 // UNIFIED TOKEN HANDLING


vlaminck (approved):
I can't actually reproduce the bug on my machine, but this has no ill effects for me

PR #435, opened 2022-04-24
Feature/add text to maps

bain2236:


bain2236:
2 issues to resolve since merging. Ruler doesn't do anything. aoe tokens are spawning in strangely, image is in top left. but selecting said image then puts a bounding box around where the token should be in the middle of the vtt

bain2236:
Should be good enough for other devs to poke holes in 😅

bain2236:
actually spotted some more things that I should sort out first
- [ ] being on a ruler still allows you to select a token, so you can't measure from the token which is the main use case for measuring
- [ ] think about how measuring can be done while dragging a token. Possibly via holding alt while already selecting a token

bain2236:
- [ ] Styling of new text settings popup (mostly done)
- [ ] Plumb the settings into all text boxes
- [ ] Redo measure tool to behave like this
_Not sure if github can handle sub checkboxes._
- - [ ] Pressing "ruler" then left click on overlay creates the first point in the measure, click again create another, double clicking ends the measuring
- - [ ] Holding alt performs "quick measure" it drops a point where your cursor is on the overlay (if it's on the overlay) clicking again will drop another point as per usual. double click or Alt up will stop drawing measuring
- - [ ] Try again to get dragging working. Thinking holding down alt while over a token doesn't initiate the draw, but sets up what is required to draw, then clicking on the token will "pick it up", clicking while it's picked up will drop a waypoint. double clicking or releasing alt will drop the token and finish the draw

PR #438, opened 2022-04-26
Fix: Combat tracker clear button sets combat status

### LINKED ISSUE #439: Bug: Combat tracker clear button is not setting combat status
Similar to the delete button issue that was fixed the clear button does not set combat status. 

When hitting clear the token context menu to add to combat does not update and still shows remove from combat.

Azmoria:
This now updates the token combat status properly when hitting clear.

Fixes #439


bain2236:
Does the job 👍

PR #449, opened 2022-04-29
Add option for hp inputs only being shown on hover or token select

### LINKED ISSUE #448: Request: Add an option for HP showing on hover, selected, focused
Request to add an option that has the `.hpbar` only visible on hover with the following requirements that Mikedave, Bain and I discussed.

- Have the fade delay be approximately the same as the name
- Have the hpbar be visible if the token is selected, the inputs focused or the token is hovered.

Azmoria:
Closes #448 

This adds an option that when enabled will:

Have the hpbar be visible if the token is selected, the inputs focused or the token is hovered.
Have the fade delay be approximately the same as the name

Example video: https://www.youtube.com/watch?v=b0Exi8hIz6k

Edit: Thought I spotted a bug so went draft but it was user error 😂

### DIFF
--- a/Settings.js
+++ b/Settings.js
@@ -137,6 +137,12 @@ function init_settings(){
 			disabledDescription: "New player tokens will have their HP/AC visible to other players. Each player will be able to see their own HP/AC as well as HP/AC of other players."
 		},
 		{
+			name: 'hidehpbar',
+			label: 'Only show HP values on hover',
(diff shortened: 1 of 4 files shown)


vlaminck:
I love this! ❤️

bain2236:
This really helps declutter the ui without impeding the DM adjusting hp values. Also applies to the player view which I also like

Azmoria:
Will reopen when I covert it to the new context menu.

PR #455, opened 2022-05-01
HP in combat tracker is hidden when player hp is hidden on the token

### LINKED ISSUE #454: Combat Tracker HP display should mirror token HP display
If a token has HP/AC hidden, then the combat tracker should also respect that setting

### LINKED ISSUE #461: Bug: Max hp is not updating in the combat log.
Bain reported max hp was not updating in the combat tracker. There is also issues with player view updating.

Azmoria:
Closes #454 
Closes #461 - see comment below

Example video: https://www.youtube.com/watch?v=H4Qre_aKEM4

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -320,7 +320,7 @@ function ct_add_token(token,persist=true,disablerolling=false){
 			entry.append($("<td/>").append(hp));
 		else
 			entry.append($("<td/>"))
-		max_hp=$("<div/>");
+		max_hp=$("<div class='max_hp'/>");
 		max_hp.text("/"+token.options.max_hp);
 		max_hp.css('font-size','11px');
 		//max_hp.css('width','20px');
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -324,6 +324,10 @@ function ct_add_token(token,persist=true,disablerolling=false){
 		max_hp.text("/"+token.options.max_hp);
 		max_hp.css('font-size','11px');
 		//max_hp.css('width','20px');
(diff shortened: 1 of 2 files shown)


vlaminck (approved):
Works perfectly!

Azmoria:
Bain mentioned max hp wasn't updating and I went to fix it noticed there would be a conflict with this. I went to remove the conflict and noticed most of the code I would be reusing from this and it was a one line change to add here after fixing the conflict. 

This will also fix #461

vlaminck (approved):

bain2236:
nice!
I did come across this which may be completely unrelated. that elf token has all configuration off. So should show the hp/ac right? it's not a player token it's a custom token. Tried with a monster token too.
Seems weird that I could show a tokens hp in the combat tracker but not on the actual token. If I enable "player has access to sheet/stats" it shows on the token as well. If I then hide the hp/ac, it removes it from the combat tracker, but not from the token

image

Azmoria:
> nice! I did come across this which may be completely unrelated. that elf token has all configuration off. So should show the hp/ac right? it's not a player token it's a custom token. Tried with a monster token too. Seems weird that I could show a tokens hp in the combat tracker but not on the actual token. If I enable "player has access to sheet/stats" it shows on the token as well. If I then hide the hp/ac, it removes it from the combat tracker, but not from the token
> 
> image

Ah yeh. Basically the same issue different option.

Azmoria:
In #470 now and I think I addressed what you were talking about too bain

PR #457, opened 2022-05-01
Add: option to move a token to the top or bottom of the stack.

### LINKED ISSUE #456: Request: Allow users to adjust which tokens are on top/bottom of the stack
There's been a quite few requests to be able to move tokens to the top or bottom.

Azmoria:
Closes #456 

This adds options in the context menu to move the tokens to top or to bottom. Currently only for single tokens. If we go this route we can maybe add group options after.

Example Video: https://www.youtube.com/watch?v=L2Azf8w_vng


bain2236:
it works, however if I lock the token in place I can't seem to send another token behind it

Azmoria:
Going to convert for the new context menu

Issue #381, opened 2022-04-02
Grid visilibity


bain2236:
Few people have mentioned grids can be hard to view based on the map

https://discord.com/channels/815028457851191326/815028457851191329/959864939781185627
https://discord.com/channels/815028457851191326/815028716040224818/959891960469016597

bain2236:
I'll try get around to poking at this on the weekend

PR #458, opened 2022-05-02
Change grid colour and width. Scene edit form update with toggle buttons. Validate map urls on focussing out of those inputs

bain2236:
Adds colour picker and line width to edit scene dialog.
Allows live update of grid changes (color/width)

- General cleanup of edit dialog, moved away from checkboxes in favour of toggles like a bunch of other places #consistency
- Moved away from using 1/0 as input options (under the hood it's still 1/0 as to not break existing scenes)
- Added colour picker and line width inputs
- Cleanup of grid rendering
- Moved snap/colour up to the main block so now the buttons are more for scaling rather than all grid options
- Removed chunks from grid wizard that were overriding snap/grid choices

https://www.youtube.com/watch?v=IfRcMtsICFY

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -385,44 +385,51 @@ function midPointBtw(p1, p2) {
   };
 }
 
-function draw_grid(){
-	const canvas_grid = document.getElementById("grid_overlay");
-	const ctx_grid = canvas_grid.getContext("2d");
-	const incrementX = window.CURRENT_SCENE_DATA.hpps;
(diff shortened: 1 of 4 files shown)


vlaminck:
Closes #381

bain2236:
Updated to have some validation on the map inputs quick video can be found here 
https://www.youtube.com/watch?v=LIdt5joz8v4
this puts some work in towards #468 but does not complete it in it's entirety. Hopefuly it should lay some ground work that can be reworked for all url inputs

PR #460, opened 2022-05-03
Fix: some old contextMenu settings are being overwritten when contextMenu closes

vlaminck:
Some of the items that used to be in the contextMenu are still being set when the contextmenu is closed. This stops setting those values

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -1588,14 +1588,8 @@ function token_inputs(opt) {
 		return;
 
 	data = $.contextMenu.getInputValues(opt, $(this).data());
-	is_monster = window.TOKEN_OBJECTS[id].options.monster > 0;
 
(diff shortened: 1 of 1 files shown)


Azmoria (approved):
Fixes the issue for me. Good job 👍 

Way to replicate the problem: Change the name in the configure menu > Open the context menu and close it on the same token > The name will change back to it's original (I used a custom token as reported).

bain2236:
Can confirm this works

vlaminck:
This will no longer be needed once the context menu rebuild is complete. It doesn't hurt to merge this, but it won't do anything soon

Azmoria:
Like vlaminck mentioned this is now fixed in #470

PR #408, opened 2022-04-18
Add ability to roll DDB dice with modifiers

vlaminck:
# What
This adds the ability to roll DDB dice with modifiers.  
`DiceRoller.js` was added, but it isn't implement in this PR. It will be implemented by some of the other open PRs.

# How It Works
`DiceRoller` works by 
1. Parsing an expression (such as 1d20+4)
2. Clicks the DDB dice buttons and then the DDB roll button
3. Captures the corresponding "dice/roll/pending" message
a. If the message metadata needs to be altered it will do so
b. The captured message is then sent as expected
5. Captures the corresponding "dice/roll/fulfilled" message
a. Parses the dice roll results
b. Calculates the original expression using the dice roll results
c. The captured message is then sent as expected

# How To Use It
Create a `DiceRoll` object, and pass it to `window.diceRoller.roll`.

## Example 1
The most basic use is to simply call it with an expression. This will roll dice on behalf of the Player or DM, and will use all the default metadata that DDB uses. This code will result in the screenshot below

    var diceRoll = new DiceRoll("2d20kh1+2");
    window.diceRoller.roll(diceRoll);

Screen Shot 2022-04-18 at 11 03 41 AM

## Example 2
To roll on behalf of a monster, you would call it like this

    var diceRoll = new DiceRoll("1d20-3");
    diceRoll.action = "Big Fist";
    diceRoll.rollType = "damage";
    diceRoll.entityId = monsterData.id;
    diceRoll.name = monsterData.name;
    diceRoll.avatarUrl = monsterData.avatarUrl;
    window.diceRoller.roll(diceRoll);

Screen Shot 2022-04-18 at 11 04 00 AM

# Demo
Here's a video demonstrating the expression "2d20kh1+2" which is rolling with advantage and a +2 modifier. The DDB dice rolled a 10 and a 4, resulting in the final expression of 10+2=12
avtt-diceRoller

PR #406, opened 2022-04-17
rolling to "self" gives option to send to players

bain2236:
resolves #272 
re did the monitor message function as an observer
watches combat log and adds in button on expanded rolls
hides button before it collapses
shows button again if opened

PR #462, opened 2022-05-04
Dice rolling improvements

### LINKED ISSUE #272: Re-introduce show to players button for private rolls
The `monitor_messages` function at the bottom of `MessageBroker.js` does this, but the class names need to be updated to match the class names on the encounters page.

### LINKED ISSUE #302: Update rpgDiceRoller formatting to look more like DDB dice rolls

vlaminck:
This is a feature branch which combines the following PRs: 
* https://github.com/cyruzzo/AboveVTT/pull/399
* https://github.com/cyruzzo/AboveVTT/pull/406
* https://github.com/cyruzzo/AboveVTT/pull/408

# Make our dice rolling look like dndbeyond, includes rolling via chat box and player monster frames (extra tab/monster sheets). Added actions, types and more commands to chatbox rolling. Added "X/each" spell slot trackers to monster sheets

- Fixed up RPG dice roller to make it look like dndbeyond dice rolls
- Added custom actions/types via chat box
- Added helper text and /help command to chat box that opens the docs on dice rolling
- Uplifted dice buttons on monster sheets that aren't from encounters (now that players can have access to a monsters sheet)
resolves #302 

Some more user readable notes:

- Rolling via chat now has more commands and those rolls look like DNDBeyond rolling
- Rolling from extra tab now shows as DNDBeyond would if they even allowed rolling from there..
- Buttons on the monster sheet owned by players more closely align with the standard DDB styling and they also have all the information that you would expect in the combat log when rolled
- Spell slots trackers for for "X/each"
- Video.. apologies for the super quiet audio, youtube has clamped it down for whatever reason will get better at this :D https://www.youtube.com/watch?v=w4HftubO1ek

# rolling to "self" gives option to send to players

resolves #272 
re did the monitor message function as an observer
watches combat log and adds in button on expanded rolls
hides button before it collapses
shows button again if opened

# Add ability to roll DDB dice with modifiers
## What
This adds the ability to roll DDB dice with modifiers.  
`DiceRoller.js` was added, but it isn't implement in this PR. It will be implemented by some of the other open PRs.

## How It Works
`DiceRoller` works by 
1. Parsing an expression (such as 1d20+4)
2. Clicks the DDB dice buttons and then the DDB roll button
3. Captures the corresponding "dice/roll/pending" message
a. If the message metadata needs to be altered it will do so
b. The captured message is then sent as expected
5. Captures the corresponding "dice/roll/fulfilled" message
a. Parses the dice roll results
b. Calculates the original expression using the dice roll results
c. The captured message is then sent as expected

## How To Use It
Create a `DiceRoll` object, and pass it to `window.diceRoller.roll`.

### Example 1
The most basic use is to simply call it with an expression. This will roll dice on behalf of the Player or DM, and will use all the default metadata that DDB uses. This code will result in the screenshot below

    var diceRoll = new DiceRoll("2d20kh1+2");
    window.diceRoller.roll(diceRoll);

Screen Shot 2022-04-18 at 11 03 41 AM

### Example 2
To roll on behalf of a monster, you would call it like this

    var diceRoll = new DiceRoll("1d20-3");
    diceRoll.action = "Big Fist";
    diceRoll.rollType = "damage";
    diceRoll.entityId = monsterData.id;
    diceRoll.name = monsterData.name;
    diceRoll.avatarUrl = monsterData.avatarUrl;
    window.diceRoller.roll(diceRoll);

Screen Shot 2022-04-18 at 11 04 00 AM

## Demo
Here's a video demonstrating the expression "2d20kh1+2" which is rolling with advantage and a +2 modifier. The DDB dice rolled a 10 and a 4, resulting in the final expression of 10+2=12
avtt-diceRoller

### DIFF
--- a/MonsterDice.js
+++ b/MonsterDice.js
@@ -1,104 +1,138 @@
 
-/// this injects roll buttons into the monster details page
-function scan_monster(target, stats, token_id=false) {
+/**
+ * Adds in rolling buttons and spell tracker inputs to monster frames that are not encounter frames.
+ * Specifically monster frames that appear to the player
+ * @param {$} target jqeury selected element to put rolling buttons into
+ * @param {object} stats stats of the monster
+ * @param {string} tokenId id of the token
+ */
+function scan_monster(target, stats, tokenId) {
+	console.group("scan_monster")
+	console.log("adding in avtt dice buttons")
+	// remove homebrew panels
+	target.find(".homebrew-creation-actions").remove();
+	target.find(".homebrew-previous-versions").remove();
+	target.find(".homebrew-details-footer").remove();
+	target.find("footer").remove()
+	target.find("#footer-push").remove();
+	target.find("#footer").remove();
+	$("iframe")
 	target.find(".integrated-dice__container").hide();
-	target.find(".mon-stat-block__description-block-content p").each(function(idx, element) {
+	// attempt to the tokens name first, failing tha
+	const creatureName = window.TOKEN_OBJECTS[tokenId]?.options.name || target.find(".mon-stat-block__name-link").text(); // Wolf, Owl, etc
+	const creatureAvatar = stats.data.avatarUrl
+	const displayName = `${pc.name} (${creatureName})`;
 
-		$(element).find("span[data-rolltype='to hit']").each(function() {
-			newblock = $("<div/>");
-			newblock.css("display", "inline-block");
+	const clickHandler = function(clickEvent) {
+		roll_button_clicked(clickEvent, displayName, creatureAvatar)
+	};
 
+	const rightClickHandler = function(contextmenuEvent) {
+		roll_button_contextmenu_handler(contextmenuEvent, displayName, creatureAvatar);
+	}
 
+	replace_ability_scores_with_avtt_rollers(target, ".ability-block__stat" ,".ability-block__heading", true)
+	replace_saves_skill_with_avtt_rollers(target, ".mon-stat-block__tidbit", ".mon-stat-block__tidbit-label", ".mon-stat-block__tidbit-data", true )
 
-			modifier = $(this).text();
-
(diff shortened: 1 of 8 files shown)


MonsterDice.js
```diff
@@ -1,104 +1,138 @@
...
-		
-		description = description.replaceAll(/\(([0-9]+d[0-9]+( [\+-] [0-9]+)?)\)/g, "(<button data-exp='$1' data-mod='' class='above-roll20'>$1</button><button data-exp='$1' class='above-roll20' data-mod='CRIT'>CRIT</button>)");
-		$(element).html(description);
-		
-		
+	// replace all "to hit" and "damage" rolls
+	$(target).find(".mon-stat-block p").each(function() {
+		let currentElement = $(this)
+		if (currentElement.find(".avtt-roll-button").length == 0) {
+			$(currentElement).find("span").each(function (){
```
bain2236, line 41:
```suggestion
			$(currentElement).find("span").each(function (){
$(currentElement).find("span[data-dicenotation]").each(function (){
```

This solves the _little death_ hyperlink issue

vlaminck:
> 1. Image shown in gamelog isn't the image on the token. Although this is probably best as you could have the same creature with different token images.

Nice catch, fixed

> 2. Roll to self button doesn't show on players side. I hope this is intentional to stop players from rolling multiple rolls to self then showing everyone the good roll.

I'm unable to reproduce this so maybe I accidentally fixed it?

> 3. Minor issue with monster sheet button injection. Came across this. Little death is a link as it's a Grim hollow spell that I've homebrewed in and as such won't have the nice on-hover

I don't think the fancy hover stuff works at all any more... need to rebuild that I guess

Issue #463, opened 2022-05-04
Bug: In current main markers are not working on monsters


Azmoria:
So it looks like in trying to get inspiration from player sheets when we took out` inspiration` from `const CUSTOM_CONDITIONS` in tokens.js at the top it stopped markers from working on monsters.

PR #464, opened 2022-05-04
Fixes not being able to put markers on monsters

Azmoria:
Fixes: #463

This bug only exists in main not the live version.

It also stops markers from going on player tokens if they don't have inspiration. This fixes that as well.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -645,7 +645,9 @@ class Token {
 			}
 		}
 		else{
-			this.options.custom_conditions.pop("Inspiration")
+			if (this.options.custom_conditions.includes("Inspiration")){
+				this.options.custom_conditions.pop("Inspiration")
+			}
 		}
 		
 		


vlaminck (approved):

bain2236:
I dropped the ball on this one. But this totally fixes it

vlaminck (changes requested):
`pop` doesn't work like that. I always pops no matter what the value is. I fixed this in your token context menu rebuild branch so this PR probably isn't needed any more, but this is how I fixed it


    if (this.options.inspiration){
        if (!this.options.custom_conditions.includes("Inspiration")){
            this.options.custom_conditions.push("Inspiration")
        }
    }
    else {
        array_remove_index_by_value(this.options.custom_conditions, "Inspiration");
    }

Azmoria:
> `pop` doesn't work like that. I always pops no matter what the value is. I fixed this in your token context menu rebuild branch so this PR probably isn't needed any more, but this is how I fixed it
> 
> ```
> if (this.options.inspiration){
>     if (!this.options.custom_conditions.includes("Inspiration")){
>         this.options.custom_conditions.push("Inspiration")
>     }
> }
> else {
>     array_remove_index_by_value(this.options.custom_conditions, "Inspiration");
> }
> ```

Yeh that makes sense I had just wrapped what was there to check if inspiration was there. This seemed to fix the issue so I left it. Closing this since your fix is in the other one

PR #470, opened 2022-05-07
Scaling Tokens, New Context Menu, Border Colors, Ping Colors

### LINKED ISSUE #255: Cleanup the token context menu
Most of the context menu options could be moved into a SidebarPanel modal. By adding a `Configure Token` option that opens a sidebar modal, we can greatly simplify the token context menu to only the things that get used a lot.
These are the best candidates to move:
- size
- options
- name
- change image 
- lock/unlock/restrict player movement

Azmoria:
This is another branch off #409 with a new context menu. I can split it out if necessary.  

Mikedave with the big assist on the new context menu. Closes #255 

Adds:
• Token Scaling
• New Context Menu
• Custom Border Colors
• Ping colors match tokens - DM can set a token name to THE DM to choose their color if they want.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -118,12 +118,20 @@ class Token {
 		this.place_sync_persist()
 	}
 
+	imageSize(imageScale) {
+		this.update_from_page();
+		this.options.imageSize = imageScale;
+		this.place_sync_persist()
+	}
+
 	hide() {
 		this.update_from_page();
 		this.options.hidden = true;
 		this.options.ct_show = false;
-		$("#"+this.options.id+"hideCombatTrackerInput ~ button svg.closedEye").css('display', 'block');
-		$("#"+this.options.id+"hideCombatTrackerInput ~ button svg.openEye").css('display', 'none');
+		if(this.options.monster) {
+			$("#"+this.options.id+"hideCombatTrackerInput ~ button svg.closedEye").css('display', 'block');
+			$("#"+this.options.id+"hideCombatTrackerInput ~ button svg.openEye").css('display', 'none');
+		}
 		this.place_sync_persist()
 		this.update_and_sync()
 		ct_persist();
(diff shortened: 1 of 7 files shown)


Azmoria:
Converted conflicting PRs and merged them in. To Top/Bottom context menu option. And option to only show hp bar on hover.

Token.js
```diff
@@ -643,9 +713,9 @@ class Token {
 			if (!this.options.custom_conditions.includes("Inspiration")){
 				this.options.custom_conditions.push("Inspiration")
 			}
-		}
-		else{
-			this.options.custom_conditions.pop("Inspiration")
+		} else{
+			array_remove_index_by_value(this.options.custom_conditions, "Inspiration");
+			array_remove_index_by_value(this.options.custom_conditions, "Inspiration");
```
bain2236, line 718: assuming this is fine as the second call won't find it and will just return? not looked into what `array_remove_index_by_value` does, but assuming it's a find by index and a splice?

TokenMenu.js
```diff
@@ -2038,6 +2253,558 @@ function build_token_auras_inputs(tokens) {
...
+				token.options.color = borderColor;
+				$("#combat_area tr[data-target='" + token.options.id + "'] img[class*='Avatar']").css("border-color", borderColor);
+				token.place_sync_persist();
+			});
+		}
+		else {
+			tokens.forEach(token => {
+				let selector = "div[data-id='" + token.options.id + "']";
+				let html = $("#tokens").find(selector);
+				let options = Object.assign({}, token.options);
```
bain2236, line 2711: html and options supposed to be assigned by not used?

bain2236:
Looks so nice! plenty of additional features that all work seamlessly

TokenMenu.js
Azmoria: This is probably part of clean up to do here and there that was missed if they aren't being used. There's definitely clean up with removing the old context menu code.

Azmoria: I'll clean this up later today - probably copy paste mistake

Azmoria: This should be fixed now as should the fading option toggles

PR #474, opened 2022-05-11
Fix: Stop group selection of player restricted tokens for players

### LINKED ISSUE #473: Bug: Players can move player restricted tokens with group select.
Reported by Bain: players can move player restricted tokens with group select.

Azmoria:
This stops group selection of player restricted tokens. This fixes #473 and is a qol improvement for players with multiple tokens to select. They won't accidently grab monsters.


bain2236:
Simple but does the job and is a very nice quality improvement for a player, especially if they're on a mount/have familiars and drag select

PR #478, opened 2022-05-11
fix: change image doesn't work for monsters

vlaminck:
this also fixes the monster configure button within encounter folders


Azmoria (approved):
This fixes both issues mentioned for me. No issues that I can find.

PR #480, opened 2022-05-11
fix for gamelog images not showing when clicked

vlaminck:
`href` is not valid on `img` tags, but magnific uses it. I'm sure it was removed because it's not valid on img tags so I added a comment to hopefully prevent this bug from coming back in the future


Azmoria:
This fixes this issue for me

Azmoria (approved):

PR #484, opened 2022-05-11
Fix: Set width or height of tokens that retain their aspect ratio to fit larger size tokens when the image isn't big enough

### LINKED ISSUE #483: Bug: Tokens larger than their source image are not changing size to match large token sizes
Reported by MaxGaming3005

image

Azmoria:
This checks which side is longer and sets that side's min-width to fit the token size if `ignore aspect ratio` is disabled. This ensures that a 1.0 scale image fits it's token size properly when the token is larger than the img src. It does this by stretching token images on the longer sides axis to maintain aspect ratio. 

Closes #483

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -414,7 +414,15 @@ class Token {
 			'max-height': tokenHeight + 'px',
 		});
 
-
+		if(!this.options.legacyaspectratio) {
+			if(token.children('img').width() >= token.children('img').height()) {
+				token.children('img').css("min-width", tokenWidth + 'px');
+			}
+			else {
+				token.children('img').css("min-height", tokenHeight + 'px');
+			}
+		}
+		
 		console.groupEnd()
 	}
 
--- a/Token.js
+++ b/Token.js
@@ -408,21 +408,29 @@ class Token {
 			token.css('--token-hpbar-display', 'block');
 		}
 		token.attr("data-border-color", this.options.color);
-	
-		token.children('img').css({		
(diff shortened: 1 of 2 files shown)


vlaminck (changes requested):
This works when making tokens larger, but gives you wide, squat tokens when you make tokens smaller.

[demo video](https://imgur.com/a/vH1uTrZ)


Screen Shot 2022-05-12 at 6 55 05 AM

Azmoria:
> This works when making tokens larger, but gives you wide, squat tokens when you make tokens smaller. This is only if you have the `ignore aspect ratio` enabled.
> 
> [demo video](https://imgur.com/a/vH1uTrZ)
> 
> Screen Shot 2022-05-12 at 6 55 05 AM

I can't reproduce this. Maybe some race condition jankyness.

Azmoria:
Alright I took a guess at the cause since I can't reproduce this. Let me know if it's still broken. If so I may need some assistance tracking down the cause. It looks like min-width isn't updating in your video which is odd. 

Any other settings on that token or was it just a fresh token?

vlaminck:
> Alright I took a guess at the cause since I can't reproduce this. Let me know if it's still broken. If so I may need some additional assistance tracking down the cause. Reading your comment again I'm pretty sure it's cause the min-width wasn't being set back to default.

Yep! that fixed it 👍

vlaminck (approved):

PR #485, opened 2022-05-12
Fix conditions showing up when player token is selected

### LINKED ISSUE #476: DMs should not be able to add conditions to player tokens
(this was accidentally added in one of the last context menu changes).
As usual, we should have a warning that Players Conditions should be done by players on their character sheet

Azmoria:
Fixes #476 

Conditions will only show up in the context menu when no player tokens are selected.
A warning is added at the bottom of markers if a player token is selected.

image
image

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -1692,10 +1692,10 @@ function init_things() {
 
 	} else {
 		init_ui();
-		init_splash();on
+		init_splash();
 	}
 
-	$("#site").append("<div id='windowContainment'></div>");	            
+	$("#site").append("<div id='windowContainment'></div>");
 }
 
 /// this is used when initializing on the character page. DDB loads the page in an async modular fashion. We use this to determine if we need to call other initialization functions during this process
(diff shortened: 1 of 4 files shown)


vlaminck (changes requested):
This works, but I wonder if we should just add the note to the bottom of the Conditions list and then just not set them for players. From a usability perspective that would allow you to just select a bunch of tokens without having to deselect the players in the selection.

Either way, we should change `Token.addCondition` to the following since adding conditions to players isn't valid any time.
```
	addCondition(conditionName) {
		if (this.hasCondition(conditionName)) {
			// already did
			return;
		}
		if (STANDARD_CONDITIONS.includes(conditionName)) {
			if (!this.isPlayer()) {
				this.options.conditions.push(conditionName);
			}
		} else {
			this.options.custom_conditions.push(conditionName);
		}
	}
```


If you continue to hide the conditions when a player is selected, we should also move the `Remove All` button to the Markers list so that it shows up in this scenario.

Azmoria:
Example video of how it currently works. I discussed with mikedave in discord about whether we should automate this completely. We decided for now we wouldn't as it might confuse players and this gives them a little more control and allows other players to make requests for conditions for things like invisibility.

https://youtu.be/slxKMQm04EE

vlaminck (approved):
fantastic!

PR #486, opened 2022-05-12
Fix: Tooltips for monster iframes.

### LINKED ISSUE #482: hovering on a spell or condition in a monster stat block doesn't display the description anymore, nor can it be sent to the gamelog with a right-click
This is broken in the 0.76 RC builds. I tried to fix it, but I think it's going to take some special work that I haven't been able to figure out yet.

The problem that I've run into is that the monster details page doesn't always update `#db-tooltip-container` in such a small iframe. The combat-tracker page (which is where this used to be) was more consistent with it which is what made it possible. I'm sure we can still do this, but we'll need to dig into it more.

This is what I've tried so far. In `Main.js` `load_monster_stat`

```
iframe.on("load", function(event) {
	if (!this.src) {
		// it was just created. no need to do anything until it actually loads something
		return;
	}



	// ... everything that's currently in this block of code goes here



	// Add some hover text to let users know they can right click tooltips to send them to the gamelog
	$(event.target).contents().find(".tooltip-hover").hover(function() {
		$(this).attr("title", "right-click to send to gamelog")
	});

	// if the user right-clicks a tooltip, send it to the gamelog
	$(event.target).contents().off("contextmenu").on("contextmenu", ".tooltip-hover", function(clickEvent) {
		clickEvent.preventDefault();
		clickEvent.stopPropagation();
		if ($("#db-tooltip-container").first().length > 0) {
			let toPost = $("#db-tooltip-container").children().first().clone();
			toPost.find(".waterdeep-tooltip").attr("style", "display:block!important");
			toPost.find(".tooltip").attr("style", "display:block!important");
			toPost.css({
				"top": "0px",
				"left": "0px",
				"position": "relative"
			});
			toPost.find(".tooltip").css({
				"max-height": "1000px",
				"min-width": "0",
				"max-width": "100%",
				"width": "100%"
			});
			toPost.find(".tooltip-header").css({
				"min-height": "70px",
				"height": "auto"
			});
			toPost.find(".tooltip-body").css({
				"max-height": "1000px"
			});
			window.MB.inject_chat({
				player: window.PLAYER_NAME,
				img: window.PLAYER_IMG,
				text: toPost.html()
			});
		}
	});

	$(event.target).contents().find("body").on("DOMSubtreeModified", "#db-tooltip-container", function(modifiedEvent) {
		
		// make the popup container visible. This is where tooltips will be injected
		let toolTipContainer = addedElement.closest("#db-tooltip-container");
		toolTipContainer.css({"visibility": "visible", "display": "block"});



	});


```

Azmoria:
First off Dndbeyond won't create tooltips on the monster page if the width of the window is <= 640. 
```
 createTooltip: function (event) {
            if (this.Disabled || !event.currentTarget.getAttribute('data-tooltip-href') || $(document).width() <= 640) {
                return false;
            }
```
The frustrating part is after they are created you can shrink the window down and they work fine. I just couldn't find a way to get it to create the stuff automatically. So if we had a way to have them created and then allow free resizing that'd be great but I couldn't figure that one out. 

I'm unsure if this is the route we want to take with the downsides but I can't think of how we'd go about different with getting the tooltips to show up in general not just in the iframe

Edit: Now fully fixes #482 

Both players and DM can send from monster stat blocks they control

image

### DIFF
--- a/abovevtt.css
+++ b/abovevtt.css
@@ -273,7 +273,7 @@ div.ct-sidebar__pane-content--is-dark-mode .sidebar-panel-content {
     left: 271px;
     top: 67px;
     right: auto;
-    width: 410px;
+    min-width:  670px;
     height: 85%;
     position: fixed !important;
     opacity: 1;
--- a/abovevtt.css
+++ b/abovevtt.css
@@ -273,7 +273,7 @@ div.ct-sidebar__pane-content--is-dark-mode .sidebar-panel-content {
     left: 271px;
     top: 67px;
     right: auto;
-    min-width:  670px;
+    width: 410px;
     height: 85%;
     position: fixed !important;
     opacity: 1;
--- a/Main.js
+++ b/Main.js
@@ -515,7 +515,7 @@ function load_monster_stat(monsterid, token_id) {
 		$("#site").prepend(container);
 	}
 
-	$(iframe).on("load", function(){
+	$(iframe).on("load", function(event){
 		let tooltipCSS = $(`<style>.hovering-tooltip{ display: block !important; left: 5px !important; right: 5px !important; pointer-events: none !important; min-width: calc(100% - 10px);} </style>`);
 		$("head", $("#resizeDragMon iframe").contents()).append(tooltipCSS);
 		$(".tooltip-hover", $("#resizeDragMon iframe").contents()).on("mouseover mousemove", function(){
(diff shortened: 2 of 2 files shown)


abovevtt.css
```diff
@@ -273,7 +273,7 @@ div.ct-sidebar__pane-content--is-dark-mode .sidebar-panel-content {
     left: 271px;
     top: 67px;
     right: auto;
-    width: 410px;
+    min-width:  670px;
```
vlaminck, line 276: I don't like how big this is, but if it's the only way, it's probably fine. Do you think it would be possible to just open the sheets to this size instead?... hmmm. I guess we might need to fake a hover event to get DDB to inject `#db-tooltip-container` 🤔

vlaminck: That was the only thing I could think of... I think this is the best approach for now, and if we can figure out how to do it later on, then we can adjust the `min-width`. For now, can you just add a comment next to the `min-width` to explain why we're setting that value? Otherwise, down the road someone will come in and be like "oh sweet, I just need to lower this!" and then accidentally break the hover states.

Azmoria: I got it fixed for smaller windows using blank space off screen in the iframe to trick it into thinking it's bigger.

PR #497, opened 2022-05-16
Fix: When a hidden token is added to combat it should now properly always default to being hidden in the combat tracker.

### LINKED ISSUE #496: Bug: Tokens not hiding in combat tracker by default when they are expected to
There are 2 scenarios that hidden tokens are not being added to the combat tracker hidden from players.

1) When the token is placed on the map hidden

2) When the token has been in the combat tracker already, made visible and then removed from the combat tracker. (This one might not be as important as it may not come up in play as much) 

I suspect this is due to some `token.options.` that's being checked not being set.

Azmoria:
Fixes #496 

This will always add hidden tokens as hidden from players in the combat tracker.

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -343,7 +343,6 @@ function ct_add_token(token,persist=true,disablerolling=false){
 			function(){
 				token.options.ct_show = undefined;
 				ct_remove_token(token);
-				ct_persist();
 			}
 		);
 		if(window.DM)
@@ -429,7 +428,8 @@ function ct_persist(){
 	  data.push( {
 		'data-target': $(this).attr("data-target"),
 		'init': $(this).find(".init").val(),
-		'current': ($(this).attr("data-current")=="1")
+		'current': ($(this).attr("data-current")=="1"),
+		'data-ct-show': window.TOKEN_OBJECTS[$(this).attr("data-target")].options.ct_show
 	   });
 	});
 	data.push({'data-target': 'round',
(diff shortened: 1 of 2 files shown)


Ellasar742 (changes requested):
I think this will block the ability to reveal hidden tokens, which was a requested feature. 

It seems like the issue is when adding a token that was originally added as hidden to the CT right? I think it's only an initial condition issues. So something like 
`if (typeof(token.options.ct_show) == 'undefined' && token.options.hidden == true){
token.options.ct_show = false;
}
else if (typeof(token.options.ct_show) == 'undefined'){
token.options.ct_show = true;
}`
(Didn't have time to test this but I think this is right) 
should handle the initial case and keep all functionality.

Azmoria:
I did not expect to be running add to combat function everytime we toggle that. My mistake I will fix that. I just assumed the toggle was working because the toggle still visually changed.

Azmoria:
That should fix the two problem cases while retaining the toggle functionality.

Ellasar742:
Yep, when I asked around previously, people wanted the CT to update actively whenever hiding a token, the easiest way to do that was to just scan for hidden and repopulate. I admit, it's a little clunky, but I it shouldn't be noticeable unless people have huge combats and are toggling a bunch of hidden tokens)

Ellasar742 (approved):
Tested and works. Should cover all the edge cases found.

Azmoria:
It wasn't updating the ct_show option on the player side. So the issue where if a token has left combat and reentered was glitcty still. The last fix should now fix all of it.

PR #498, opened 2022-05-16
Fix: grid toggles will no longer always show as on

### LINKED ISSUE #499: Bug: Grid toggles (snap and show) always default to on when opening the menu
The grid toggles in the scene panel always default to on when opening the menu.

Azmoria:
Fixes #499

Grid toggles for scenes no longer will always show as on. This also adds the grid toggle setting to apply automatically to the DM view when toggled.

This does not fix the player view needing to be refreshed to take effect. Although the DM can just click players again to force a refresh it is not great for potatos. That can be a separate fix as that issue already existed.

### DIFF
--- a/ScenesPanel.js
+++ b/ScenesPanel.js
@@ -278,7 +278,18 @@ function edit_scene_dialog(scene_id) {
 						form_toggle("dm_map_usable",null, false, handle_basic_form_toggle_click)
 						)
 				);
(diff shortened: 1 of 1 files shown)


cyruzzo:
the player view needing to be refreshed is fixed by another PR + a change in the backend
I'll merge this now and probably release an hotfix soon

PR #500, opened 2022-05-17
Move Scenes list to sidebar

### LINKED ISSUE #503: As “legacy” tag to monsters in the tokens panel
Now that Monsters of the Multiverse is out, we should add the “legacy” tag like DDB does to help differentiate monsters

### LINKED ISSUE #508: Move the List of Scenes into the Sidebar
Instead of a window at the top, scenes should be listed in their own tab in the sidebar. 

DMs should be able to 
* create/edit/delete scenes (create/edit should open the existing modal)
* create/update/delete folders (create/edit should open a sidebar modal like token folders does)
* move scenes into/out of/between folders
* Move DM to a scene independently of players
* Move players to a scene independently of the DM
* Search for scenes by name

### LINKED ISSUE #509: Allow moving tokens between folders

vlaminck:
Closes #508 
Closes #503
Closes #509

# What

1. Move the list of Scenes into a new Scenes tab in sidebar utilizing the `SidebarListItem` framework that was added during the the Tokens Panel unification. 
2. Allow Scenes to be added to folders. (cannot be moved yet)
3. Allow Searching of Scenes
4. Refactored some of the `SidebarListItem` framework to be more generic and reusable.
5. This is only available if `window.CLOUD` is true. Non-migrated users will continue to use the old window at the top of the screen.

# Other Considerations

We discussed not showing images in the sidebar to reduce the initial download and memory impact. I chose to leave it in for now so we could all play with it before making a final decision on it.

# Demos
### Folder and scene creation
avtt-scenes-panel-demo-1

### Display larger image on hover
avtt-flyout-styling

### Scene and folder deletion
avtt-scenes-panel-deletion

### Scene search
avtt-scenes-panel-search

### Moving Scenes between folders
https://user-images.githubusercontent.com/584771/169146013-16312f54-9bee-48ac-bcc4-e15684a925b2.mov

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -104,6 +104,10 @@ Mousetrap.bind('esc', function () {     //deselect all buttons
     $('#select-button').click();
     $("#tokenOptionsClickCloseDiv").click();
     $(".draggable-token-creation").addClass("drag-cancelled");
+    $(".draggable-sidebar-item-reorder").addClass("drag-cancelled");
+    try {
+        $( '.ui-draggable-dragging' ).draggable("option", { revert: true }).trigger( 'mouseup' ).draggable("option", {revert: false })
+    } catch (whoCares) { }
 });
 
 
--- a/Main.js
+++ b/Main.js
@@ -691,6 +691,20 @@ function init_controls() {
 		b2ImageDivWrapper.append(b2ImageDiv);
 		b2.append(b2ImageDivWrapper);
 		sidebarControls.append(b2);
+
+		if (window.CLOUD) {
+			let b3 = $("<div id='switch_scenes' class='tab-btn hasTooltip button-icon blue-tab' data-name='Scenes' data-target='#scenes-panel'></div>").click(switch_control);
+			let b3ImageDiv = $('<div></div>');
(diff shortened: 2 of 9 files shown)


bain2236:
couple of things but even still this is soooooooooooooo nice and polished compared to the last dropdown. 
- I would say the image popout is a little fast that I wasn't expecting it :D despite literally seeing a video of it yesterday I completely forgot.
- "New scene" buttons on hover is a copy paste mistake
- when changing scene the red active colour for the players flickers. I throttled my network speed and it does look like when it's loading a new scene it doesn't know where the players are.

vlaminck:
> * I would say the image popout is a little fast that I wasn't expecting it :D despite literally seeing a video of it yesterday I completely forgot.

I left this in for now. I thought it would be good to get others playing with it, and then we can remove it before releasing it if we want to.

> * "New scene" buttons on hover is a copy paste mistake

Fixed. good catch!

> * when changing scene the red active colour for the players flickers. I throttled my network speed and it does look like when it's loading a new scene it doesn't know where the players are.

I was blindly removing the class when the DM button was pressed. I changed it to explicitly remove the class from DM buttons only, and now it doesn't flicker.

PR #506, opened 2022-05-21
Fix dice streaming

Azmoria:
There's still a lot wrong with this but thought others could look at it and some wanted to see it. 

This works with firefox now. 

Some of the issues if anyone wants to tackle them - this will probably be pretty slow going for me:

- [x] Add the join button to player view if the DM enabled it before they launched AboveVTT

- [x] Destroying/removing the elements when it's turned off for players. <s>Partially complete. Need to close RTC connections</s>

- [x] Have players join after it's been enabled (? Do we want to leave a way for people not to stream. It'd probably be better to build a disable button on the player side but currently if they join after it's been enabled they won't stream their dice unless it's reset)

- [x] It starts off low resolution but small getting bigger until it's the size of the original roll window. This can result in some dice going off screen if the original window is larger. Not sure there's much we can do about it starting low res as I think this is a WebRTC buffering thing. Dice from firefox clients start larger. I guess it's how the browsers process the video before sending it. Marking this as done unless someone knows a way to do something about this
- [x] Create join/leave stream button once enabled

- [x] <s>Local connections now work for me. Need to check if external connections are a server problem or code adjustments are needed. </s> I think this was a networking issue on my end. I connected to my modem directly and could then connect to my friend. Maybe a fluke not sure.

- [x] <s>The video connection offers sometimes clash. A better way to stagger the calls would be better or a way to determine which should answer the offer. I think this causes the need to toggle it off and back on to force a reconnect currently.  </s>I added politeness logic to determine which client should answer.

<s>Addressing this somewhat I added new servers. They are free to use, can be used in I think all software but also open source projects. The connections seem to be more consistent and I no longer get errors requesting a STUN server (I added on from the same) More info here:

https://www.metered.ca/tools/openrelay/
https://www.metered.ca/tools/openrelay/terms-and-conditions</s>

I switched this to a google server as the aboveVTT turn and the one I found weren't working today externally.

- [x] Contain the dice to screen despite coming from a larger window area

- [x] There's some duplicate elements. 

- [x] The DM rolls to self show. We can either try to find a way to only stream rolls to `everyone` or disable DM stream rolls all together. 

- [x] I'm having some issues getting all the clients to connect to each other. The DM seems to be able to connect to all of them without issue most of the time however the player view seems sporadic with other players. I could use some help with this. I don't know if it's a race condition or the codes in the wrong spot. I've melted my brain on this too much today. Edit: There may be other issues as I tested with someone else and was only able to get this to work locally. Still with some clients getting stuck connecting to each other. I tried a different server that is free just to test and that also didn't work. It's possible the answer is out pacing the offer or ice stuff as well. Or something to do with them both making requests around the same time. NOTE: This might need more tweaking but I got the players connecting to each other now 🥳

I think there's probably more to fix but these are the main ones

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -2698,7 +2698,7 @@ function init_buttons() {
 
 	window.STREAMPEERS={};
 	window.MYSTREAMID=uuid();
-	window.JOINTHEDICESTREAM=true;
+	window.JOINTHEDICESTREAM=false;
 
 
 
--- a/Main.js
+++ b/Main.js
@@ -2698,7 +2698,7 @@ function init_buttons() {
 
 	window.STREAMPEERS={};
 	window.MYSTREAMID=uuid();
-	window.JOINTHEDICESTREAM=false;
+	window.JOINTHEDICESTREAM=true;
 
 
 
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -679,24 +679,7 @@ class MessageBroker {
 				if( (!window.MYSTREAMID))
 					return;
 				const configuration = {
-    				iceServers:  [ {
-					     	urls: "stun:openrelay.metered.ca:80",
-					    },
-					    {
-					      urls: "turn:openrelay.metered.ca:80",
-					      username: "openrelayproject",
-					      credential: "openrelayproject",
-					    },
-					    {
-					      urls: "turn:openrelay.metered.ca:443",
-					      username: "openrelayproject",
-					      credential: "openrelayproject",
-					    },
-					    {
-					      urls: "turn:openrelay.metered.ca:443?transport=tcp",
-					      username: "openrelayproject",
-					      credential: "openrelayproject",
-					    }]
+    				iceServers:  [ {urls: "stun:stun.l.google.com:19302"}]
   				};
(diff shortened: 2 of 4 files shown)


cyruzzo:
Hm..
I need to test this more.
It seems to kill my laptop currently. I get "super slow animations" even on the source.
Also.. didn't test firefox.. I remember that firefox used to "continuosly send the data".  So you would get a continuos stream (not sure if this one fixes it)

Azmoria:
> Hm.. I need to test this more. It seems to kill my laptop currently. I get "super slow animations" even on the source. Also.. didn't test firefox.. I remember that firefox used to "continuosly send the data". So you would get a continuos stream (not sure if this one fixes it)

My friends and I only got slow animations but only during the initial buffer. I also got some slow animations when running it on the same machine with multiple 3+ browsers. None of us had issues but I did suspect it might have issues on lower end machines which is why we left it per player. 

As for the firefox part - I don't know how webRTC functions in firefox vs chrome tbh but I experienced less issues in firefox myself. You're probably a better test than my/my friends setup for performance if you experience continuous performance problems with only the 1 open.

PR #513, opened 2022-05-24
Draw text to map, Draw menu cleanup. Token measuring by default, token measuring waypoints via drop and pickup within ruler fadeout

### LINKED ISSUE #315: Allow users to add text to maps
As a user, I want to add text labels to my maps similar to how I can add drawings, but without having to hack tokens. 

Ideally, users would be able to set and edit things like 
* font family
* font size
* font color
* stroke width
* stroke color
* text alignment
* background color
* map position

Fine tuning the value and position of a text label will be important.

bain2236:
Resolves #315 
No issues created for allowing token measuring, it was something that just organically came about when talking in the dev channel

- Draw menu cleanup, no more multiple colours in the draw menu, just an rgb colour picker 
- Drawing functionality cleanup (would have liked to have done polygon cleanup too as I think there could be improvements there)
- Drawing menu usability cleanup. they're now buttons so you can tab through them. up down still works and numbers work too (despite not showing the user they can use numbers)
- Drawing occurs on a seperate temp layer and is baked into `window.DRAWINGS/window.REVEALS` respectively
- Text menu added
- Text draws a rect that turns into a text box with a settings panel
- Text is put into `window.drawings` for ease. operations that run over window.drawings perform a filter of drawings or text
- 🥇 🍺  🦝  Emoji's can be drawn using `Windows Key + .` no idea what mac users do..
- `CTRL + z` now works on all the drawing menus that have an undo button (I looked into the undo deleted token but I think it's a dead function as it didn't work out hte box?)

- Ruler has a fadeout
- Tokens auto measure. dropping and picking back up within the fadeout of the ruler will create a waypoint
- No change to how waypoints with when the measure button is on
- Token measuring uses border colour of the token
- Tokens now snap to grid when moving (unless `CTRL` is pressed or snapping is disabled)

There's too much stuff to spend ages recording videos. So try it out instead 😵‍💫 

Big shout out to @vlaminck  and @Azmoria for the help
A mention out to all the other awesome-devs and Above-vtt team for all the feedback throughout this whirlwind of a task


cyruzzo:
This seems to introduce a bug.
Try to grab a big map (I tested with Tales of the Yawrning Portal / Sunless Citaded / Fortress level).
You'll be able to only draw on the upper part of the map.
Also, the ruler only work on the upper part of the map

cyruzzo:
Another thing. In the past we had  performance issue with having the "ruler on" while dragging tokens. This somehow mde the browser "repaint" the window and caused general slowiness especially on big maps. If I remember correctly the issue was more annoying on firefox. So we need to check that too.

PR #517, opened 2022-05-26
support non-square tokens

vlaminck:
This came out of a conversation about line AoE tokens. I'm not sure we would want to support it for any other tokens, but it could work ok. I don't love that the borders don't follow the rotated shape, but we can make that better over time if we decide to support this.
avtt-non-square-token-support


vlaminck:
moved to a PR against @bain2236's AoE branch https://github.com/bain2236/AboveVTT/pull/6

PR #560, opened 2022-06-08
Build Monster stat blocks locally instead of loading iFrames

vlaminck:
This builds monster stat blocks instead of loading an iFrame every single time. This requires very little network data, way less memory usage (because no iframes), and it's way faster. We fetch the monster data from the DDB API once and then cache it. Every time after that, we render from cached objects.

This is what the stat blocks loading iFrames looks like https://imgur.com/a/L7zU4yL
And this is what building locally looks like https://imgur.com/a/gWlpdv5

If we ever want to use iFrames again for any reason, we can easily do so by calling `use_iframes_for_monsters()`. We can also switch back by calling `stop_using_iframes_for_monsters()`. I hope we never need to tell users to call either of these functions, but they exist just in case.

### DIFF
--- a/EncounterHandler.js
+++ b/EncounterHandler.js
@@ -297,31 +297,6 @@ class EncounterHandler {
 		});
 	}
 
-	fetch_monsters(monsterIds, callback) {
-		if (typeof callback !== 'function') {
-			console.warn("fetch_monsters was called without a callback.");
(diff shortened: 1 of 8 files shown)


vlaminck:
Closing this as it has been merged into #576

PR #561, opened 2022-06-09
Token Updates

vlaminck:
# Public Release Notes
Token Updates by @Azmoria#7532, @mikedave#6365, and new developer @H2owsome#0852 🚀 🥳 
- Added Token "styles" that override token settings based on the style you choose. Styles include:
    - "Circle" (Great for portrait tokens)
    - "Square" (Great for portrait tokens)
    - "Virtual Mini w/ Round Base" (Great for top-down tokens)
    - "Virtual Mini w/ Square Base" (Great for top-down tokens)
    - "No Constraint" (most customizable)
- Monster HP can now be set to average, max, or rolled.
- All token options can now be overridden on a per-token basis as well as a per-folder basis.
    - You can override token options on players, monsters, custom tokens, and any folder within the My Tokens folder!
    - Token option overrides are hierarchical, meaning:
        1. Changing token options in the settings tab applies to all tokens
        2. Changing token options on a folder applies to all tokens within that folder
        3. Changing token options on a token applies to only that token
    - When you place a token on a scene, the options get applied in that order.
- D&D Beyond Portraits can now be used as tokens! Find them in a new folder within the Tokens Panel.

# Technical Release Notes
- Token options have moved to sidebar flyouts, and reuse the same builder functions
- MyToken objects have been migrated to a new TokenCustomization object. 
- PC and Monster customizations have also been migrated to the new TokenCustomization object.
- We are (very nearly) ready to migrate all TokenCustomization objects to the cloud. We still need the cloud work.
- SidebarListItems are now addressed by id instead of path. 
    - This is a much cleaner implementation, more performant, less brittle, and avoids naming collisions (scene folders still need to be unique, but scene names and scene folder names can be the same).

# Demos

## Token Style
Here are all the styles in order of circle, square, virtual mini round, virtual mini square, no constraint.
Screen Shot 2022-06-17 at 7 44 19 AM

## Monster HP
The default is "average" which is what we always used to do. Here are 4 tokens that were placed using the "roll" setting.
Screen Shot 2022-06-17 at 7 49 40 AM

## Settings Panel Organization
The global token options are now in a flyout to make it consistent with other parts of the app. This also helps to clean up the settings panel a bit.
Screen Shot 2022-06-17 at 6 50 16 AM

## Token Customization
Player tokens, monster tokens, My Tokens, and My Tokens folders can now override any of the options that exist in the global settings. These have always been available as global settings and on placed tokens so it makes sense to allow them on token definitions as well. In addition to the options menu flyout, I've also included the image scale option here which will greatly ease the setup time for DMs that use images that need to be scaled slightly. 
At this level, the options are all drop downs instead of toggles because they we need to offer a way to force the value to true or false while also allowing to not force any setting. So the options become [true, false, not overridden].
Screen Shot 2022-06-17 at 6 51 44 AM

## D&D Beyond Portraits
On the left, you can see the list of races that DDB provides portraits for. On the right are all the Human portraits (which can be accessed by clicking the gear like any other token)
ddbPortraits

## Hover Text Flyouts
Instead of showing the hover text over the top of the sidebar, they are now displayed in a "flyout" to the left of the sidebar.
Screen_Shot_2022-06-05_at_9 55 41_AM

# Other things to note
### Token Options as a toggle vs select list
1. Global options are a toggle because these are not overriding anything. They are either on or off.
2. On a token definition (My Token, PC, etc.) have select lists because they can override the global settings by setting them to on or off, but we also want to allow them to use the global settings. So the options become "on", "off", or "whatever the global setting is". 
3. Placed tokens are toggles because they are either on or off. Once a token has been placed, those settings have been applied to the token and are essentially "locked" in place. We could change these to select list and just reference other overrides/global, but we don't do that now, and I'm not sure that we should. It would allow changing all placed tokens from square to circle by changing the global setting, but only if you didn't override it anywhere... I think that would feel confusing and disjointed.

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -102,6 +102,7 @@ Mousetrap.bind('q', function () {       //collapse/show sidebar. (q is next to t
 });
 
 Mousetrap.bind('esc', function () {     //deselect all buttons
+    console.log("Mousetrap.bind('esc'");
     stop_drawing();
     $('#select-button').click();
     close_token_context_menu();
@@ -110,6 +111,17 @@ Mousetrap.bind('esc', function () {     //deselect all buttons
     try {
         $( '.ui-draggable-dragging' ).draggable("option", { revert: true }).trigger( 'mouseup' ).draggable("option", {revert: false })
     } catch (whoCares) { }
+    let elementPreventingSidebarClose = $(".prevent-sidebar-modal-close");
+    if (elementPreventingSidebarClose.length > 0) {
+        if (elementPreventingSidebarClose.hasClass("sidebar-flyout") || elementPreventingSidebarClose.closest(".sidebar-flyout").length > 0) {
+            // it's a flyout... close it
+            elementPreventingSidebarClose.remove();
+        }
+        // any others that we want to close on esc?
+    } else {
+        // only close the sidebar if there isn't something on the screen explicitly trying to keep it open
+        close_sidebar_modal();
+    }
 });
 
 //menu specific shortcuts, select the nth element of menu when it's open
--- a/Load.js
+++ b/Load.js
@@ -73,6 +73,7 @@ let scripts = [
 	{ src: "ChatObserver.js" },
 	{ src: "DiceContextMenu/DiceContextMenu.js" },
 	{ src: "TokensPanel.js" },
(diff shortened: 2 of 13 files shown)


vlaminck:

PR #563, opened 2022-06-10
Fix: token offset when dragging without grid snap enabled

vlaminck:


vlaminck:

PR #564, opened 2022-06-10
fix: stop calling DM only function on player page

vlaminck:


vlaminck:

PR #566, opened 2022-06-13
Add: Token health options

Rishabh2:
Added an option to the Token Settings menu for monster hp generation type. The option is a dropdown with three choices: Average, Roll, and Max. New tokens are generated according to the setting.

image

### DIFF
--- a/Settings.js
+++ b/Settings.js
@@ -339,7 +339,6 @@ function redraw_settings_panel_token_examples() {
 		} else {
 			hpnum = 10;
 		}
-		console.log(`HP Num is ${hpnum}`);
 		let tok = new Token(default_options());
 		tok.options.size = items.width();
 		tok.options.max_hp = hpnum;
--- a/Settings.js
+++ b/Settings.js
@@ -3,61 +3,78 @@ const token_setting_options = [
 		name: 'hidden',
 		label: 'Hide',
 		enabledDescription: 'New tokens will be hidden from players when placed on the scene',
-		disabledDescription: 'New tokens will be visible to players when placed on the scene'
+		disabledDescription: 'New tokens will be visible to players when placed on the scene',
+		type: 'Toggle'
 	},
 	{
 		name: 'square',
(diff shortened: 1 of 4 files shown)


hypergig (commented):
Looks nice and clean to me. I have wanted this for a long time. @vlaminck how do you want to deal with the new settings type in settings panel?

vlaminck:
I've started a conversation in Discord about it, but this is maybe a better place for that conversation. https://discord.com/channels/815028457851191326/832199739398946817/985882199087804416

@Azmoria and I have also started cleaning up the `token_setting_options` object, and I recommend we change the structure to something like this
```
{
  name: 'defaultmaxhptype',
  label: 'Monster Max HP Calculation',
  options: [
    { label: 'Average', value: 'avg', description: 'Monster Max HP will be set to average HP amount' },
    { label: 'Roll', value: 'roll', description: 'Monster Max HP will be rolled individually' },
    { label: 'Max', value: 'max', description: 'Monster Max HP will be set to maximum HP amount' }
  ],
  type: 'dropdown'
},
{
  name: 'hidden',
  label: 'Hide',
  options: [
    { label: 'Hidden', value: true, description: 'The token is hidden to players.' },
    { label: 'Visible', value: false, description: 'The token is visible to players.' }
  ],
  type: 'toggle'
},
{
  name: 'tokenStyleSelect',
  type: 'dropdown',
  label: 'Token Style',
  options: [
    { label: 'Circle', value: 'circle', description: 'The token is round, we set other things, yadda yadda yadda' },
    { label: 'Square', value: 'square', description: 'The token is a square, we set other things differently, yadda yadda yadda' },
    ...
  ]
},
```

If we decide to go with this structure, I think it would make sense for that change to be part of #561.
At which point, it might make sense to make this PR against my branch since @Azmoria and I are treating it like a dev branch anyway. We might as well include this in there, too.

Rishabh2:
Closing to merge to vlaminck's branch

PR #569, opened 2022-06-19
Constrain tokens to map #469

### LINKED ISSUE #469: Tokens are able to move off the map
If you select a token and hold the arrow key, the token will not stop at the edge of the map. This can make it impossible to access the token.

pensan:
Fixes #469 

Performance on the `Token#drag` method is mediocre, since the token is still moved to the maximum allowed position with each fired event.

I fear that stuff may break if we return out of the `Token#drag` method when we are outside of the scene.

For configuration, I added `Token#SCENE_MOVE_GRID_PADDING_MULTIPLIER` which defines how many grid fields a token can be moved outside of the scene.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -259,7 +259,6 @@ class Token {
 		// Sets how far tokens are allowed to move outside of the scene
 		const movePadding = this.options.size * this.SCENE_MOVE_GRID_PADDING_MULTIPLIER;
 		// Stop movement if new position is outside of the scene
-		console.warn(this.sceneBoundaries.left - movePadding);
 		if (
 			top  < (this.sceneBoundaries.top - movePadding)    || 
 			top  > (this.sceneBoundaries.bottom + movePadding) ||
@@ -1199,6 +1198,9 @@ class Token {
 				x: 0,
 				y: 0
 			};
+
+			const movePaddingWhileDragging = self.options.size * self.SCENE_MOVE_GRID_PADDING_MULTIPLIER;
+
 			tok.draggable({
 				scroll: false,
 				stop:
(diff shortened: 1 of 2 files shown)


Token.js
```diff
@@ -49,6 +52,14 @@ class Token {
 		if (typeof options.conditions == "undefined") {
 			this.options.conditions = [];
 		}
+
+		// Store scene boundaries when token is created
```
vlaminck, line 56: I would convert this to a function instead of defining it at the time of creation. It's _probably unlikely_ that the scene size would change, but not impossible.

vlaminck (commented):
This mostly works well, and will be a huge quality of life improvement! I think the math is just a _little_ off for different sized tokens, though. It seems to be function of the token size as seen here: https://imgur.com/a/8ctdyHF

Token.js
bain2236: Should also probably live in the sceneshandler. Feels a bit weird that the token class itself knows the boundaries of the scene. Same for the scene_move_grid_padding_multiplier

bain2236:
You're still able to place a token outside of the constraint, however moving it then does snap it to the edge of the constraint. Would be nice to check if the "place_token_at_center" or whatever the actual function is called is going to place it outside of the constraint and place it at the edge if that's the case. Might be more hassle than it's worth for an edge case

pensan:
I refactored the my code. It should work now.

Regarding:
> You're still able to place a token outside of the constraint, however moving it then does snap it to the edge of the constraint. Would be nice to check if the "place_token_at_center" or whatever the actual function is called is going to place it outside of the constraint and place it at the edge if that's the case. Might be more hassle than it's worth for an edge case

With the news changes to then "TokensPanel.js" it is no longer possible to place tokens outside of the map:
https://github.com/cyruzzo/AboveVTT/blob/accceda5371d16d20af1d33750636c26bbfc67d1/TokensPanel.js#L547-L557

Issue #258, opened 2022-03-13
Limit token size


vlaminck:
There should be some upper bound on token sizes. We had a user with performance issues that were cause by a token that was set to 14533333 px

RuneKeeperOfTheUr:
PLEASE FOR THE LOVE OF GOD NO! 

STOP REMOVING FEATURES AND ADDING REALLY CRAPPY LIMITATIONS! First no tokens outside the map and now you're trying to screw over custom token size? HTF are we supposed to run Spelljammer properly if you do that? _**FFS -NO-.**_ This is an AWFUL idea. **PLEASE DO NOT DO THIS.**

vlaminck:
> STOP REMOVING FEATURES AND ADDING REALLY CRAPPY LIMITATIONS! ... _**FFS -NO-.**_ This is an AWFUL idea. **PLEASE DO NOT DO THIS.**

Calm down. There's no need to yell or belittle ideas. 

> First no tokens outside the map...

Tokens can go outside the map, just not as far as they used to. We did this because tokens were getting lost at locations the browser couldn't scroll to. If you can't place tokens outside the map at all, that's a bug. Please raise it as such. If you want to be able to place tokens further off the map, ask for that as a feature request. 

> ... and now you're trying to screw over custom token size?

My understanding is that this ticket is meant to prevent tokens from being excessively larger than the map because that causes performance issues. If you need tokens that are larger than your map, you should consider using that token as the map itself, or change your scale so that your token fits within the map. In your example of Spelljammer, you simply need whatever space-looking image you want for your background, and then use the wizard to make sure your grid size can accommodate large ships. People do it all the time with ships on the ocean, and spelljammer ships are no different. If you need help setting up a map to accommodate large ships, there are a lot of friendly and helpful people in the Discord that can help you.

pensan:
For reference: if we want to allow tokens to move more then '1 grid size' step outside of the scene, there is configuration value to specify it:

`SCENE_MOVE_GRID_PADDING_MULTIPLIER` (https://github.com/cyruzzo/AboveVTT/pull/570/files#diff-ab4097c962c4e0be21f37f7b9beedf07e976bafd4d6c91fcaece8405cfec4096L46)

I assume we could set this to 2 or 3 without causing to much problems.
This would allow tokens to move twice or tribble their size outside of the map.

Azmoria:
[...]

PR #570, opened 2022-06-19
Add limits for token size #258

pensan:
Tried to add scaling limits (#258).

Used `25px` as min size, and `1000px` as max size in the Token class. Please change the values if you find other values more appropriate.

Also added a clamp to the `Token Image Scale` slider that matches the HTML form input values.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -242,14 +242,17 @@ class Token {
 
 		this.update_from_page();
 
-		if(this.isLineAoe()){
+		if(this.isLineAoe()) {
 			// token is not proportional such as a line aoe token
 			this.options.gridHeight = Math.round(newSize / parseFloat(window.CURRENT_SCENE_DATA.hpps));
 		}
-		else{
+		else {
 			this.options.size = newSize;
 		}
-		this.place_sync_persist()
+
+		this.place_sync_persist();
+
+		this.prepareWalkableArea();
 	}
 
 	imageSize(imageScale) {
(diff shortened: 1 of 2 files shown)


pensan:
Closes #258

Token.js
```diff
@@ -135,6 +138,9 @@ class Token {
 	}
 
 	size(newsize) {
+		// Clamp token size to min/max token size
+		newsize = clamp(newsize, this.MIN_TOKEN_SIZE, this.MAX_TOKEN_SIZE);
```
vlaminck, line 142: I think we should clamp the upper bound to `$("#scene_map").width()` as long as that exists

bain2236:
Shouldn't the max of a token be the same size as the limits going on in your other branch? If they were both merged in you could have a token largeer than the map and not be moveable. Currently I can end up with something like this which feels strange as a max size. it might also be nice to show hte user when they've hit the max size for a token within the context menu token adjustments flyout. Although I think there's some changes in the big token PR that fixes some things in that menu
image

pensan:
I updated the logic to better fit the needs.

The scaling is now adjusted when the token size changes. 
Allowing for a max token size of `$("#scene_map").width()`.

Scaling is then adjusted to the new token size.
A token of 'full map with' size will only be able to have a image scale of "1" or smaller.

If the token size is changed to a larger value, the scale is reset to "1".

The DOM inputs are also updated with new max values when token size changes.

pensan:
The conflict should be resolved by using both revisions.
We need all 3 variables

pensan:
Rebased and refactored to work with the new Token/TokenMenu changes.

Also fixes a bug in relation to https://github.com/cyruzzo/AboveVTT/issues/469 which occurred when a token size was changed after initial load.

PR #571, opened 2022-06-20
fix ruler fadeout not clearing temp canvas

vlaminck:
Somewhere along the way, these functions were duplicated, and each were updated in different ways. This cleans that up


vlaminck:

PR #575, opened 2022-06-20
Show warning when Beyond20 browser extension is enabled

pensan:
Adds a simple check to detect Beyond20 when removing the splash screen and shows an alert for now.


pensan:
Refactored the code a little bit into its own module.

Should show a message on load now if any DOM elements with class names starting with `beyond20` are found.

We could delay the check if necessary, but it worked on may local setup.

cyruzzo:
sorry my fault.... I see that some people are using beyond20 successfuly with abovevtt so even if I personally requested this I'm closing it.

PR #577, opened 2022-06-21
Add option to Group Tokens in Right-click menu

Rishabh2:
When selecting multiple tokens, tokens can be grouped so that clicking any one of them selects all of them.

Grouped tokens act as a single unit when clicked, right-clicked, or dragged.

image

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -59,7 +59,7 @@ class WaypointManagerClass {
 			backgroundColor: "rgba(255, 255, 255, 0.7)"
 		}
 	}
-	
+
 	resetDefaultDrawStyle(){
 		this.drawStyle = {
 			lineWidth: Math.max(25 * Math.max((1 - window.ZOOM), 0), 5),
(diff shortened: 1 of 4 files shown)


bain2236:
Nice. For icons we usually use https://fonts.google.com/icons?selected=Material+Icons I believe.
I'd also move the grouping down in the menu, stat blocks should always be at the top.

Rishabh2:
> Nice. For icons we usually use https://fonts.google.com/icons?selected=Material+Icons I believe. I'd also move the grouping down in the menu, stat blocks should always be at the top.

Icons have been updated and button has been moved

vlaminck:
If we choose to support this feature, I think it's more of a way to combine multiple tokens into a single token rather than a shortcut for selecting tokens. For the benefit of anyone that wasn't in the Discord conversation, the idea here is to allow grouping tokens together in a way that allows them to act as one. For example, a horse pulling a wagon could be grouped together and would then act as a single token rather than 2 separate tokens.

vlaminck:
relates to https://github.com/cyruzzo/AboveVTT/issues/579

Rishabh2:
The only remaining conlifcts are whitespace, the version of each conflict from this Pull Request can be selected to resolve the conflict.

PR #598, opened 2022-07-11
Fix aoe rotation on load

### LINKED ISSUE #591: token.place() do not correctly set the rotation of the new AoE tokens ( was: new AoE Tokens don't rotate on player view)
As reported by seraphim with a nice video

https://drive.google.com/file/d/13e96lBxu_hrv80bDJWz3iG5DJYe62AqD/view?usp=sharing

If as DM you rotate a cone or line.. they don't rotate on player view..
this is probably a problem on "rendering" the token.

Azmoria:
Oops this should Fix #591 for initial loading

### DIFF
--- a/AOETemplates.js
+++ b/AOETemplates.js
@@ -276,8 +276,8 @@ function build_aoe_token_options(style, shape, countGridSquares, name = "") {
     return options
 }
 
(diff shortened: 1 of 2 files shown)


cyruzzo:
yep. works for me. thanks

PR #603, opened 2022-07-13
Fix selection box rotation

Azmoria:
Another missed part of #591 

This doesn't fix the ruler. 

I also ran into another issue where long line AoEs rotated still gets stopped by the token bounds as if it wasn't rotated. It may be worth and easy fix of just rotating the whole token for line aoes instead of the token-image since it likely won't have health, conditions etc as an AoE. Assuming that fixes this - not sure it would the bounds that contain the tokens might be based off something else.


bain2236:
That would probably work. Could also change the boundary to work off the centre of the token rather than the edge. The same function could also be used to fix the measuring

PR #614, opened 2022-07-20
Fix: Build hidden sidebar for players so sound can be played properly

Azmoria:
This builds the sound panel for the players so the DM can control it. It seems it wasn't being built for players at all and sounds couldn't be played.

May also work better with #613 due to some syncing issues. 

Credit to @pensan for digging through this as well

### DIFF
--- a/SidebarPanel.js
+++ b/SidebarPanel.js
@@ -26,12 +26,12 @@ function init_sidebar_tabs() {
     update_pclist();
   }
 
-  if (window.DM) { // players will eventually have audio volume control in the settings panel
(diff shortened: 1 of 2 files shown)


pensan:
Not sure if we need the changes in the `SoundPad.js` since they 'only' create UI elements, which are only visible to the DM.

Azmoria:
> Not sure if we need the changes in the `SoundPad.js` since they 'only' create UI elements, which are only visible to the DM.

Right you are. oops I'll fix that

Issue #597, opened 2022-07-11
Fog get resetted after editing a scene


cyruzzo:
If you edit an already existing scene (for example cause you want to change the snapping settings),  fogs get cleared after you press save (may need to check if the issue is client-side or on the back-end).

PR #616, opened 2022-07-21
Fix fog get resetted after editing a scene  #597

pensan:
The server adds some FogData when the `update_scene` event is fired on the server and `isnewscene` is set.

```
if(recvMessage.data.isnewscene){
  delete recvMessage.data.isnewscene;
  promises.push(ddb.put({ // also initialize fog data
    TableName: process.env.TABLE_NAME,
      Item: {
        campaignId: campaignId,
        objectId: "scenes#"+recvMessage.data.id+"#fogdata",
        data: [[0, 0, 0, 0, 2, 0]],
      }
  }).promise());
}
```

Currently, the `ScenesHandler.persist_scene` function is always called with `isnew = true` in the ScenesPaneljs.
This will lead to scene updates are always treated as 'new' which causes the server to add the reveal.
When later the `custom/myVTT/fetchscene` is fired, the data send by the server will always include the 'reveal all' data.

This PR will set the `isnew` flag to be always false when clicking the save button.
I put it into a new variable so we can easily extend it with some logic for 'new' scenes.

I am however not sure, what side effects this may cause.

In my tests the fog works know, also for new maps where the `isnew` flag should be true.
Maybe someone with more knowledge on how exactly the fog data is used can provide me some insights?

### DIFF
--- a/ScenesPanel.js
+++ b/ScenesPanel.js
@@ -430,7 +430,8 @@ function edit_scene_dialog(scene_id) {
 		}
 
 		if(window.CLOUD){
-			window.ScenesHandler.persist_scene(scene_id,true,true);
+			const isNew = false;
+			window.ScenesHandler.persist_scene(scene_id, isNew);
 		}
 		else{
 			window.ScenesHandler.persist();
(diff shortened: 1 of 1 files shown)


cyruzzo:
ok.. it seems to work.

PR #619, opened 2022-07-27
Fix/QoL token size changes on grid change

Azmoria:
When the grid is changed the tokens don't change with it. This fixes that. QoL for when tokens have already been dropped before noticing the grid.

Example:
https://www.youtube.com/watch?v=8pBsB_uSE2g


cyruzzo:
nice

Issue #147, opened 2021-11-11
Scenes creation and importing needs to be more intuitive for new users


cyruzzo:
1) If pressing "super mega wizard" on a "pre-configure/ aligned map" we should give a warning and let use rethink that.

2) DM_MAP should be an advanced feature. PLAYER_MAP should maybe be something like "Import image from Url"

3) Maybe the "Super Mega Wizard" should be hidden under an "advanced tab" IF the map is already aligned

hypergig:
I almost would want to see a quick add input box that takes a url and creates a scene straight from it in one step.

Then you can go back into the settings and tweak the scene if necessary.

Etus12:
Another thought I had is to add the import/export buttons to the scenes panel at the top (along/instead Settings tab)

Azmoria:
I think we're at a good point for this at the moment - we very rarely get questions about it anymore

PR #621, opened 2022-07-28
Scenes panel adjustments

Azmoria:
- DM map row only appears if enabled
- Update placeholder text of map lines
- Window is moveable
- Sized to fit content as menus appear
- When clicked into the URL lines and Name will select all text automatically for QoL.
- Slight appearance adjustments

Addresses Part 2 of #147 

Example:

https://www.youtube.com/watch?v=6p40ompCV-Q


cyruzzo:
nice step in the right direction

PR #654, opened 2022-08-19
Improve ruler performance while moving a token in player view. Improve token selection/movement performance in firefox.

Azmoria:
I'm pretty sure this fixes at least a portion of the player view drag while measuring performance issue.

See example video - left is current, right is this PR:

https://www.youtube.com/watch?v=LX9zVdeZr0c

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -383,6 +383,32 @@ function is_token_under_fog(tokenid){
 		return false;
 }
 
+function check_single_token_visibility(id){
+	console.log("check_single_token_visibility");
+	if (window.DM || $("#fog_overlay").is(":hidden"))
+		return;
+	var canvas = document.getElementById("fog_overlay");
+	var ctx = canvas.getContext("2d");
+			var auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
+			var selector = "div[data-id='" + id + "']";
+			let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
+			if (is_token_under_fog(id)) {
+
+				$(selector).hide();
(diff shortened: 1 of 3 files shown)


cyruzzo:
this seems better. Before I look into it... I surely remember that firefox was having more issues with dragging and measuring.
Have you tested it with firefox ?

Azmoria:
So I'm not having issues with firefox when dragging. It seems to be when the token is dropped. At this point it loops through updating the tokens and for some reason on firefox this is taking a huge amount of time on maps with lots of tokens. This delays the ruler disappearing and seems to be unresponsive for a short while depending how many tokens are on the map.

This is also happening on live it seems so I don't think this is newly introduced and seems to have to do with looping through the token objects in token.js  around line 1477. Seems to take a really long time no matter how I adjust it. Might be a separate issue to the lag seen in the above video.

```						
if (self.selected) 
	for (id in window.TOKEN_OBJECTS) {
		if ((id != self.options.id) && window.TOKEN_OBJECTS[id].selected) {
			var curr = window.TOKEN_OBJECTS[id];
			var ev = { target: $("#tokens [data-id='" + id + "']").get(0) };
			curr.update_and_sync(ev);
		}
	}
}
```

Azmoria:
So I've just pushed up some fixes - it was looping through all tokens for each token that was selected when trying to see if they were in fog. This now just checks single tokens on loop.

Azmoria:
Here's an example of the issue on live and fix applied from my testing with firefox

https://www.youtube.com/watch?v=jZwlJFlMluE.

cyruzzo:
ok, totally merging this as it surely improve stuff..
I'd like to still wait a bit before we enable by default the token measuramente while dragging..

PR #660, opened 2022-08-25
Save combat tracker status of tokens to cloud. Persist CT through session scene swaps. Option to add tokens from CT to scene with the same ID if not present already. Sync tokens with the same id across scenes. Copy paste from other scenes (different IDs).

### LINKED ISSUE #301: Allow tokens to be copied from one scene and pasted on another
There are two parts to this.

1. We don't currently allow player tokens to be copied. We would need to allow player tokens to be copied, but not pasted.
2. We only hold a reference to the copied token id. We don't actually duplicate the tokens until they are pasted. We would need to either duplicate the token objects when they are copied, or when changing scenes.

### LINKED ISSUE #313: Persist Combat Tracker in the cloud

Azmoria:
Closes #313 
And Closes #301 

Fixed and new PR of #653 and #658 (I was test merging and fixing conflicts without making a new branch first oops) with less commits

Save combat tracker initiative per scene using token options.

Also persist the combat tracker throughout a session - this means it will bring other scenes combatants with you to the next. Easy to clear still with the clear button and won't erase the previous scenes save.

This allows for combat across scenes during a session while retaining individual scene saves using token options ct_show (existing), init, current and round. I tried round with scene but that forces a refresh for players so I swapped it to the current token.

https://www.youtube.com/watch?v=DAZ_rHMlCZ4

Edit: added an `place` button that replaces the find button when a token isn't on the current scene.

https://user-images.githubusercontent.com/65363489/186967037-1626c994-ac5d-4b1a-b7a6-34861121c496.mp4

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -111,7 +111,7 @@ function init_combat_tracker(){
 
 			window.StatHandler.rollInit($(this).attr('data-monster'),function(value){
 				element.find(".init").val(value);
-				ct_reorder(false);
+				setTimeout(ct_reorder(), 500);
 			});
 			setTimeout(ct_persist,5000); // quick hack to save and resync only one time
 		});
@@ -165,7 +165,7 @@ function init_combat_tracker(){
 					
 			window.StatHandler.rollInit($(this).attr('data-monster'),function(value){
 				element.find(".init").val(value);
-				ct_reorder(false);
+				setTimeout(ct_reorder(), 500);
 			});
 			setTimeout(ct_persist,5000); // quick hack to save and resync only one time
 		});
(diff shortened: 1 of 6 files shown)


cyruzzo:
I still need to review this.. but my first question is:

since you moved it to "sessionStorage".. it's lost once the DM closes the page right ? This is not something people will like.

Azmoria:
> I still need to review this.. but my first question is:
> 
> since you moved it to "sessionStorage".. it's lost once the DM closes the page right ? This is not something people will like.

It's also saved on the tokens in ct_show (which already exists) and a new init options (for initiative). So each token on the scene will have it's combat status loaded in. When they move to another scene the tokens saved there will be added back if they are using multiple scenes. The session data is mostly for updating the player view and getting the data to move between scenes.

Azmoria:
Syncing tokens added from the combat tracker across scenes was already in just added to the name change. Copy paste from other scene was a 1 or 2 line change with this build so I added it. I can easily take that out again if you'd like.

Azmoria:
Updated example with some explaination.

https://youtu.be/z02Ii8Bvhx4

PR #672, opened 2022-09-09
Fix - add missing monster vulnerabilities

Azmoria:
Just noticed this tonight while playing with a gloomstalker and skeletons in combat that vulnerabilities are missing. This should fix that. 

Do we also want to consider hiding skills and saving throws when there aren't any to match dndbeyonds format?

    --------->


vlaminck:
> Do we also want to consider hiding skills and saving throws when there aren't any to match dndbeyonds format?

I would say yes. Thanks for catching these!

vlaminck (approved):
This code looks exactly as I would expect it. Very nicely done 👍

PR #675, opened 2022-09-14
Load in the Combat Tracker local storage on first load of 0.80 and save it to tokens/cloud

Azmoria:
As natemoonlife pointed out currently the beta doesn't load in local storage from the CT. This could be frustrating for users who ended mid combat.

This will load in the local storage people currently have on the first load. Next version I think we can delete the local storage but this allows people to go back to 0.79 and keep their local CT if there is a problem with the beta. Can always change this to delete the local storage instead if you would prefer.

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -729,7 +729,9 @@ function ct_update_popout(){
 }
 
 function ct_load(data=null){
-	
+
+
+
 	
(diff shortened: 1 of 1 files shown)


cyruzzo:
no need to delete it now.. 
it's ok as it is.

Issue #391, opened 2022-04-05
Journal Panel Remake


cyruzzo:
The journal panel needs some love.

Minimal Missing features:
- UI style sucks. It should have collapsable chapters.
- rename notes / chapters
- reorder notes / chapters

The note editor in the popup is not always sized like the popup

Azmoria:
I think the minimum requirements have been met here. We can reopen/create a new issue if we want to make further enhancements.

PR #698, opened 2022-10-10
#391: Added basic UX to Journal Panel

SJ-Jordaan:
Add Collapse/Expand to chapters
Add Sorting to chapters and notes
Add basic styling
Add renaming of chapters/notes

Remaining TODOs:
View/Edit modal
Delete confirmation


cyruzzo:
Sorry for the late reply.
1) Checking your PR it seems to me that on "player side" the journal no longer show the visible notes
2) The renaming feature is nice. But when I press the "edit icon" near to the note I expect to edit the note. Not to rename.

PR #701, opened 2022-10-18
Fix audio sync on player resize window

### LINKED ISSUE #700: Audio list doesn't rebuild on player view resize
Edit: I put up a quick simple fix for this if we want to do it before Audio 2.0 https://github.com/cyruzzo/AboveVTT/pull/701

Since the sidepanel rebuilds for players when resizing enough it looks like the audio info gets wiped out too and doesn't get rebuilt. This causes sound to stop for players and won't start again. 

There's been quite a few audio issues. We can try and fix this or look to test hypergigs audio if he's ready for it. I was doing some UI stuff on it not too long ago and was talking to him about it here: https://discord.com/channels/815028457851191326/930926587392692335/1024820620875284601

Azmoria:
Fixes #700 

This is a quick simple fix for this if we want to include it. Request a sync when resizing is triggering a rebuild of the sidebar.

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -2263,6 +2263,7 @@ function init_character_page_sidebar() {
 			init_sheet();	
 			inject_chat_buttons();
 			init_zoom_buttons();
+			window.MB.sendMessage('custom/myVTT/syncmeup');
 		}
 	}, 1000);
 }


cyruzzo:
"syncme up" triggers a lot of messages and it's quite expensive . I need to test this more to make sure that we don't get too much messages exchanged

Azmoria:
makes sense can always wait until audio 2.0 anyway

PR #715, opened 2022-10-26
Add investigation and insight passives to player rows

Azmoria:
Requested in discord by Renato and I think it would be pretty useful. 

Example:

image


Azmoria:
Closing as I believe after feedback on discord #716 provides more info and is a better fit

Azmoria:
Closing as after feedback on discord I believe #716 to be a better option.

PR #732, opened 2022-11-03
Potential Fix - jitsi embed

### LINKED ISSUE #713: Video Jitsi 429 too many requests in embedded iframe
Only thing I could find on this was https://community.jitsi.org/t/embedded-jitsi-meet-returns-429-too-many-requests/118043

We may be too big for their free service for embedded video now? Not sure this is fixable but raising the issue anyway. We may want to make the UI popout only if that's the case.

Azmoria:
So I found another official jitsi implementation. 8x8 offers free video that uses jitsi and acts as a relay. 

Found the server here:
https://jitsi.github.io/handbook/docs/community/community-instances/

Now I don't know if they have the same thing going on where it will lock us out when more people are using it I couldn't find any explicit limit on searches but it's currently working for me.

If we find this server also has the same limit - unless there's another free server amongst the community servers I don't see us being able to fix this.

Maybe fixes #713

### DIFF
--- a/Jitsi.js
+++ b/Jitsi.js
@@ -18,7 +18,7 @@ function create_jitsi_button() {
 }
 
 function init_jitsi(tileLayout) {
-	const domain = 'meet.jit.si';
+	const domain = '8x8.vc';
 	const options = {
 		roomName: 'aboveVTT-' + find_game_id(),
 		width: '100%',


Azmoria:
I think I found the limits to this and it won't work for us. It's just jitsi as a service official server - which does have costs for higher monthly users.

Azmoria:
Actually now I'm not sure

> 8x8 Video Meetings, powered by Jitsi, is part of [8x8 X Series](https://www.8x8.com/x-series), an enterprise-grade cloud communications, video collaboration and contact center solution built on a modern, open cloud technology platform, and is also available as a **free and unlimited** standalone service at [8x8.vc](https://8x8.vc/)

PR #743, opened 2022-11-10
Fix - monster statblock styling in player view doesn't match DDB style

Azmoria:
This adds classes from the character sheet version of monster stat blocks and makes a few edits in the css to match the styling.

Before
image

After
image


vlaminck (approved):
Thank you! This was an ugly one

PR #744, opened 2022-11-11
Fix - only change token combat status on hide/reveal if already in combat (stops them from being added to combat on next scene load)

Azmoria:
So currently even if `ct_show` is undefined when hiding or showing a token it sets it to true or false. If it's undefined we want to leave this undefined until it's added to the combat tracker at which point it will determine which it should be. If it's not undefined we allow it to swap appropriately.

The reason for this is if it's combat status `ct_show` is defined next time a scene is loaded they will be added to the combat tracker. Like combatants in combat across scenes due to `ct_show` being true or false.


vlaminck (changes requested):
This would be easier to understand if you changed it to be 
```
if(this.options.ct_show !== undefined){
  this.options.ct_show = trueOrFalseOrWhatever;
}
```

It works as-is, but I would probably delete the if block here because it seems pointless. Maybe add a small comment explaining it, too?

Azmoria:
Yeah that makes sense I'll swap that and add a comment about why it's required.

vlaminck (approved):

PR #746, opened 2022-11-18
QoL - Disable ping when not on select tool (stop pings when drawing polygons and other drawings)

Azmoria:
This returns the double click function if not on the select tool for pings. I've found when drawing/fog polygons especially it pings frustratingly often. This should reduce messages and avoid unnesecarry pings.


vlaminck (approved):
very nice

PR #756, opened 2022-12-12
QoL - Allow token image split urls on `, ` (comma space). Allows mass add of gdrive images easily or other images with comma seperation.

Azmoria:
Gdrive gives multilinks a comma seperated this would allow quick addition of those links to tokens.

Example video:
https://discord.com/channels/815028457851191326/815028716040224818/1051686680551235604

### DIFF
--- a/SidebarPanel.js
+++ b/SidebarPanel.js
@@ -216,7 +216,10 @@ class SidebarPanel {
           if(imageUrl.startsWith("data:")){
             alert("You cannot use urls starting with data:");
           } else {
-            imageUrlEntered(parse_img(imageUrl));
+            let imageUrlSplit = imageUrl.split(', ');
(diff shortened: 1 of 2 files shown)


vlaminck (approved):

SidebarPanel.js
```diff
@@ -216,7 +216,10 @@ class SidebarPanel {
           if(imageUrl.startsWith("data:")){
             alert("You cannot use urls starting with data:");
```
vlaminck, line 217: Looking at this one again, I think we need to check for `data:` on all of them. Either that, or we need to update the `parse_img` function to check for that. I don't expect people to try sneaking any in, but we better guard against it just in case.

vlaminck (changes requested):

vlaminck (approved):

PR #779, opened 2023-02-05
Add Prev button to combat tracker

### LINKED ISSUE #742: Add a "previous" button to the combat tracker
Similar to the "next" button that already exists, but allows the DM to move the combat tracker turn back one. If the DM accidentally moves forward too soon, the only way to go back is to press the next button until it gets back to the previous turn. They then have to reduce the combat turn by one. This is super tedious, and a "previous" button would be really nice to have.

Azmoria:
Closes #742

Example:

https://user-images.githubusercontent.com/65363489/216809375-d4123b86-18b7-4f77-85d4-1a7b5d81e350.mp4

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -219,7 +219,7 @@ function init_combat_tracker(){
 		ct_persist();
 	});
 	
-	next=$("<button id='combat_next_button'><u>N</u>EXT</button>");
+	next=$("<button id='combat_next_button'>NEXT</button>");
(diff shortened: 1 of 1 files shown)


Azmoria:
Yeah, we can do that. I'll do it shortly.

Azmoria:
Ok this should stop the button from going into rounds < 1 now since we start at round 1.

vlaminck (approved):

PR #780, opened 2023-02-05
Allow players to end their own turn.

### LINKED ISSUE #741: Allow players to end their own turn in the combat tracker
This would allow players to move the combat tracker forward when they are done with their turn rather than making the player tell the DM that their turn is over and then having the DM press the "next" button. 

* This new "end turn" button should only be present when it is actively the player's turn.
* The DM should get some sort of notification that the player has ended their turn. Examples of this notification could be:
  * A prompt at the top of the DM screen or over/next to the combat tracker
  * Flash the combat tracker some how, or flash the highlighted row a few times
  * Send a message in the gamelog

I'm not sure what the best way to notify the DM is, but it should get their attention when a player ends their turn so they don't just sit there waiting for the player.

Azmoria:
I combination with @cyelis1224's work (https://discord.com/channels/815028457851191326/930926587392692335/1056453785603559434) I think this will close #741

Example
https://user-images.githubusercontent.com/65363489/216824783-429a4e01-ecc4-4e21-a020-02cd30f5af1e.mp4

Edit: To do 

- [x] Add to popout
- [x] Extend draggable corner down like the DM's view

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -288,7 +288,10 @@ function init_combat_tracker(){
 
 	endplayerturn=$('<button id="endplayerturn">End Turn</button>');
 	endplayerturn.click(function(){
-		window.MB.sendMessage('custom/myVTT/endplayerturn');
+		let data = {
+			from: window.PLAYER_ID,
+		}
+		window.MB.sendMessage('custom/myVTT/endplayerturn', data);
 		$("#endplayerturn").toggleClass('enabled', false);
 	});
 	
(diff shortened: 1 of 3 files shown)


vlaminck (changes requested):
I think you need to use `addClass` and `removeClass` instead of `toggleClass`. The button stays up all the time for me.

I also think you should pass the `window.PLAYER_ID` in the `custom/myVTT/endplayerturn` message, then have the DM validate that it is currently their turn just to be safe. If the DM clicks next before they receive the event from the player, it will move the combat tracker forward twice.

avtt-pr780-allwaysAvailable

vlaminck:
I was using Firefox if that matters

vlaminck (approved):

PR #785, opened 2023-02-06
Since hp bars aren't always 'bar' relabel it to HP Meter

Azmoria:
Open to other naming suggestions as well


vlaminck (approved):
I like the word "meter" for this

PR #787, opened 2023-02-06
 QoL - Add cancel wizard button to mega wizard & Wizard dialogs are dragable

Azmoria:
Adds a basic cancel wizard button - until we decide what other improvements we want to make I've just placed it at the top of the screen for ease of access. 

https://youtu.be/-hahDotEUFE


vlaminck (commented):
This will be handy to have, but could we just add the cancel button next to the "Press when it's good enough" button?

Azmoria:
The dialog gets cleared and new stuff added to it. We could readd the button each time but I'd have to do the layout for each. I can change it though if you think it'd be better. I think being able to cancel at any step is ideal.

vlaminck:
> I also plan to do a pass of the wizard UI at some point. Just was building the workings and coming back for UI once we have what we want there.

I wonder if we should just prioritize that work and rebuild the whole thing right now.

vlaminck:
All right, after discussing on Discord, let's go forward with these changes, and use them to inform the final UI updates. I'll make an Issue for that work so we can track it.

vlaminck (approved):

PR #788, opened 2023-02-07
QoL - Add minor adjustment bars to grid wizard

Azmoria:
For when maps you just can't quite get to line up these bars should allow for more precise adjustment when needed. We can pretty them up as we adjust the UI.

https://www.youtube.com/watch?v=WOtus3jyNdk


vlaminck (approved):
Similar to #787 this will be used to inform the final UI rebuild of the grid alignment wizard

PR #789, opened 2023-02-07
Fix - game log popout wasn't loading sometimes.

Azmoria:
This adds a backup popout game log cleanup in a timeout. Until I can dive into why it's not triggering the onload event sometimes this is a quick fix.


vlaminck (approved):
Since this is just a workaround, can you create an Issue to track the work necessary to actually fix it?

PR #794, opened 2023-02-08
beta fix: inconsistent dice roll button injection

vlaminck:
There are 2 main parts to this

1. Custom dice roll buttons were not being injected when the DM opened character sheets.
2. Updating the notes field didn't always update the button

### DM
To get this working on the DM side of things, I had to update `getCharacterStatModifiers` to dig through and parse `window.PLAYER_STATS` as well as dig into the currently open iframe to find the proficiency bonus.  
I also had to update `replaceModifiersInSlashCommand` to allow parsing on pages other than the character page  
Finally, I had to pass the iFrame to `observe_character_sheet_dice_rolls` when the character sheet opened

### Parse on update
To inject the buttons when the notes field was updated, I had to update the MutationObserver to actually iterate through the mutations looking for updates to the elements we care about. I pulled the injection logic into a new `inject_dice_roll` to DRY up that work.

### related
https://github.com/cyruzzo/AboveVTT/pull/706

### DIFF
--- a/DiceRoller.js
+++ b/DiceRoller.js
@@ -521,20 +521,39 @@ function replace_gamelog_message_expressions(listItem) {
 }
 
 function getCharacterStatModifiers() {
-    if (!is_characters_page()) {
-        return undefined;
+    if (is_characters_page()) {
+        let stats = $(".ddbc-ability-summary__secondary");
+        return {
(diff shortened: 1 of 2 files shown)


Azmoria (approved):
Looks good - works for DM and players without issue.

PR #795, opened 2023-02-09
add peerjs and pass cursor location across peer to peer connections

vlaminck:
# What

1. Create peer to peer connections using `peerjs` for passing high throughput, low priority events between all connected players. 
2. Stream cursor position to all connected peers.
3. Stream rulers to all connected peers.
4. Add an easy way for DMs to temporarily enable/disable cursor/ruler streaming 
5. Add a settings to enable/disable cursor and ruler streaming

# Why

1. Streaming high throughput events over the websocket isn't efficient or cost effective. Sending direct via a peer to peer connection is both effecient and cost effective. The only downside to these connections is that events are not guaranteed to be delivered. For low priority events, like cursor position, that's just fine.
2. Pings are nice, but exact cursor position is way better.
3. Seeing rulers is also very nice, especially while dragging tokens
4. It's important for DMs to be able to hide what they are doing from time to time

# What's Next

1. I wouldn't mind streaming token position while dragging.
    - This will require some work to make sure we don't overwrite the real token movement events
3. Anything else that requires a lot of events to be sent really quickly

# Demo
This is two Chrome browsers on the top, and two Firefox browsers on the bottom. They are all on the same computer, but I have also tested two computers on separate networks. I have yet to try many computers with some on poor networks or weak computers.
avtt-cursor-position-demo

avtt-ruler-stream-demo-3

I've also added a "player online" indicator in the tokens and players panel.
Screenshot 2023-02-08 at 8 23 02 PM

Settings options
Screenshot 2023-02-12 at 10 46 32 AM

DMs can quickly and easily disable cursor/ruler streaming by toggling this button
Screenshot 2023-02-12 at 10 49 26 AM


Settings.js
```diff
@@ -222,7 +222,17 @@ function avtt_settings() {
...
+		},
+		{
+			name: 'showCursorToPeers',
+			label: 'Show cursor to other players',
+			type: 'toggle',
+			options: [
+				{ value: true, label: "Show Cursors", description: `When you move your cursor, others will see where your cursor is. Similarly, you will see other player's cursors.` },
+				{ value: false, label: "Don't Show Cursors", description: `You are not sharing your cursor position with other players. You are also not seeing other player's cursors.` }
+			],
+			defaultValue: true
```
CollinHerber, line 235: Are we able to see if the player is a DM? Maybe have this disabled by default for DMs?

MessageBroker.js
```diff
@@ -1013,6 +1013,14 @@ class MessageBroker {
 					 }
 				}
 			}
+
+			if (msg.eventType === "custom/myVTT/peerReady") {
```
CollinHerber, line 1017: We should consider moving to a switch, or returning within the if statements to break out from the logic to reduce execution size

CollinHerber (approved):

Settings.js
vlaminck: `window.DM` will show us if we're the DM. As a DM myself, I definitely want this enabled by default. However, I think you're thinking about this in a way that I hadn't. The way I built it is all or nothing, but allowing DMs to see other cursors without sending theirs is an interesting idea! We could add another setting for DMs to not send their cursor, but still receive the player's cursors. I can see some DMs not wanting to show their cursor to players. That being said, cursors are below fog of war so players won't see anything that a DM is doing in fog. So maybe that's enough. We could always add this in later if enough people ask for it, but I'm also open to including it right away. Thoughts?

MessageBroker.js
vlaminck: I agree, but that's a much larger change to this file than I'm willing to do in this PR. That feels like a separate "cleanup" PR

Settings.js
CollinHerber: I don't see myself using the feature if it's all or nothing. I tend to have a good bit of the map revealed, but am twiddling with hidden tokens frequently that would give my players a potential insight to where things are. 
I think without the "hide dms movements" I'd have to pass on enabling this.

With that in mind, I would like to see it added in this PR but at the same time I could see your use-case would be solved so the value is there.

Up to you!

MessageBroker.js
CollinHerber: Ahh, I just pulled up the entire file and there is TONS of these. Yea perhaps.

CollinHerber (approved):

Azmoria (approved):
From my testing with players this works really well and is a very nice feature to have.

PR #799, opened 2023-02-10
Editable, Moveable, adjusts with scale changes to stay in postion and better quality text

Azmoria:
Changing text to SVG allows it to scale better with zoom giving it better quality.

We've had some canvas performance issues, although text isn't changed often 1 less large canvas may help. Scaling text in canvas doesn't look great.

This starts off by allowing the DM to move the text if they have the text tool selected. It works off the same data so it does pull in old text and convert. If your players are on the old version new text will still draw as canvas drawings. 

It does add a few extra data points (id, backround fill) as we no longer need to fill the background rect seperate and a way to identify them was needed.

This will also allow us to potentially edit the text ~~in the future~~ now. 

https://user-images.githubusercontent.com/65363489/217996328-8b076056-fd70-4033-93b4-83fffa413fe7.mp4

https://user-images.githubusercontent.com/65363489/218035618-c3e94502-6f0d-4c28-bc3a-c7e79a4cf6ee.mp4

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -689,8 +689,10 @@ function reset_canvas() {
 	ctxScale('temp_overlay');
 
 	ctxScale('grid_overlay');
-	ctxScale('text_overlay');
 	ctxScale('draw_overlay');
+
+	$("#text_div").css("width", $("#scene_map").width()*window.CURRENT_SCENE_DATA.scale_factor);
+	$("#text_div").css("height", $("#scene_map").height()*window.CURRENT_SCENE_DATA.scale_factor);
 
 	var canvas = document.getElementById("fog_overlay");
 	var ctx = canvas.getContext("2d");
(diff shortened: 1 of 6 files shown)


CollinHerber (approved):
Amazing!

vlaminck:
This is absolutely huge! This has been a feature request for a really long time, and it will be so nice to have. Here are my observations with this PR:

1. Text that was added before this change doesn't show up for me at all. If I change back to the `main` branch, the text shows up again. I think we need to figure out a way to either migrate old text drawings, or continue to show the old ones if they exist. If we can't migrate them, we're going to end up with a weird hybrid scenario that we'll need to think through.

2. The stroke size is handled a little differently. It crowds inward instead of growing outward. Not a huge deal, but it would be nice to have larger stroke sizes on the outside of the fonts. However, the old way caused weird artifacts on sharp letters like M so this is a huge improvement in that area:
Screenshot 2023-02-10 at 2 20 55 PM
Screenshot 2023-02-10 at 2 34 09 PM

3. I think we should add a `Select` option to the Text submenu. Being able to select and move drawings while also being able  to draw them gets a little confusing. I got into a weird state where I couldn't select existing drawings because it kept giving me the drawing window. So having separate `Select` and `Draw` submenus would make your actions more explicit.

4. Players don't see text one scene load. All text is missing from the Player's side until you draw a new text, move a text, or take some text action.
avtt-no-text-on-load

Azmoria:
Old text should show up - it does for me. I'll double check I didn't break that. I did code it to migrate so maybe just a bug.

vlaminck:
These are the errors I get when I load up your PR
Screenshot 2023-02-10 at 3 23 56 PM

vlaminck (approved):
All 4 of my concerns have been addressed! This is amazing! 🎉

PR #802, opened 2023-02-13
Add pause feature - while paused players  won't be able to move tokens on the map

### LINKED ISSUE #800: [Feature] Lock All Tokens on Scene
When I saw the dynamic lightning video one of the things that came to mind was "Ohh I don't have to hide my tokens, they could just come into view when they are within the light" - However one problem that popped into my head was players going gung-ho and moving all around like crazy, so then came the idea of locking ALL tokens into place. Basically this would prevent all tokens from being moved by players. This would allow the DM to prevent any further changes happening and perhaps given time to catch up or do whatever they need to do for the scene.

Food for thought

Azmoria:
I think this Closes #800

Example:

https://user-images.githubusercontent.com/65363489/218415321-e55b77fe-be0d-4d13-808e-4aae5f4b9015.mp4


Azmoria:
Adjusted the appearance in general and added a background

image

PR #804, opened 2023-02-13
Add join button to character page & inject custom roll buttons

### LINKED ISSUE #297: Remove query params from url

### LINKED ISSUE #300: Add "Join AboveVTT" button to the character sheet.
As a player, it would be convenient to be able to start the AboveVTT session from your character sheet.

This requires https://github.com/cyruzzo/AboveVTT/issues/297 to be completed first

vlaminck:
This does a few things:
1. It pulls some core functionality to a new `CoreFunctions.js` file, and pulls some functions into a new `CharactersPage.js` file
    * These functions are available while AVTT is running
    * These functions are available on the characters page while AVTT is not running
2. This adds a new "Join AboveVTT" button on the characters page when AVTT is not running
3. This adds a new "Exit AboveVTT" button on the characters page when AVTT is running
4. This removes the campaign secret from the url
5. This remove the campaign id from the url on all pages except the encounters page

Screenshot 2023-02-13 at 9 18 07 AM

Screenshot 2023-02-13 at 9 18 52 AM

avtt-dice-on-character-page

closes #297 
closes #300


Azmoria (approved):
So the only issue I see is due to dndbeyond loading by replacing it with react when clicking 'view' from my characters this doesn't load in. 

This was a problem in beyond20 at one point and they solved it by adding the page before to the manifest and they were already using an observer to load stuff.

Clicking the name or opening a sheet from my campaigns still works as expected.

vlaminck:
> So the only issue I see is when clicking 'view' from my characters this doesn't load in. This is due to the way dndbeyond loads the sheets when using the view button from 'my characters' - they replace the page using react and don't actually load it far as I can tell.

Interesting. Good catch! I'll see what I can figure out.

vlaminck:
Thanks for the insight @Azmoria! I got it working when transitioning from the `/characters` page, and as long as I was on that page, I added "Join AboveVTT" buttons to each character card.

PR #839, opened 2023-02-15
Journal Panel update

Azmoria:
Draft Styling Cyelis journal to match scenes.

Work so far:

https://user-images.githubusercontent.com/65363489/219177667-e8bda1c1-3d73-4225-bd70-aa145aaec479.mp4

Feel free to pick this up and add to it. I probably want to style the buttons a bit more - especially the save icon when adding a note. Also I think I might have broke the save button - have to hit enter atm to create a note. I'll look at that later.

Player view is wonky still and I have to get rid of the delete button there
image


Azmoria:
I think this is ready to take a look at. 

before sort button
https://cdn.discordapp.com/attachments/930926587392692335/1076241395867734117/2023-02-17_15-36-27.mp4

now with sort button
https://discord.com/channels/815028457851191326/930926587392692335/1076252615043727552

PR #842, opened 2023-02-16
Fix: peer cursors and rulers were not being cleared all the time

vlaminck:
The problem here was the way that I was using the debounce functions for fading them out. The way that a debounce function works is by calling `setTimeout`, storing the id of the timeout, and clearing/resetting it every time it's called. So any time the timeout gets cleared, it essentially delays the execution of the function. I was calling the same debounce function with different playerId variables which ended up clearing the previous execution which may or may not have been called with a different playerId. Since the timeout id is unique to function, but not unique to the parameters it's called with, fading a playerId would clear out the execution of other playerId fade outs. 

The solution is to store a debounce function for each playerId. An alternative could be to track timestamps from every peer event, and then fade them out within a setInterval function. The reason I went with separate debounce functions, is because setTimeout is significantly more efficient than date parsing. So while this sets a few different setTimeout functions, it's still more efficient than setting Date.now() for every event.

### DIFF
--- a/PeerCommunication.js
+++ b/PeerCommunication.js
@@ -464,8 +464,11 @@ function update_old_player_card(playerId, isConnected, peerColor) {
 
 /** Sets up the peer fade functions. See {@link init_peer_fade_function} */
 function init_peer_fade_functions() {
-  if (!window.PEER_FADE_FUNCTIONS) {
-    window.PEER_FADE_FUNCTIONS = {};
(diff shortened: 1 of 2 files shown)


Azmoria (approved):
Testing locally looks good. 👍 

I haven't run into the issue of them sticking around in my tests with a friend but it might be cause of only 1 other person.

PR #846, opened 2023-02-16
Beta Feature Improvement: stream token position while dragging

vlaminck:
This cleans up the event stream quite a bit. Rather than sending separate events for cursor and ruler, we now send a single event that either includes ruler data or it doesn't. While a player is dragging a token, we also send that token's id in the event which the other peers use to move a clone of the token around the scene. Combining the ruler and cursor events significantly reduces the number of events we stream. I also removed some unused data from the events which reduces the event size. So while this adds token streaming, the overhead is actually lower than what it used to be.

avtt-token-drag-streaming-demo-1

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -723,6 +723,11 @@ function ct_list_tokens() {
 	return tokenIds;
 }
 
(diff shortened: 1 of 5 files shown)


Azmoria (approved):
Very nice, works well.

PR #847, opened 2023-02-17
Bug workaround: store and load the sceneslist from localStorage

vlaminck:
We see reports of the scenes list not loading. We've tracked this to the `sceneslist` event not coming down from the server. The primary cause of that seems to be that the event data is too large. Until we can get some cloud work done to change the scenelist event to a REST request, this will act as a fallback to local storage.

The scenes list will load from localstorage, then if/when the sceneslist event comes, it will overwrite scenes and localstorage. If the event doesn't come down, then they will just continue to operate with localstorage.

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -671,7 +671,13 @@ class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 
 
 		if(window.CLOUD){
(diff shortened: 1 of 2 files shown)


Azmoria (approved):
Good back up 👍

PR #849, opened 2023-02-17
Add change theme button and pull theme color more consistently

vlaminck:
### Change Theme button

The replaces the whisper button on the current players card. Since whispering to yourself doesn't makes sense, I chose to replace it.

Screenshot 2023-02-17 at 9 33 39 AM

### Changing Theme demo
avtt-character-theme-change

### Changing Portrait demo
avtt-character-image-update


Azmoria:
this is useful. Could we also have changing the portrait update their token in real time instead of forcing the DM to back out?

Azmoria (approved):

PR #850, opened 2023-02-18
Add Rotate as group grabber

### LINKED ISSUE #578: Allow rotating a group of tokens about an axis
If you have multiple tokens selected, a new type of "grabber" (similar to the eye grabber that rotates tokens individually) should appear in the corner(s) of the rotation bounding box. Dragging this new grabber should rotate the entire group around a central point rather than rotating each token to face the grabber. This would allow you to set up a horse pulling a wagon with several PCs/NPCs on it, then select them all, and rotate them all to face East when they turn at a fork in the road.

This would likely require disabling grid snapping for all selected tokens. Otherwise, they would all jump out of position as soon as you tried to moved them. I'm not sure if we want to add a new token option to handle this or not.

Azmoria:
Closes #578
 
 Adds another grabber in the top right corner to rotate as group. 
 
 Example
 https://cdn.discordapp.com/attachments/930926587392692335/1076490062327463966/2023-02-18_08-06-29.mp4

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -2513,6 +2513,12 @@ function rotation_towards_cursor(token, mousex, mousey, largerSnapAngle) {
 	return Math.round(degrees / snap) * snap
 }
 
+function rotation_towards_cursor_from_point(tokenCenterX, tokenCenterY, mousex, mousey, largerSnapAngle) {
+	const target = Math.atan2(mousey - tokenCenterY, mousex - tokenCenterX) + Math.PI * 3 / 2; // down = 0
+	const degrees = target * radToDeg;
+	const snap = (largerSnapAngle == true) ? 45 : 5; // if we ever allow hex, use 45 for square and 60 for hex
+	return Math.round(degrees / snap) * snap
+}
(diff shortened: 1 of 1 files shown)


Azmoria:
Both these things should be fixed - it was an error involving a player ID token. I had only tested it with monsters. Missed some quotes that made it not work. Should be fine now.

https://user-images.githubusercontent.com/65363489/220124370-5fd141d9-227d-4c57-b523-684bd69955b0.mp4

Azmoria:
It should be doing that. Hmm it is on my end.

https://user-images.githubusercontent.com/65363489/220126956-989babb9-9e52-43f9-9d6f-fc4819cef57f.mp4

Azmoria:
ok it looks like if a token hasn't been rotated before it doesn't work 

this is how it should look with that token style. I'll look at why it's not working for tokens that haven't been rotated.


https://user-images.githubusercontent.com/65363489/220127924-25073861-f556-4e0b-bdc1-9c1469fde610.mp4

PR #851, opened 2023-02-18
Feature: Parse multiple dice roll commands in character sheet notes

vlaminck:
This updates the regex parsing to globally match anything up to another `/`, and then iterates over all matches.  
For example:  
`Some Note: /hit 2d20kh1 Advantage /dmg 2d10 big damage` would find these matches  
`['/hit 2d20kh1 Advantage ', '/dmg 2d10 big damage']` as seen here  
Screenshot 2023-02-18 at 8 35 01 AM

This conveniently allows us to have other notes before dice roll commands, and if you add a trailing `/` you can have them after the dice commands, too.

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -66,17 +66,28 @@ function inject_dice_roll(element) {
     console.debug("inject_dice_roll already has a button")
     return;
   }
-  const text = element.text();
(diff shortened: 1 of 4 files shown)


Azmoria (approved):
works as expected. I like being able to include text too dropping labels between/before/after roll buttons etc is nice

PR #876, opened 2023-02-23
Beta - Allow player tokens will see other players vision if reveal light is enabled. They will still be able to select a token to get that specific tokens vision

Azmoria:
This is pretty essential in my games. I have players who play more than one PC in some games with a smaller number of players. Currently they are unable to see other player token's vision. This allows them to check their vision if reveal light to players is on. It acts a little different from other tokens in this sense but I think it's worth the trade off for the capability. Especially since it's not the only way pc tokens behave differently. Since there is no 'Player Accessible Stats' equivalent for player tokens.

https://user-images.githubusercontent.com/65363489/220834791-c3f4f58e-e32b-46f6-8a49-0e52b6fd626f.mp4

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -2854,7 +2854,7 @@ let particle = new Particle(new Vector(200, 200), 1);
   		
 	  for(j = 0; j < selectedTokens.length; j++){
 	  	let tokenId = $(selectedTokens[j]).attr('data-id');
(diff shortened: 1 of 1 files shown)


Azmoria:
yeh that makes sense might fill the observer option people want for stream etc too. I stuck this on reveal to keep options light especially on the first pass. But an observer/all player light option wouldnt be bad to build.

Azmoria:
I'm leaving this open until we decide if we want this a setting or default it one way or the other. 

#878 has the same changes but with a window.playerObserverSetting that can be toggled true/false to try both view options.

PR #889, opened 2023-02-24
Tutorial Map and Import full scenes with all data

Azmoria:
Starting some tutorial maps and this adds a function to import a scene with all it's data. This could be used to import maps with walls or other things instead of just grid data.


Azmoria:
a minor adjustment to be made with non-full screen but I can get it before next beta hopefully if you want to merge this

Azmoria:
Made some css adjustments for non-full screen - dndbeyond was adjusting it already based on window size but our container is smaller then the page it's originally for so I mimicked their changes but on larger window sizes. Tried to choose adjustments sizes based on when things start to get cut off or look wonky.

Fixed the overlap on the titlebar. 
 
Made the custom url section smaller and towards the right.
 
 https://i.imgur.com/EZZJR2G.mp4

Issue #915, opened 2023-03-04
Token option/style for always displaying names


Azmoria:
There's been quite a few requests for being able to always display names on tokens - for map pins, labels etc that they can have notes on. 

This could be a good candidate for a token style as I'm not sure we want a full option for this or not. It really depends if people want to be able to display creature names all the time as well or it's primarily for labels.

PR #917, opened 2023-03-04
Map Pin token style

Azmoria:
An option for #915 

map pin token style


Azmoria:
> Are we also going to add that map pin token? Adding this to the top of `built-in-tokens.js` array
> 
> ```
>   {
>     "name": "Map Pin",
>     "folderPath": "/Overlays",
>     "image": "url-from-stumpy.png",
>     "disableborder": true,
>     "square": true
>   },
> ```

yeh we can do that I got it from one of Nates exports when testing something for him. Once we get the url from stumpy I'll throw it in.

Issue #754, opened 2022-12-09
Add basic support for Universal VTT files


vlaminck:
These are just JSON files with the map as a base64 encoded string. Base64 encoded maps have been a problem in the past, but I think as long as we only accept the url where these files are hosted, it shouldn't be a problem. The problem with base64 encoded images is that they are really big strings that often exceed the max size of the websocket messages. Passing a url to a universal vtt file doesn't have that problem because it's just a url like any other image that we support.

### File extensions
These are the file extensions that I know of, but it might not even matter. If we know that it's a universal vtt file, we can just treat it as a `.json` file.
* `.dd2vtt`
* `.df2vtt`
* `.uvtt`

### JSON structure
There's more to the file than just this, but these are the only parts we care about
```
{
  "resolution": {
    "map_origin": {
      "x": 0,
      "y": 0
    },
    "map_size": {
      "x": 12,
      "y": 16
    },
    "pixels_per_grid": 120
  },
  "image": "<base64-encoded-image>"
}
```

### Creating a scene
During scene creation, we might be able to look at the file extension, but we might need a new option for "Universal VTT" similar to our "Is Video" option. We could then parse the JSON, and convert the `resolution` to our system, and save it the same way that we save all other maps.

### Rendering the image
We should pass the url to the file the same way that we pass urls to images. When we parse an image, we just have to check if it's a universal vtt file, then parse the JSON and render the base64 encoded image.

Azmoria:
This has been working for a bit and haven't had any real big issues. Closing this as I think it's complete for now.

PR #918, opened 2023-03-04
Import uvtt files

Azmoria:
For #754

Some ui work to do. And if you want to take a stab at remote hosted files - I'm tapped out on that atm. 

Currently this only applies to the current scene. But I think if we use the full import in my other PR we can do tokens, walls, etc there for other scenes.

https://i.imgur.com/gxxR5bi.mp4


Azmoria:
I think this is ready for a basic review and maybe a beta - I can pretty it up and add the local import for current scenes when I've got more time.

I got a basic UI for importing UVTT files in now. 

Supported sites: Google Drive and Dropbox.

Example:
https://i.imgur.com/JDola6v.mp4

Azmoria:
> I think this PR is ready to be merged and put into a beta. 👍 If you have time, I'd be curious to know if you see anything obvious about map 3, but I'm also fine with optimizing later.

I won't have time to test it today but looking at it I'm guessing `"pixels_per_grid": 200` might be the problem. It may be making the map too large to handle (canvas performance issues as the canvas size increases). Is there a way to export with a smaller grid and scale up instead? I don't have the program so just looking for options before we get flooded with questions lol

PR #923, opened 2023-03-07
Even better error messages

vlaminck:
# What

This Fetches Github Issues that are open and have been marked with the `bug` label. It then uses Fuse.js to do a fuzzy match on the title and body of the issue to see if a similar error has been reported. If it finds any, it will inject them into the error message as a list of links. Any issues that have been marked with the `workaround` label will be highlighted to hopefully draw the user to look at it. 

This also adds a button for users to open a github issue in case we weren't able to find any matching issues. 

Finally, this adds a Github Action that looks at every new/edited Issue, and adds a `potential-duplicate` label so we can easily find and close duplicates. If people start spamming this feature, I'll update it to automatically delete duplicates.

# Demo

Screenshot 2023-03-06 at 8 40 16 PM

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -79,7 +79,7 @@ function showError(error, ...extraInfo) {
     const container = $(`
       <div id="above-vtt-error-message">
         <h2>An unexpected error occurred!</h2>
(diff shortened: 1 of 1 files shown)


Azmoria (approved):
awesome this will be very helpful

PR #935, opened 2023-03-09
fix older characters are missing from window.pcs

vlaminck:
ostrichwrangler on Discord helped me track this down. They had a character that wasn't showing up in the tokens sidebar. The root cause of it was the DDB API endpoint that I was using to fetch window.pcs didn't include the character. We were able to find a different (probably newer) endpoint that did. Changing to this endpoint fixes the problem, but it also gives us way more information. I updated DiceRoller to use the abilities in window.pcs if it exists, and I think we can clean up a lot of `DDBCharacterData` now, too. I'm going to start doing that in a different PR, though, because that will be a larger change.

Side Note: If you edit a character, the old DDB API endpoint will start sending the updated character data. I've seen this happen in other places as well. That hopefully won't be a problem for us, but if we have odd issues with characters not updating or showing up in some places, I think a workaround could be to change the character portrait, save, then change the portrait back, and save again.

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -72,7 +72,7 @@ function inject_dice_roll(element) {
   let updatedInnerHtml = element.text();
   for (const command of slashCommands) {
     try {
-      const diceRoll = DiceRoll.fromSlashCommand(command[0], window.PLAYER_NAME, window.PLAYER_IMG);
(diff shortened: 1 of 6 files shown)


Azmoria:
Wonder if this is why we never found artificer stuff.

Azmoria (approved):

PR #1008, opened 2023-03-19
Fix - Only reload the scene once when updating the scene from the edi dialog.

Azmoria:
In order to avoid getting two fetchscene messages back from the server I check if we're updating the current scene or not when saving the edit dialog.

I'm not 100% sure why this works as I haven't looked at the server side but we get the fetchscene message back twice on the current scene if we don't check this.


Azmoria:
ok I moved the check to be around the click. It won't click the dm button if the scene is the current scene. Otherwise click it to move us there.

PR #1018, opened 2023-03-20
Start collecting pc data from sheet

Azmoria:
Putting up a draft of this. It's got the basics working I think: hp, armor, conditions, deathsaves.

basics


Azmoria:
starting a check list of stuff we need to collect. I'll add more too it as I remember/have it in front of me.

PR #1019, opened 2023-03-20
Fix - Snap to grid miscalculation

vlaminck:
This feels like it shouldn't be this easy, and I'm worried that this will break something else considering this function hasn't changed in months. Everything I've tested seems to work, though.

I also fixed a bunch of bad `for` loops. I often see `i is not defined` because some of these execute before a different variable had the chance to set it globally. None of these should be global variables.

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -192,7 +192,7 @@ function init_combat_tracker(){
 	
 	clear=$("<button>CLEAR</button>");
 	clear.click(function(){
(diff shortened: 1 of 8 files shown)


Azmoria (approved):
My bad on the for loops - learned something new got so used to typing them that way didn't even think about declaring them as let. 

I'm going to merge this and see how it affect my snapping thing too - maybe we only need 1 of them. But I think the calculation for wall collision should still be changed.

PR #1021, opened 2023-03-21
clear drawings on scene change

vlaminck:
not sure if this is the best way to do this, but it seems to work


Azmoria (approved):
If this fixes it I think this is fine. 

These happen later in this same function and I don't see this issue so it might be a race condition or I was just doing things in the wrong order.

PR #1036, opened 2023-03-22
Feat - add some Midjourney tokens

Azmoria:
Currently up to elves in the token list in discord

### DIFF
--- a/built-in-tokens.js
+++ b/built-in-tokens.js
@@ -551,7 +551,513 @@ const builtInTokens = [
       "https://i.imgur.com/zrN3Vnt.png",
       "https://i.imgur.com/eKfE1AV.png"
     ]
+  },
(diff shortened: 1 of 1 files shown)


Azmoria:
Yeah I took out the 2nd slash when I realized I couldn't go another folder deep - left midjourney so it would be easy to replace all when we decided if we wanted another layer or not. Working on it now again. I prefer to remove it if we aren't adding another folder layer.

Azmoria:
Swapped to token alternates - it allows for both options of random or specific. So I feel it's more versatile and does clean up the menus a bit after testing both.

Still open to another folder idea but I think they should somehow be seperate to the current tokens of the same type since the art style is mostly black and white and not everyone will want that in their random tokens likely.

PR #1058, opened 2023-03-25
Fix - Tokens being seen outside expected vision

Azmoria:
So this is due to shared light going through walls and the tokens are in the bounds of the shared vision div but only in line of sight of the main token. This ends up revealing tokens in darkness.

I've solved this by limiting the range of vision to a tokens light. The side effect of this being dragging a token will not reveal any additional parts of the map as they do now. You'll also be locked to moving your token no further then where you could see since the token movement boundarys are determined by the same layer. 

I'll think on if there's ways to keep that functionality without tanking performance but for now this at least fixes the issue of tokens no appearing.


Azmoria:
yeh next beta is probably better

PR #1075, opened 2023-03-29
Performance - Keep line of sight polygons stored so we don't have to recalculate them if a token/walls haven't changed. (the majority of time only 1 token is moved)

Azmoria:
This makes it so that line of sight polygons aren't recalculated for every token when only one has moved.

particle.look takes quite a bit of time/processing so this should help with performance some.


Josh-Archer (approved):
Nice, don't think our merge conflicts will be bad. Also, I know my approval means nothing but feels like I should approve 😆

PR #1076, opened 2023-03-29
Improve Fog DOM read/write performance

Josh-Archer:
Improving fog by removing unused selects, passing around the context and canvas some to reduce grabbing it, and move some logic/selects to only run when needed. Use `putImageData` instead, it is much more performant.

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -247,6 +247,7 @@ function read_inspiration(container = $(document)) {
   return false;
 }
 
+// Good canidate for service worker
 function init_characters_pages(container = $(document)) {
   // this is injected on Main.js when avtt is running. Make sure we set it when avtt is not running
   if (typeof window.EXTENSION_PATH !== "string" || window.EXTENSION_PATH.length <= 1) {
--- a/Fog.js
+++ b/Fog.js
@@ -378,9 +378,9 @@ function is_token_under_fog(tokenid){
 	if(window.DM)
 		return false;
 	let canvas = document.getElementById("fog_overlay");
-	let ctx = canvas.getContext("2d");
+	let ctx = canvas.getContext("2d", {willReadFrequently: true});
 	let canvas2 = document.getElementById("raycastingCanvas");
-	let ctx2 = canvas2.getContext("2d");
+	let ctx2 = canvas2.getContext("2d", {willReadFrequently: true});
 
 
 	let left = (parseInt(window.TOKEN_OBJECTS[tokenid].options.left.replace('px', '')) + (window.TOKEN_OBJECTS[tokenid].options.size / 2)) / window.CURRENT_SCENE_DATA.scale_factor;
@@ -436,25 +436,23 @@ function check_single_token_visibility(id){
 	console.log("check_single_token_visibility");
 	if (window.DM || $("#fog_overlay").is(":hidden"))
 		return;
-	var canvas = document.getElementById("fog_overlay");
-	var ctx = canvas.getContext("2d");
-			let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
-			let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
-			let selector = "div.token[data-id='" + id + "']";
-			let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
-			let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
 
-			if (!window.TOKEN_OBJECTS[id].options.revealInFog && (is_token_under_fog(id) || (playerTokenHasVision && window.CURRENT_SCENE_DATA.darkness_filter > 0 && !is_token_under_light_aura(id)))) {
-				$(selector).hide();
-				$(auraSelector).hide();
-			}
-			else if (!window.TOKEN_OBJECTS[id].options.hidden) {
-				$(selector).css('opacity', 1);
-				$(selector).show();
-				if(!window.TOKEN_OBJECTS[id].options.hideaura && id != playerTokenId)
-					$(auraSelector).show();
-				//console.log('SHOW '+id);
-			}
+	let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
+	let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
+	let selector = "div.token[data-id='" + id + "']";
+	let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
+	let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
+
+	if (!window.TOKEN_OBJECTS[id].options.revealInFog && ((playerTokenHasVision && window.CURRENT_SCENE_DATA.darkness_filter > 0 && !is_token_under_light_aura(id)) || is_token_under_fog(id))) {
+		$(selector, auraSelector).hide();
+	}
+	else if (!window.TOKEN_OBJECTS[id].options.hidden) {
+		$(selector).css('opacity', 1);
+		$(selector).show();
+		if(!window.TOKEN_OBJECTS[id].options.hideaura && id != playerTokenId)
+			$(auraSelector).show();
+		//console.log('SHOW '+id);
+	}
 }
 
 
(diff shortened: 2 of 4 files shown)


Fog.js
```diff
@@ -981,9 +978,8 @@ function redraw_light_walls(clear=true){
 		let adjustedScale = scale/window.CURRENT_SCENE_DATA.scale_factor;
 
 		if (shape == "line" && ($('#wall_button').hasClass('button-enabled') || ($('#fog_button').hasClass('button-enabled') && $('[data-shape="paint-bucket"]').hasClass('button-enabled')))) {
-			canvas = document.getElementById("temp_overlay");
-			ctx = canvas.getContext("2d");
-			drawLine(ctx,x, y, width, height, color, lineWidth, scale);		
+			let raycastingContext = document.getElementById("raycastingCanvas").getContext("2d");
```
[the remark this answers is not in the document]
Azmoria: I think you have the right idea removing raycasting from the top - raycasting isn't used in this function anymore I think that was just left over from something else. I think just removing this canvas/context grab as well and just using temp_overlay from above is the way to go. We draw the walls there when using any wall tools or the bucket fill tool.

Raycasting is for line of sight and is drawn in redraw_light.

You may have already checked this but this runs in a couple other files. Not sure if those need updating though with how you've set this up.

Fog.js
```diff
@@ -2358,7 +2354,7 @@ function bucketFill(ctx, mouseX, mouseY, fogStyle = 'rgba(0,0,0,0)', fogType=0){
 	let distance = 10000;
   	particleUpdate(mouseX, mouseY); // moves particle
 	particleLook(ctx, window.walls, distance, fog, fogStyle, fogType); 
-	redraw_light_walls();
+	redraw_light_walls(ctx);
```
Azmoria, line 2357: This errors since it's reading ctx as the canvas variable in redraw_light_walls.

Maybe we just pass on ctx to redraw_light_walls? If ctx is undefined then grab the canvas and context? I think anywhere we can pass on the canvas we also have the context of temp_overlay.

Fog.js
```diff
@@ -464,26 +462,27 @@ function check_single_token_visibility(id){
...
-		setTimeout(do_check_token_visibility,1000);
+		window.NEXT_CHECK_TOKEN_VISIBILITY = Date.now() + 1000;
+		setTimeout(do_check_token_visibility, 1000);
 		return;
 	}
 }
 
 function do_check_token_visibility() {
 	console.log("do_check_token_visibility");
-	if (window.DM || $("#fog_overlay").is(":hidden"))
```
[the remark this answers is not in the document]
Josh-Archer: Yep, in the parent function `check_token_visibility` it does the check as well. This function wasn't called anywhere else. So it was redundant

Fog.js
```diff
@@ -3176,17 +3156,12 @@ function redraw_light(){
...
 
-  		
   		drawPolygon(offscreenContext, lightPolygon, 'rgba(255, 255, 255, 1)', true); //draw to offscreen canvas so we don't have to render every draw and use this for a mask
-		
-  		
 	}    	
   }
   context.drawImage(offscreenCanvasMask, 0, 0); // draw to visible canvas only once so we render this once
-  $('#VTT').css('--vision-mask', `url('${offscreenCanvasMask.toDataURL('image/png', 0)}')`) // make image ask of offscreen canvas
+  $('#VTT').css('--vision-mask', `url('${offscreenContext.putImageData(offscreenContext.getImageData(0, 0, canvasWidth, canvasHeight), 0, 0)}')`) // make image ask of offscreen canvas
```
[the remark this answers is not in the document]
Josh-Archer: Yeah, I saw quite a bit improvement from this. Removing it should be even better! Nice!

Fog.js
Josh-Archer: Ill play a little with this and see which looks better

Fog.js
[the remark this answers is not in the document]
Josh-Archer: I started to try and find them but then quickly was getting lost. I actually am working on a second more in depth performance pr. Im testing some WebGL stuff to see how much quicker that makes it too. If we remove the canvas/context grab here, we could see another 20%ish bump from my tests with high level of complexity/density map

CharactersPage.js
```diff
@@ -247,6 +247,7 @@ function read_inspiration(container = $(document)) {
   return false;
 }
 
+// Good canidate for service worker
```
Josh-Archer, line 250: Called this out so i don't forget

Token.js
```diff
@@ -40,18 +40,13 @@ const availableToAoe = [
...
-		}
-		else{
+		if(window.walls?.length < 5){
 			redraw_light_walls();	
 		}
 		redraw_light();
 		if(!window.DM)
 			check_token_visibility();
-}, 100);
+}, 250);
```
Josh-Archer, line 49: increased debounce to make it feel smoother while person is moving  quickly. Thoughts? I thought about changing this depending on image size? shorten for smaller images since processing is quicker?

Fog.js
```diff
@@ -500,9 +500,10 @@ function do_check_token_visibility() {
 		let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
 		let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
 
-		pixeldata2 = pixeldata2.filter(function(color, index) {return (index + 1) % 4 != 0});
+		//Combining some and filter cut down about 140ms for average sized picture
+		let someFilter = function(color, index) {if ((index + 1) % 4 != 0 && color == 255) return true;};
```
Azmoria, line 504: we had the same idea here lol. I think we might be able to improve this more using just the edges of the token but I'm not sure as I think we'd need 4 get image datas for that a single line of pixels for each side of the square.

I had done index % 4 == 0 instead to just check the red of rgba don't think it saves much if any time though since it's still looping through them. This checks rgb and leaves out a which is fine.

I'll close my PR since you've got it here. 👍

Token.js
Azmoria: If it feels smoother it's good with me. I just had it in there for when moving multiple tokens as once it didn't get called for all of them dropping. And I had to worry less about doubling up on it.

Fog.js
Josh-Archer: Nice! Great minds think alike lol. Ill make that change. Also, I almost have a web worker implementation working for the fog redraw, was thinking see web worker performance then eventually move a lot over to service worker. Little bit easier getting web worker up and running. Ill open a pr for that tomorrow probably, gonna make these changes tonight and push up

Fog.js
```diff
@@ -418,15 +417,12 @@ function is_token_under_light_aura(tokenid){
...
+
 		if(horizontalMiddle > bounds.left && horizontalMiddle < bounds.right && verticalMiddle > bounds.top && verticalMiddle < bounds.bottom){
 				
 
 			let pixeldata = window.lightAuraClipPolygon[auraId].canvas.getContext('2d').getImageData(parseInt(window.TOKEN_OBJECTS[tokenid].options.left.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, parseInt(window.TOKEN_OBJECTS[tokenid].options.top.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[tokenid].sizeWidth()/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[tokenid].sizeHeight()/ window.CURRENT_SCENE_DATA.scale_factor).data;
-
-			pixeldata = pixeldata.filter(function(color, index) {return (index + 1) % 4 != 0});
-			if(pixeldata.some(color => color == 255))
-				return true;	
+			return pixeldata.some(function(color, index) {return (index) % 4 == 0 && color == 255});
```
Josh-Archer, line 425: Changed this a little. Realized we could shorten and just return the result of some

Azmoria:
I'll do a test run later tonight and put an approval up after if all goes well which it looks like it will. Thanks for doing this, excited for the web worker stuff too.

Fog.js
```diff
@@ -436,25 +432,23 @@ function check_single_token_visibility(id){
...
-			}
+
+	let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
+	let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
+	let selector = "div.token[data-id='" + id + "']";
+	let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
+	let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
+
+	if (!window.TOKEN_OBJECTS[id].options.revealInFog && ((playerTokenHasVision && window.CURRENT_SCENE_DATA.darkness_filter > 0 && !is_token_under_light_aura(id)) || is_token_under_fog(id))) {
+		$(selector, auraSelector).hide();
```
Azmoria, line 443: Only issue I've found during testing

This looks like it needs to be `$(selector + "," + auraSelector).hide();` for multiple string selectors
quick issue

Everything else seems to be performing well and I think it can be merged after this change 👍

Josh-Archer: Ahhh yep, nice catch. Fixing now.

Fog.js
```diff
@@ -995,9 +989,7 @@ function redraw_light_walls(clear=true){
 		let adjustedScale = scale/window.CURRENT_SCENE_DATA.scale_factor;
 
 		if (shape == "line" && ($('#wall_button').hasClass('button-enabled') || ($('#fog_button').hasClass('button-enabled') && $('[data-shape="paint-bucket"]').hasClass('button-enabled')))) {
-			canvas = document.getElementById("temp_overlay");
-			ctx = canvas.getContext("2d");
-			drawLine(ctx,x, y, width, height, color, lineWidth, scale);		
+			drawLine(canvas, x, y, width, height, color, lineWidth, scale);
```
Azmoria, line 992: Sorry missed this yesterday just noticed while testing the memory stuff - canvas should be ctx in this draw line.

Josh-Archer: 🤦‍♂️ yep, nice catch

Azmoria (approved):

Azmoria:
I think this is good to go now - very nice job.

Josh-Archer:
I think you have to merge it because I don't have write permissions, thank you for the review!

Azmoria:
Yeah I'll merge after @vlaminck (mikedave in discord) takes a look. We try to do 2 approvals where we can even if we only need 1 to merge. More eyes the better is all =) - I don't think theres anything wrong.

Josh-Archer:
> Yeah I'll merge after @vlaminck (mikedave in discord) takes a look. We try to do 2 approvals where we can even if we only need 1 to merge. More eyes the better is all =) - I don't think theres anything wrong.
> 
> If you have time to review PR's we wouldn't mind extra eyes either. We've only been doing 1 approval on our PRs since most devs are too busy irl to review and we can't review our own lol.

Awesome! Love it! Definitely will review whenever I can. Still very new to the repo but I'll do my best 😀

Issue #914, opened 2023-03-04
Close text controller on tool swap


Azmoria:
It was pointed out to me the text controller has to be manually closed. I think we should close it on tool swap to non-text.

Just didn't want to forget about it.

Josh-Archer:
Ill take this one!

PR #1084, opened 2023-04-02
Close text controller on tool swap, esc, clear all

Josh-Archer:
https://github.com/cyruzzo/AboveVTT/issues/914

### DIFF
--- a/Text.js
+++ b/Text.js
@@ -402,6 +402,15 @@ function create_moveable_text_box(x,y,width, height, text = undefined) {
         minHeight: 55
     });
 
+    document.querySelectorAll('.menu-button')
+    .forEach(occurence => {
+        let id = occurence.id;
+        if (id && (!id.includes("text") || id === "text_button"))
+        occurence.addEventListener('click', function() {
+            document.getElementById("text_controller_inside")?.remove();
+        }, { once: true});
+      });
+
     $(textInputInside).mousedown(function () {
         frame_z_index_when_click($(this));
     });
@@ -523,7 +532,10 @@ function handle_draw_text_submit(event) {
  * @param {Event} e 
  */
 function handle_key_press(e) {
-    if (e.key == "Escape") $(this).parent().remove();
+    if (e.key == "Escape") {
+        $(this).parent().remove();
+        document.getElementById("text_controller_inside")?.remove();
+    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Text.js
```diff
@@ -402,6 +402,15 @@ function create_moveable_text_box(x,y,width, height, text = undefined) {
         minHeight: 55
     });
 
+    document.querySelectorAll('.menu-button')
+    .forEach(occurence => {
+        let id = occurence.id;
+        if (id && (!id.includes("text") || id === "text_button"))
+        occurence.addEventListener('click', function() {
+            document.getElementById("text_controller_inside")?.remove();
+        }, { once: true});
```
Josh-Archer, line 411: Allows this to only run once. Clean up everywhere would be way more code, because of how it would remove all click event handlers on the button. I think this is the simplest and least bug prone way to do it. If you guys know another way, lmk!

Josh-Archer: Note: the event listener does get picked up by gc after the one run

Azmoria: something like the below in `init_buttons()` in fog.js would be an alternative. Maybe add something about if the text box exists. 

Binds it to the whole row of buttons but only triggers if the class is a header button. I think init_buttons only runs the once and we don't need to rebind an event listener that way. 

I'm sure they both work fine though. Whichever you feel is better tbh you seem to have a better grasp of that stuff than I do. 

I'm not really sure what running it once means? Is this so that it gets shut down since we're making it everytime we're making the box here?

I don't think we need to remove it on text clear necessarily as I think we're still on the text dropdown menu - but it would reopen when drawing a new text box or editing it anyway so not a big deal to me. Same with the esc bind you've added when writing text. I think as long as we're on the text main button it's ok to have open still but I don't have a huge preference to it being open or closed.

```
	buttons.on('click', '.drawbutton.ddbc-tab-options__header-heading', function(event){
		if(event.target.id != 'text_button'){
			$("#text_controller_inside").remove()
		}
	})
```

Josh-Archer: Thank you for the feedback! Definitely feedback I am looking for lol. 

The only problems I see with that approach are the following:
1. Could unnecessarily register the event handler (i.e. - its there but someone never uses it)
2. Kind of related to 1, but it fires every time the button is clicked and we decide no action is needed because the controller is already gone. 

I can't really add the once flag on the handler because they would need to be persistent if done in `init_buttons()`. I would vote we leave as is. I think it is more performant (runs less times/uses less memory since the handlers only will run 1 vs X times). I think `init_buttons` is the easier place to read/find though. It is your guys code base though, so let me know what you want and Ill change it if that is what you guys decide!

Azmoria (approved):

PR #1090, opened 2023-04-03
Feat - Add filter for `Show only monsters I have access to`

Azmoria:
I made this awhile back - I was unsure of the additional requests to ddb and where to put the setting. I didn't want to mess with the monster filter since we grab that from ddb but not 100%. Threw a setting in experimental for now.

Putting it up as a PR for thoughts on it.

The below is with a tester account that doesn't own content and the one homebrew monster at the top there. It loads in monsters 10 at a time as we usually do but if we don't have 10 owned monsters in the request moves to the next 'page' of requests until we get 10.

only show owned


Azmoria:
yeah, orignally i thought it would be best to include it eith those filter. It doesnt affect me cause I own everyhting but if you own a decent number of books I would find it a pain to try and filter all the books I own/have access to manually.

Its also a highly requested feature on dndbeyond in general on their forums/in their discord. Lots of users seem to have issues even knowing what they have access to or own.

I think in part since some things are shared with them its hard to select the sources they have access to when they wernt even the ones who purchased them. That and i wouldnt be surprised to learn people just forget what they own haha. I this this is techically content they have access too not just owned.

Either way though Im happy to leave it out - on or off its the exact same on my end 😆

Azmoria:
So yeh, I double checked and this shows any monster you have access too - including those shared from other campaigns. I think it might be rather hard to figure out which sources are shared with you for some people so this is probably more useful than I originally thought.

I also moved it to filters instead of experimental settings. 

only show owned


Let me know what you think.

Josh-Archer (approved):

PR #1093, opened 2023-04-05
change scene import item width

vlaminck:
This has been bothering me. I changed the width to 33% so they all layout in a row.

Screenshot 2023-04-04 at 7 17 17 PM


Azmoria:
We either should adjust the sizing on the actual map imports as well or just apply this to the main import window. It causes some odd looking map imports when not full screen.

image

Azmoria (approved):

PR #1094, opened 2023-04-05
persist scene folders in the cloud

vlaminck:
This needs to be cleaned up still, but it seems to be working.

Rather than building the folder list dynamically based on `scene.folderPath`, this changes scene folders to be scene objects without any details, and sends them to the cloud alongside scenes.

The dynamic building of the folder structure had a few bugs in it, and was pretty messy. It also got really messed up if there was a naming conflict with either other folders or scene names. This does some clean up, and changes it so that each folder can have a unique id so naming conflicts don't matter any more. I still need to take one code cleanup pass before we merge this, but I think it's ready for testing.

### DIFF
--- a/ScenesPanel.js
+++ b/ScenesPanel.js
@@ -1462,7 +1462,7 @@ function redraw_scene_list(searchTerm) {
 	// this is similar to the structure of a SidebarListItem.Folder row.
 	// since we don't have root folders like the tokensPanel does, we want to treat the entire list like a subfolder
 	// this will allow us to reuse all the folder traversing functions that the tokensPanel uses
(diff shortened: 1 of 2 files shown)


Azmoria:
Gave this a quick test. Ran into a couple issues - if you need me to check for other info that might help let me know.

Loading into a campaign all the scenes loaded in outside folders first, then the folders loaded in empty. A refresh/rejoin didn't put them back in the folders.

Creating a new folder gives me this error:
image

And I can't seem to move the scenes back into folders.
move folders back

Azmoria:
When moving scenes between folders or into folders it works great. But moving them out of folders to the main list I get this error. All the other issues look like they've been resolved though 👍 

image

Moving out of folders

Issue #1105, opened 2023-04-06
Fonts Blocked by Chrome sometimes


Josh-Archer:
https://discord.com/channels/815028457851191326/852495991227809802/1093551161740627978

Sometimes fonts are blocked by chrome https://code.google.com/p/chromium/issues/detail?id=446473. 

Workaround is to put them in `web_accessible_resources`

PR #1106, opened 2023-04-06
Update manifest.json to fix font blocking 

Josh-Archer:
Fix font block by chrome

https://github.com/cyruzzo/AboveVTT/issues/1105

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -39,6 +39,7 @@
 		"assets/*",
 		"audio/*",
 		"images/*",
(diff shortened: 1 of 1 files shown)


Azmoria:
This error has always been there and most people when we ask them to check send us these first or these are the only ones that show for them even with no impact on them. It'll be nice to have them gone for trouble shooting.

Azmoria (approved):

Issue #1104, opened 2023-04-06
Temp HP on characters gets added every time a token moves


vlaminck:
![gif](https://media.discordapp.net/attachments/815028804103307354/1093542213541187625/AboveVTT-bug-HPIncreaseOnMove.gif)

Josh-Archer:
Is this only for DM, player or both movements?

Josh-Archer:
> Is this only for DM, player or both movements?

Nvm, I found the issue. Ill take this

PR #1107, opened 2023-04-06
Fix temp hp bug

Josh-Archer:
https://github.com/cyruzzo/AboveVTT/issues/1104

I don't think we need to run `update_and_sync` on movement from what I can tell. The issue is we are appending the temp hp when we get the hp to set for the currentHp in the hp object. This was the best way from what I can tell vs hacking it/adding another hp getter/setter in place of the one we use on `baseHp`. If I missed something, and we do want `update_and_sync` to run let me know and ill go the longer approach by modifying how its getting/setting hp.

Also, went ahead and modified the `var` to `let` for GC improvements and found one small for loop that was using the same objects twice in scope. These pieces aren't part of this bug, so if you want I can break it out. Just figured should tackle it while I am in there.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -459,8 +459,8 @@ class Token {
 		}
 		let imageScale = (this.options.imageSize != undefined) ? this.options.imageSize : 1;
 
-		var selector = "div[data-id='" + this.options.id + "']";
-		var tokenElement = $("#tokens").find(selector);
+		let selector = "div[data-id='" + this.options.id + "']";
+		let tokenElement = $("#tokens").find(selector);
 		tokenElement.css("--token-rotation", newRotation + "deg");
 		tokenElement.css("--token-scale", imageScale);
 		tokenElement.find(".token-image").css("transform", `scale(var(--token-scale)) rotate(var(--token-rotation))`);
@@ -525,7 +525,6 @@ class Token {
 		this.options.top = top + 'px';
 		this.options.left = left + 'px';
 		this.place(100);
-		this.update_and_sync();
 	}
 
 	snap_to_closest_square() {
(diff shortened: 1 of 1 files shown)


Azmoria:
If we take out the update/sync it won't sync to the dm/server when a player moves the token from what I can see in testing.

This has to do with how we changed collecting data.
Around line 818/819 if you change `this.hp =` to
```
	this.hp = old.find(".hp").val() - this.options.hitPointInfo.temp;
```
it should fix it in all the places we call update_and_sync or update_from_page. As it's likely happening elsewhere as we use these functions in other places. Just something we missed from our new data.

Josh-Archer:
Ah I didn't test that sync piece. I missed that happening. Okay I'll change this when I get some time tonight. Thanks for feedback

Josh-Archer:
Hope you don't mind, I just removed that line. Felt like we were fixing the the issue after it happened vs fixing the issue from happening. I think this will solve both and save processing. I know very minimal but it all adds up :). If you are against this approach lmk. I tested what I know it impacted but if you see something I missed lmk

Azmoria:
As long as we don't need to grab the hp from the token before sending the update in any of the places we call this this should be fine. I haven't checked all the places we run it but it seems to work for the things I have tested.

 I tried removing the max hp one too as they should both be removable if the one is but it causes issues if removed. I'm guessing we're sending the current hp elsewhere for monsters too and were doubling up on it here.

Azmoria (approved):

Azmoria:
One request can you rename this PR to something about fixing the temp hp bug so when we pull the name for beta/release notes general users will understand what it is.

PR #1108, opened 2023-04-06
Fix - clicks shoud go through locked tokens for players

Azmoria:
Reported in discord locked tokens currently block click events for players. This should resolve that.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -2439,10 +2439,12 @@ function toggle_player_selectable(tokenInstance, token){
 	if (tokenInstance.options.locked && !window.DM){
 		tokenImage?.css("cursor","default");
 		tokenImage?.css("pointer-events","none");
+		token.css("pointer-events", "none");
 	}
 	else{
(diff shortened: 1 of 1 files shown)


Token.js
```diff
@@ -2439,10 +2439,12 @@ function toggle_player_selectable(tokenInstance, token){
 	if (tokenInstance.options.locked && !window.DM){
 		tokenImage?.css("cursor","default");
 		tokenImage?.css("pointer-events","none");
+		token.css("pointer-events", "none");
```
Josh-Archer, line 2442: Would it be worth turning off other events too like our keybinds on it? Would save some listeners just sitting there

Josh-Archer (approved):

Token.js
Azmoria: we could - then add them back if unlocked. Just for players though DM's still need to access the context menu etc.

I think we can do this in another PR though. I'm going to merge this in for the fix for now.

PR #1110, opened 2023-04-07
Fix - Only show shadow/line of sight layer when walls are on darkness 0 maps

Azmoria:
Reported in discord full brightness maps are darkened until a token is placed on the map. Even with no walls.

This was due to the shadow/line of sight layer being visible even when no walls are placed. No tokens = no line of sight, made a slightly dark map even without walls. This ensures there are more walls then just the border of the map before creating LoS shadows.

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -728,15 +728,18 @@ function reset_canvas() {
  	else if(window.DM){
  		$('#raycastingCanvas').css('opacity', '');
  	}
- 	if(darknessfilter == 0){
- 		$('#light_container').css('mix-blend-mode', 'unset');
- 		$('#light_container').css('background', '#0000');
- 		$('#light_container').css('opacity', '0.3');
- 	}
+ 	if(darknessfilter == 0 && window.walls.length>4){
+ 		$('#light_container').css({
+ 			'mix-blend-mode': 'unset',
+ 			'background':  '#0000',
+ 			'opacity': '0.3'
+ 		});
  	else{
- 		$('#light_container').css('mix-blend-mode', '');
- 		$('#light_container').css('background', '');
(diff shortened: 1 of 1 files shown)


Fog.js
```diff
@@ -1017,8 +1017,18 @@ function redraw_light_walls(clear=true){
 		}
 		let drawnWall = new Boundary(new Vector(x/adjustedScale/window.CURRENT_SCENE_DATA.scale_factor, y/adjustedScale/window.CURRENT_SCENE_DATA.scale_factor), new Vector(width/adjustedScale/window.CURRENT_SCENE_DATA.scale_factor, height/adjustedScale/window.CURRENT_SCENE_DATA.scale_factor))
 		window.walls.push(drawnWall);
-
 	}
+	let darknessfilter = (window.CURRENT_SCENE_DATA.darkness_filter != undefined) ? window.CURRENT_SCENE_DATA.darkness_filter : 0;
+ 	if(darknessfilter == 0 && window.walls.length>4){
+ 		$('#light_container').css('mix-blend-mode', 'unset');
```
Josh-Archer, line 1023: nit: I think you can stack these so the browser doesn't try rendering each time.

`$('#light_container').css({'mix-blend-mode': 'unset', 'background': '#0000',  'opacity': '0.3'});

Josh-Archer (approved):

Fog.js
Azmoria: this is true, didn't really think about that rerendering. I've used both ways of doing it throughout a lot of the code as have other people. It might be worth going through and adjust this in other places too.

I'll look into it if/when I've got the time. 

Thanks for the tip.

Issue #1121, opened 2023-04-09
More Performance Gains


Josh-Archer:
[Lighthouse results on performance and suggestions](https://googlechrome.github.io/lighthouse/viewer/?gist=3a0784dfd3f863bbc80129fe5ba4f702)

![image](https://user-images.githubusercontent.com/5824152/230749200-0b0e7b9e-f74d-446f-8027-26dc6eb31643.png)

PR #1122, opened 2023-04-09
Various improvements with running fog/vision as async or in background

Josh-Archer:
Just wanted any initial feedback. 

2 bugs right now I am working through:

**1. Drag through walls
2. On load in, DM needs to fire event to show the tokens to the player (i.e. drag a token)**

I introduced `async` so that we could have control return to the UI while we are processing things, caching` imagedata` in some places for the session (this might be causing the drag bug), and `scheduler` to schedule tasks off the main thread when its something that is important but shouldn't block UI either. 

Video attached shows how the game play flows and the second one is how main flows on same map, this is me spamming movement and dragging will work similarly. Alongside a small algo improvement moving to the for loop vs some (noticing decent improvement the bigger the file gets).  

The cached data and other improvements I am seeing ~1.5s per movement to ~1s (and the 1s is actually split up into 2 or 3 tasks, so that the user can still action the FE during it per recommendation from [Google blog](https://web.dev/optimize-long-tasks/))

https://user-images.githubusercontent.com/5824152/230798726-199679d3-f88b-4aef-b345-724cf502dbb2.mp4
https://user-images.githubusercontent.com/5824152/230798728-e0b79697-c518-4b23-9e1c-aa774cfa24d0.mp4

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -56,7 +56,7 @@ function mydebounce(func, timeout = 800){
   let timer;
   return (...args) => {
     clearTimeout(timer);
-    timer = setTimeout(() => { func.apply(this, args); }, timeout);
+    timer = setTimeout(async () => {await func.apply(this, args); }, timeout);
   };
 }
 function find_currently_open_character_sheet() {
--- a/Fog.js
+++ b/Fog.js
@@ -465,53 +465,74 @@ function check_token_visibility(){
 	}
 	else if(!window.NEXT_CHECK_TOKEN_VISIBILITY  || (window.NEXT_CHECK_TOKEN_VISIBILITY - Date.now() <  -1000)){
 		window.NEXT_CHECK_TOKEN_VISIBILITY = Date.now();
-		setTimeout(do_check_token_visibility(), 1);
+		do_check_token_visibility();
 		return;
 	}
 	else {
 		window.NEXT_CHECK_TOKEN_VISIBILITY = Date.now() + 1000;
-		setTimeout(do_check_token_visibility, 1000);
+		do_check_token_visibility;
 		return;
 	}
 }
 
-function do_check_token_visibility() {
+async function do_check_token_visibility() {
 	console.log("do_check_token_visibility");
 	let canvas = document.getElementById("fog_overlay");
 
 	if (canvas.style.diplay == "none")
 		return;
-	let ctx = canvas.getContext("2d");
+	let ctx = canvas.getContext("2d",  { willReadFrequently: true });
 	let canvas2 = document.getElementById("raycastingCanvas");
-	let ctx2 = canvas2.getContext("2d");
+	let ctx2 = canvas2.getContext("2d",  { willReadFrequently: true });
 
+	let promises = [];
 	for (let id in window.TOKEN_OBJECTS) {
+	if (id.includes(window.PLAYER_ID)) {
+		let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
+		let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
+		$(auraSelector).show();
+	};
+	 promises.push(new Promise(_ => setTimeout(_ => {
 		let left = (parseInt(window.TOKEN_OBJECTS[id].options.left.replace('px', '')) + (window.TOKEN_OBJECTS[id].sizeWidth() / 2)) / window.CURRENT_SCENE_DATA.scale_factor;
 		let top = (parseInt(window.TOKEN_OBJECTS[id].options.top.replace('px', '')) + (window.TOKEN_OBJECTS[id].sizeHeight() / 2)) / window.CURRENT_SCENE_DATA.scale_factor;
 		let pixeldata = ctx.getImageData(left, top, 1, 1).data;
-		let pixeldata2 = ctx2.getImageData(parseInt(window.TOKEN_OBJECTS[id].options.left.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, parseInt(window.TOKEN_OBJECTS[id].options.top.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeWidth()/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeHeight()/ window.CURRENT_SCENE_DATA.scale_factor).data;
 		let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
 		let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
-		let selector = "div.token[data-id='" + id + "']";
+		let tokenSelector = "div.token[data-id='" + id + "']";
 
 
 		let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
 		let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
 
 		//Combining some and filter cut down about 140ms for average sized picture
-		let someFilter = function(color, index) {return (index) % 4 == 0 && color == 255};
-		
-		if (!window.TOKEN_OBJECTS[id].options.revealInFog && (pixeldata[3] == 255 || (!pixeldata2.some(someFilter) && playerTokenHasVision && (window.CURRENT_SCENE_DATA.darkness_filter > 0 || window.walls.length>4)) || (playerTokenHasVision && window.CURRENT_SCENE_DATA.darkness_filter > 0  && (!is_token_under_light_aura(id) && pixeldata[2] == 0 && window.CURRENT_SCENE_DATA.darkness_filter > 0)))) {
-			$(selector + "," + auraSelector).hide();
+		let someFilter = function(ctx) {
+			let pixelData = ctx2.getImageData(parseInt(window.TOKEN_OBJECTS[id].options.left.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, parseInt(window.TOKEN_OBJECTS[id].options.top.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeWidth()/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeHeight()/ window.CURRENT_SCENE_DATA.scale_factor).data;
+			//return (index) % 4 == 0 && color == 255
+			for (let i = 0; i < pixelData.length; i+=4) {
+				if (pixelData[i] === 255) {
+					return true;
+				}
+			}
+
+			return false;
+		};
+
+		if (!window.TOKEN_OBJECTS[id].options.revealInFog 
+			&& (pixeldata[3] == 255
+				|| playerTokenHasVision 
+				&& ((window.CURRENT_SCENE_DATA.darkness_filter > 0  && !is_token_under_light_aura(id) && pixeldata[2] == 0 && window.CURRENT_SCENE_DATA.darkness_filter > 0))
+					|| ((window.CURRENT_SCENE_DATA.darkness_filter > 0 || window.walls.length>4) && !someFilter(ctx2)))) {
+			$(tokenSelector + "," + auraSelector).hide();
 		}
-		else if (!window.TOKEN_OBJECTS[id].options.hidden) {
-			let selectors = []
-			$(selector).css({'opacity': 1, 'display': 'block'});
+		else if (!window.TOKEN_OBJECTS[id].options.hidden ) {
+			$(tokenSelector).css({'opacity': 1, 'display': 'block'});
 			if(!window.TOKEN_OBJECTS[id].options.hideaura && id != playerTokenId)
 				$(auraSelector).show();
 			//console.log('SHOW '+id);
 		}
-	}
+	}, 700)));}
+
+	await Promise.all(promises);
 	console.log("finished");
 }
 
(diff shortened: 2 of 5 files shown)


Josh-Archer:
1121

Josh-Archer:
#1121

Fog.js
```diff
@@ -465,53 +465,74 @@ function check_token_visibility(){
...
 	let canvas2 = document.getElementById("raycastingCanvas");
-	let ctx2 = canvas2.getContext("2d");
+	let ctx2 = canvas2.getContext("2d",  { willReadFrequently: true });
 
+	let promises = [];
 	for (let id in window.TOKEN_OBJECTS) {
+	if (id.includes(window.PLAYER_ID)) {
+		let auraSelectorId = $(".token[data-id='" + id + "']").attr("data-id").replaceAll("/", "");
+		let auraSelector = ".aura-element[id='aura_" + auraSelectorId + "']";
+		$(auraSelector).show();
```
Azmoria, line 493: I'm not sure this needs to be here and we don't always want to show the players aura. I think this will just cause it to flash as it gets hidden below anyway.

Token.js
```diff
@@ -2372,14 +2373,15 @@ class Token {
...
-		toggle_player_selectable(this, token)
-		//check_token_visibility(); // CHECK FOG OF WAR VISIBILITY OF TOKEN
-		debounceLightChecks();
+		let promises = [];
+		promises.push(new Promise(() => setTimeout(this.build_stats(token), 100)));
+		promises.push(new Promise(() => setTimeout(this.toggle_stats(token), 100)));
+		promises.push(new Promise(() => setTimeout(this.update_health_aura(token), 100)));
+		promises.push(new Promise(() => setTimeout(this.update_dead_cross(token), 100)));
+		promises.push(new Promise(() => setTimeout(toggle_player_selectable(this, token), 100)));
+		promises.push(new Promise(() => setTimeout(debounceLightChecks(), 100)));
```
Azmoria, line 2382: Can you explain the benefit of making these timeouts?

Fog.js
```diff
@@ -465,53 +465,74 @@ function check_token_visibility(){
...
-			let selectors = []
-			$(selector).css({'opacity': 1, 'display': 'block'});
+		else if (!window.TOKEN_OBJECTS[id].options.hidden ) {
+			$(tokenSelector).css({'opacity': 1, 'display': 'block'});
 			if(!window.TOKEN_OBJECTS[id].options.hideaura && id != playerTokenId)
 				$(auraSelector).show();
 			//console.log('SHOW '+id);
 		}
-	}
+	}, 700)));}
```
Azmoria, line 533: Why the 700ms timeout for this or timeout at all really? Is this required for promises? I don't really know anything about them.

Azmoria (commented):
I'll have to test this in game after you've got it in a place you think it's ready. The async/promise stuff is beyond me. So as far as code I can't really review it atm but I'll play test when ready.

Azmoria:
So I just tried to replicate what you have in the video and I can't reproduce this issue on live or in beta. 

Redraw_light already only runs on drop for mouse movements and should only run when not spamming the movement with arrow keys. I am quite concerned about the flashing too. I'd rather short occasional lag then flashing personally. I'm not against including it if it feels better if we can get rid of the flashing/delayed update in vision/light.

This is what I'm getting on beta - redraw_light already shouldn't be run until you stop spamming for 250ms due to the debounce. I'm guessing your not spamming that fast - if that's the case I suggest we change that for keyboard movement only somehow or tell people to use mouse movement as it shouldn't be running anywhere near that often since it only runs on drop. 

shouldn't run till after

Fog.js
Azmoria: Ah maybe this is how you're splitting them up?

Fog.js
Azmoria: There is actually a bug with this below. 

```
	if(!window.TOKEN_OBJECTS[id].options.hideaura && id != playerTokenId)
				$(auraSelector).show();
				
```

should be 
```
	if(!window.TOKEN_OBJECTS[id].options.hideaura || id == playerTokenId)
				$(auraSelector).show();
				
```

Fog.js
```diff
@@ -3196,7 +3215,7 @@ function redraw_light(){
   		drawPolygon(offscreenContext, lightPolygon, 'rgba(255, 255, 255, 1)', true); //draw to offscreen canvas so we don't have to render every draw and use this for a mask
 	}    	
   }
-	context.drawImage(offscreenCanvasMask, 0, 0); // draw to visible canvas only once so we render this once
+	await scheduler.postTask(() => context.drawImage(offscreenCanvasMask, 0, 0), {priority: "user-visible"}); // draw to visible canvas only once so we render this once
```
Azmoria, line 3218: Just was looking at this again, I _think_ if you move the above draw of a full black square used to reset the canvas to be in the same task here it will fix the flashing. I think what's happening is the black rect is being drawn the clear then this is taking some time to run?

Josh-Archer: Ah yeah! That definitely would be it. Nice catch, thank you!

Josh-Archer: That priority basically says prioritize user input over this task and will only run over tasks the browser deems as "background"

Azmoria: ah that makes sense - thanks for explaining some of these comments are just my own lack of knowledge and picking your brain lol

Josh-Archer:
> I'll have to test this in game after you've got it in a place you think it's ready. The async/promise stuff is beyond me. So as far as code I can't really review it atm but I'll play test when ready.

Here is a good article about breaking up long running tasks if you want to read more about it. I'm mostly a backend guy, so this is very new to me as well :)

Fog.js
Josh-Archer: You give great feedback, so I always appreciate your reviews

Fog.js
```diff
@@ -465,53 +465,74 @@ function check_token_visibility(){
...
 	else if(!window.NEXT_CHECK_TOKEN_VISIBILITY  || (window.NEXT_CHECK_TOKEN_VISIBILITY - Date.now() <  -1000)){
 		window.NEXT_CHECK_TOKEN_VISIBILITY = Date.now();
-		setTimeout(do_check_token_visibility(), 1);
+		do_check_token_visibility();
 		return;
 	}
 	else {
 		window.NEXT_CHECK_TOKEN_VISIBILITY = Date.now() + 1000;
-		setTimeout(do_check_token_visibility, 1000);
+		do_check_token_visibility;
```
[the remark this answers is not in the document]
Josh-Archer: Ah yeah, this happened when I was playing with it being async. Thanks for catching

Fog.js
```diff
@@ -465,53 +465,74 @@ function check_token_visibility(){
...
 		let playerTokenHasVision = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
 
 		//Combining some and filter cut down about 140ms for average sized picture
-		let someFilter = function(color, index) {return (index) % 4 == 0 && color == 255};
-		
-		if (!window.TOKEN_OBJECTS[id].options.revealInFog && (pixeldata[3] == 255 || (!pixeldata2.some(someFilter) && playerTokenHasVision && (window.CURRENT_SCENE_DATA.darkness_filter > 0 || window.walls.length>4)) || (playerTokenHasVision && window.CURRENT_SCENE_DATA.darkness_filter > 0  && (!is_token_under_light_aura(id) && pixeldata[2] == 0 && window.CURRENT_SCENE_DATA.darkness_filter > 0)))) {
-			$(selector + "," + auraSelector).hide();
+		let someFilter = function(ctx) {
+			let pixelData = ctx2.getImageData(parseInt(window.TOKEN_OBJECTS[id].options.left.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, parseInt(window.TOKEN_OBJECTS[id].options.top.replace('px', ''))/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeWidth()/ window.CURRENT_SCENE_DATA.scale_factor, window.TOKEN_OBJECTS[id].sizeHeight()/ window.CURRENT_SCENE_DATA.scale_factor).data;
```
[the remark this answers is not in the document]
Josh-Archer: Ah, ill fix

Token.js
```diff
@@ -2138,13 +2143,14 @@ class Token {
...
 					let canvas = document.getElementById("raycastingCanvas");
-					let ctx = canvas.getContext("2d");
+					let ctx = canvas.getContext("2d", { willReadFrequently: true });
 					let playerTokenId = $(`.token[data-id*='${window.PLAYER_ID}']`).attr("data-id");
 					let playerTokenAuraIsLight = (playerTokenId == undefined) ? true : window.TOKEN_OBJECTS[playerTokenId].options.auraislight;
 					if(!window.DM && playerTokenAuraIsLight){
 						const left = (tokenPosition.x + (self.options.size / 2)) / window.CURRENT_SCENE_DATA.scale_factor;
 						const top = (tokenPosition.y + (self.options.size / 2)) / window.CURRENT_SCENE_DATA.scale_factor;
-						const pixeldata = ctx.getImageData(left, top, 1, 1).data;
+						const pixeldata = JSON.parse(sessionStorage.getItem("pixelData")) ?? ctx.getImageData(left, top, 1, 1).data;
```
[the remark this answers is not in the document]
Josh-Archer: Oh okay, I was just trying to cache the pixel data to save some of the fetching but it wouldn't work in this scenario. Removing this

Fog.js
Josh-Archer: Yes I was, your solution is much better. Thank you

Token.js
Josh-Archer: Was more just to keep it flowing for the user. I think the best bet would to just background debounce them. Ill play with background tasks and just debounce the ones not debounced already

Fog.js
Josh-Archer: The setTimeout will just prevent them running for longer than 700 ms. If they do, it will kill it. But that isn't needed really with the awaits because control will return back to the user. Gonna remove

Main.js
```diff
@@ -2431,10 +2431,7 @@ function init_ui() {
 
 
 	const rayCasting = $("<canvas id='raycastingCanvas'></canvas>");
-	rayCasting.css("top", "0");
-	rayCasting.css("left", "0");
-	rayCasting.css("position", "absolute");
-	rayCasting.css("z-index", "22");
+	rayCasting.css({"top": "0", "left": "0", "position": "absolute", "z-index": "22"});
```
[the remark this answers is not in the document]
Josh-Archer: Also more efficient, it makes browser render all in one go vs breaking it up potentially!

Fog.js
```diff
@@ -465,53 +465,74 @@ function check_token_visibility(){
...
+			for (let i = 0; i < pixelData.length; i+=4) {
+				if (pixelData[i] === 255) {
+					return true;
+				}
+			}
+
+			return false;
+		};
+
+		if (!window.TOKEN_OBJECTS[id].options.revealInFog
```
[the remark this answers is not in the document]
Josh-Archer: Im not sure im the best to name these. I am having difficult reading them too. I tried to make it a little simpler but it introduced bugs, so just went for the break up to try and make it look at little bit easier to read. I would like to do a PR to go through and make a lot of our if checks simpler where possible

CoreFunctions.js
```diff
@@ -56,7 +56,7 @@ function mydebounce(func, timeout = 800){
   let timer;
   return (...args) => {
     clearTimeout(timer);
-    timer = setTimeout(() => { func.apply(this, args); }, timeout);
+    timer = setTimeout(async () => {await func.apply(this, args); }, timeout);
```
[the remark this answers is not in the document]
Josh-Archer: Frees up the thread for user input. You can have debounce that still locks the thread potentially

Josh-Archer:
Just debounce - image

Await/Async - image


The blocking calls are reduced and if you actually expand out there are way fewer. You really only notice the difference on high end computer when you try moving between the redraw and debounce. For lower end computers, I think it will make a drastic difference for them. 

I think I covered all of your guys concerns, let me know if you have any more.

Josh-Archer:
> Just debounce - image
> 
> Await/Async - image
> 
> The blocking calls are reduced and if you actually expand out there are way fewer. You really only notice the difference on high end computer when you try moving between the redraw and debounce. For lower end computers, I think it will make a drastic difference for them.
> 
> I think I covered all of your guys concerns, let me know if you have any more.

Doesn't look like its faster because I zoomed in more on one lol. But the thread to redraw on movement is now 185~ms vs 260~ms on my computer and this means users can also input while that processing is happening. I think this will get amplified for slower computers

Fog.js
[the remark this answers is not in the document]
Azmoria:
```
		const hideThisTokenInFog = (!window.TOKEN_OBJECTS[id].options.revealInFog); //we want to hide this token in fog or darkness
		
		const inFog = (pixeldata[3] == 255); // this token is in fog
		
		const fullyOutOfLoS = (!someFilter(ctx2) && playerTokenHasVision && (window.CURRENT_SCENE_DATA.darkness_filter > 0 || window.walls.length>4)); //somefilter checks if any pixel is white - if it is this token isn't in darkness. We also make sure darkness > 0 and we have walls setup.
		
		const notInLight = (playerTokenHasVision && !is_token_under_light_aura(id) && window.CURRENT_SCENE_DATA.darkness_filter > 0); // this token is not in light, the player is using vision/light and darkness > 0
		
		if (hideThisTokenInFog && ( inFog || fullyOutOfLoS || notInLight )) {
			$(selector + "," + auraSelector).hide();
		}
		
		```
		
Here's what each section is and how we could split it up to something readable. The last one (notInLight) I was able to simplify as some of it was just leftover from old lighting and some duplicate code.

Azmoria (approved):
Very nice, I did add one more comment for the checking token visibility section for the hard to read code. I think what I've got in that comment explains each section and splits them up readable. If you want to include it with the simplifying checks PR or this one that's fine.

PR #1124, opened 2023-04-10
QoL - Adjust redraw light delay to capture holding an arrow key to move.

Azmoria:
While trying to replicate/figure out what was happening in #1122 I noticed that the debounce for redraw light would run before holding a movement key down triggered the next function call. 

For my setup 500ms captures the next call when holding down the key. I think this resolves some of the issue seen in the above PR as it won't stutter before letting go of the key for scenes with lots of walls or lower end pcs.

I'm going to hold off on this and the other improvements by Josh-Archer until next release since we're so close.

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -48,7 +48,7 @@ const debounceLightChecks = mydebounce(() => {
 		redraw_light();
 		if(!window.DM)
 			check_token_visibility();
-}, 250);
+}, 500);
 
 
 function random_token_color() {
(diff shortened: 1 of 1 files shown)


Josh-Archer:
This definitely helps. I think we still need to break the long running tasks up some. But I think this is a minimum should do while I continue to refine mine and apply your feedback (thanks for that by the way!)

Josh-Archer (approved):

Azmoria:
> This definitely helps. I think we still need to break the long running tasks up some. But I think this is a minimum should do while I continue to refine mine and apply your feedback (thanks for that by the way!)

sounds good 👍 - I updated some of my comments and added one where I think the flashing might be. Thanks for looking at all these improvements.

PR #1126, opened 2023-04-12
Fix - some player data sync issues, debounce token updates in update_pc_with_data

vlaminck:
This does 2 things:
1. MessageBroker is now listening to the DDB `character-update` event again. It was removed by mistake in 0.90
    * ~~I think this actually needs to be updated to fetch from the API because we expect the change to be included in the event which DDB doesn't send.~~ This has been done
2. `window.pcs` continues to update whenever we get any character-update event, but token updates are debounced. I did it this way because you can't pass variables into a debounced function.
    * ~~As I write this, I realize, I need to clear `window.PC_TOKENS_NEEDING_UPDATES` after updating the token~~ this has been done

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -18,13 +18,13 @@ const sendCharacterUpdateEvent = mydebounce(() => {
       characterId: window.PLAYER_ID,
       pcData: pcData
     });
+    update_pc_with_data(window.PLAYER_ID, pcData);
   } else {
     tabCommunicationChannel.postMessage({
       characterId: window.location.href.split('/').slice(-1)[0],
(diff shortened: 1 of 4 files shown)


Azmoria (approved):
Testing this it eliminates all the extra token messages.

Tabs and the iframes still work as expected sending data we collect locally to the abovevtt tab/main window. With ddb api as backup now.

Gives functionality back to the app when the api is cooperating. 

👍

Josh-Archer (approved):

cyruzzo (approved):
tested. seems to work nicely while solving duplication issues.

PR #1134, opened 2023-04-14
Debounce token movement events

Josh-Archer:
Fix spamming events to ws for each keypress

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -1233,9 +1233,9 @@ class MessageBroker {
 			if(window.all_token_objects[data.id] == undefined){
 				window.all_token_objects[data.id] = t;
 			}
-			t.sync = function(e) { // VA IN FUNZIONE SOLO SE IL TOKEN NON ESISTE GIA					
+			t.sync = mydebounce(function(e) { // VA IN FUNZIONE SOLO SE IL TOKEN NON ESISTE GIA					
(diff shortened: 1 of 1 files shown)


MessageBroker.js
```diff
@@ -440,7 +440,7 @@ class MessageBroker {
 			}
 
 			if (msg.eventType == "custom/myVTT/token" && (msg.sceneId == window.CURRENT_SCENE_DATA.id || msg.data.id in window.TOKEN_OBJECTS)) {
-				self.handleToken(msg);
```
Azmoria, line 443: This doesn't actually debounce this. Iirc something about it needing to be in a defined function. We use const in other places to define it outside the function we're using it in and it's clear it's working in sync so I assume that behaves similarly.

I don't think we want to either as otherwise it will only listen to the last token update - when moving a group of three tokens this message is sent three times but only 1 token would be listened too. I think we're ok listening to all the token events that come in - we just want to debounce sending them.

I'm approving the PR anyway though since it doesn't have an effect atm one of us can remove it later when we have time. Very nice - this feels more consistent with dragging and improves messages 🔥

Azmoria (approved):

MessageBroker.js
Josh-Archer: Ah this was accidental commit from when I was playing around. Thanks for catching. Im just gonna remove it real quick!

PR #1136, opened 2023-04-14
prevent messages from characters not in the campaign

Azmoria:
Messages from character tabs not in the campaign should be ignored.


Azmoria:
Maybe once we resolve issues like this maybe we can use tab communication to pull in data for characters with a button press on the sheet to add them as tokens in a campaign they don't exist in?

Josh-Archer (approved):
These are the best changes, one liners 🔥

PR #1138, opened 2023-04-15
Hex grids draft - function for drawing only

Azmoria:
I've looked at this a couple times but I think the grid draw here will work well. I'd like to include this as an option when we redo scene settings / the grid wizard. This includes the 2 common grid types for hexs. Just functions for drawing them at the appropriate size to tokens/rulers for now.

We'll have to setup a way to set the size in the wizard and adjust snap to grid - this shouldn't be too difficult just set an offset depending what column/row it falls in. We may want to do some transform on the canvas when we need to stretch it a bit when maps don't have perfect hexs.

Feel free to grab this if anyone wants to continue work on it.

hex grids

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -647,6 +647,78 @@ function redraw_grid(hpps=null, vpps=null, offsetX=null, offsetY=null, color=nul
 	}
 	gridContext.stroke();
 }
+function redraw_hex_grid(hpps=null, vpps=null, offsetX=null, offsetY=null, color=null, lineWidth=null, subdivide=null, dash=[], columns=true){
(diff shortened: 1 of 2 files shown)


Azmoria:
> This will be really nice to have. The offset math might get tricky, but even if the measurement isn't exact for the first pass, it'll still be way better than what we currently have.

Yeah, it's not too bad I got it mostly working, just need to do some minor adjustments I think. It'll need more testing either way especially if we have to skew the hex's at all but a bit at a time anyway.

snaphex

Josh-Archer (approved):
This is awesome!

PR #1139, opened 2023-04-15
feat - Add ability to display a source book chapters in-app

vlaminck:
# What
This just displays the source book url in a draggable resizable window. It doesn't expose a way to do it in-app. I'm not sure how we want to do that, but my first thought was to allow users to pick a source from a dropdown. Then we could scrape the table of contents to show the list of chapters as folders in the Journals tab. Clicking a chapter would then call `render_source_chapter_in_iframe`

# How
Simply call `render_source_chapter_in_iframe` with a url to a DDB source book. If the url has a `#section-id` that section will be the only thing rendered. For example, `render_source_chapter_in_iframe("https://www.dndbeyond.com/sources/doip/loggers-camp")` will display the entire Loggers Camp chapter, whereas `render_source_chapter_in_iframe("https://www.dndbeyond.com/sources/doip/loggers-camp#Preparations")` will only render the Preparations section within the Loggers Camp chapter.

# TODO
I'm not the best with the draggable resizable window stuff so I imagine there are some improvements that need to happen there. We do it in a billion places, and each one is slightly different. I tried to make a generic, reusable version in an attempt to DRY that up. It'd be nice to make this generic version more robust and use it in place of all the others at some point.

# Demo
https://imgur.com/a/nLavJxp

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -871,8 +871,10 @@ function find_or_create_generic_draggable_window(id, titleBarText, addLoadingInd
   if (addLoadingIndicator) {
     container.append(build_combat_tracker_loading_indicator(`Loading ${titleBarText}`));
     const loadingIndicator = container.find(".sidebar-panel-loading-indicator");
-    loadingIndicator.css("top", "25px");
-    loadingIndicator.css("height", "calc(100% - 25px)");
+    loadingIndicator.css({
+      "top": "25px",
+      "height": "calc(100% - 25px)"
+    });
   }
 
   container.show("slow")


Azmoria:
For the generic window we could probably set the default popout function as a part of the popout if statement. Then have a callback for anything we might need to do to the popout since those vary depending on the content we are popping out. This'll be really useful though. I think I could even add a style in tinyMCE that would run this function on clicking a link.

Azmoria (approved):

CoreFunctions.js
```diff
@@ -844,3 +844,125 @@ function areArraysEqualSets(a1, a2) {
...
+    "z-index": "10000",
+    "display": "none"
+  });
+
+  $("#site").append(container);
+
+  if (addLoadingIndicator) {
+    container.append(build_combat_tracker_loading_indicator(`Loading ${titleBarText}`));
+    const loadingIndicator = container.find(".sidebar-panel-loading-indicator");
+    loadingIndicator.css("top", "25px");
```
Josh-Archer, line 874: nit: I would stack these for painting efficiency. `loadingIndicator.css({"top": "25px", "height": "calc(100% - 25px)"});

Josh-Archer (approved):

PR #1172, opened 2023-04-24
persist scenes individually

vlaminck:


Azmoria (approved):
This resolves the issue for me for old and new maps

PR #1179, opened 2023-04-29
Support Copy Paste at Cursor and Copying of Notes with clean up of deleted token notes

Josh-Archer:
Fix notes not cleaning up correctly, eventually eating all localstorage

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -3545,10 +3545,9 @@ function paste_selected_tokens(x, y) {
 			window.TOKEN_OBJECTS[id].selected = false;
 			window.TOKEN_OBJECTS[id].place_sync_persist();
 		}
-		window.TOKEN_OBJECTS[newId].selected = true;
-		window.TOKEN_OBJECTS[newId].place_sync_persist();
+
 		if (id in window.JOURNAL.notes) {
-			window.JOURNAL.notes[newId] = window.JOURNAL.notes[id];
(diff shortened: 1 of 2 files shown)


Token.js
```diff
@@ -3543,6 +3547,16 @@ function paste_selected_tokens() {
 		}
 		window.TOKEN_OBJECTS[newId].selected = true;
 		window.TOKEN_OBJECTS[newId].place_sync_persist();
+		if (id in window.JOURNAL.notes) {
+			window.JOURNAL.notes[newId] = window.JOURNAL.notes[id];
+			let copiedNote = window.JOURNAL.notes[newId];
```
Azmoria, line 3552: I think you need to make a clone/copy of the note here. When I copy and paste tokens and then edit the note - refreshing changes all their notes to be the same. I think this is cause you're just referencing the original (not 100% sure though)

This section should also probably be moved above the place of the token so that the note appears when pasted instead of on the next token update.

Example video
https://i.imgur.com/U7kkCRj.mp4

Azmoria: looks good otherwise and good catch on the notes not being deleted on token delete.

Josh-Archer: Ah yeah, nice catch. I was thinking that was going to happen, but could get it to trigger. Thanks for catching!

Azmoria (approved):
Looks good 👍

PR #1185, opened 2023-05-04
Allow themed join/exit button and fix wrap bug

Josh-Archer:
Did this for fun on the side because seemed like easy nice thing to have while I was fixing issue with our button not wrapping correctly 

image

Now the button will never overlap and style (and updates with it) to the color this made it easier to find for new people and I think was a nice addition to bring attention to AVTT on the page. And added hover support

image
image

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -559,32 +559,26 @@ function inject_join_exit_abovevtt_button() {
   if ($(".ddbc-campaign-summary").length === 0) return;     // we don't have any campaign data
   if ($("#avtt-character-join-button").length > 0) return;  // we already injected a button
 
+  $(".ct-character-sheet-desktop > .ct-character-header-desktop").css({display: "inline-flex"})
   const desktopPosition = $(".ct-character-sheet-desktop > .ct-character-header-desktop > .ct-character-header-desktop__group--gap");
   const tabletPosition = $(".ct-character-sheet-tablet .ct-main-tablet > .ct-main-tablet__campaign");
   const mobilePosition = $(".ct-character-sheet-mobile .ct-main-mobile > .ct-main-mobile__campaign");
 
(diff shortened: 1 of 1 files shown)


CharactersPage.js
```diff
@@ -655,6 +649,13 @@ function observe_character_theme_change() {
             // console.log("theme_observer is calling find_and_set_player_color", mutation, node);
             const newColor = node.innerHTML.match(/#(?:[0-9a-fA-F]{3}){1,2}/)?.[0];
             if (newColor) {
+              let button = $("#avtt-character-join-button");
+              //$(".ct-character-sheet-desktop > .ct-character-header-desktop > .ct-character-header-desktop__group--gap");
+              let color = $(".ct-character-header-desktop__button").css("border-color") ?? "black";
```
Azmoria, line 654: You can find the theme color a little easier in all layouts with $('.dice-toolbar').css('--dice-color') 
Currently this doesn't look like it grabs it in mobile layout.

I find it a little distracting being fully colored rather than just the border like the other buttons. I'm sure I'd get used to it but if you guys like it though I'm good with putting it in I use my own theme anyway lol.

CharactersPage.js
```diff
@@ -559,32 +559,26 @@ function inject_join_exit_abovevtt_button() {
   if ($(".ddbc-campaign-summary").length === 0) return;     // we don't have any campaign data
   if ($("#avtt-character-join-button").length > 0) return;  // we already injected a button
 
+  $(".ct-character-sheet-desktop > .ct-character-header-desktop").css({display: "inline-flex"})
```
Azmoria, line 562: This unfortunately pushes the edit button off the other side of the screen for me when not full screen with beyond20 enabled. 

I think the button overlapping the player header tidbits (xp race etc.) is also only an issue with beyond20 enabled and not full screen?

It's still a good issue to fix since so many of our users use beyond20.

image

Azmoria:
I think this has mostly been resolved since this PR

Issue #1141, opened 2023-04-17
Auto center on Next token in Combat


Josh-Archer:
Suggested in https://discord.com/channels/815028457851191326/815028716040224818/1097272191713812500.

Auto center on the next token up in combat tracker when 'Next' button pressed.

Might be a good first time contributor feature.

Azmoria:
This is available as an option in the combat tracker settings now

PR #1988, opened 2024-03-21
Auto center on Next token in Combat

TachyonicSpace:
this pull request is for https://github.com/cyruzzo/AboveVTT/issues/1141
This also enables a setting for the user to want to have the scene move to different tokens on highlight, as well as adding that setting to the highlight message, and sending that message to the players once the DM hits the next button on the combat tracker

the only changes i made are: 
combat tracker, lines 264, 265, 266
message Broker, lines 598
Settings: 309 - 319

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -261,6 +261,9 @@ function init_combat_tracker(){
 				$("[data-current] button.openSheetCombatButton").click();
 			}
 			let newTarget=$("#combat_area tr[data-current=1]").attr('data-target');
+			window.TOKEN_OBJECTS[newTarget].highlight(!get_avtt_setting_value("highlightScrolling"));
+			window.MB.sendMessage("custom/myVTT/highlight", {id: newTarget});
+			
 			if(window.TOKEN_OBJECTS[currentTarget] != undefined){
 				delete window.TOKEN_OBJECTS[currentTarget].options.current;
 				delete window.TOKEN_OBJECTS[currentTarget].options.round;
@@ -451,9 +454,9 @@ function ct_add_token(token,persist=true,disablerolling=false){
 	if(token.options.stat)
 		entry.attr('data-stat', token.options.stat)
 
-
+	
 	let video = false;
-
+	
 	if(['.mp4', '.webm','.m4v'].some(d => token.options.imgsrc.includes(d))){
 		img = $("<video disableRemotePlayback muted width=35 height=35 class='Avatar_AvatarPortrait__2dP8u'>");
 		video = true;
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -595,7 +595,7 @@ class MessageBroker {
 			}
 			if (msg.eventType == "custom/myVTT/highlight") {
 				if (msg.data.id in window.TOKEN_OBJECTS) {
-					window.TOKEN_OBJECTS[msg.data.id].highlight(true);
+					window.TOKEN_OBJECTS[msg.data.id].highlight(!get_avtt_setting_value("highlightScrolling"));
 				}
 			}
 			if (msg.eventType == "custom/myVTT/pointer") {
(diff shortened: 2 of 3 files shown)


MessageBroker.js
```diff
@@ -1227,7 +1227,7 @@ class MessageBroker {
 		//Security logic to prevent content being sent which can execute JavaScript.
 		data.player = DOMPurify.sanitize( data.player,{ALLOWED_TAGS: []});
 		data.img = DOMPurify.sanitize( data.img,{ALLOWED_TAGS: []});
-		data.text = DOMPurify.sanitize( data.text,{ALLOWED_TAGS: ['video','img','div','p', 'b', 'button', 'span', 'style', 'path', 'svg', 'a'], ADD_ATTR: ['target']}); //This array needs to include all HTML elements the extension sends via chat.
+		data.text = DOMPurify.sanitize( data.text,{ALLOWED_TAGS: ['img','div','p', 'b', 'button', 'span', 'style', 'path', 'svg', 'a'], ADD_ATTR: ['target']}); //This array needs to include all HTML elements the extension sends via chat.
```
Azmoria, line 1230: Just a couple conflicts in this file to sync to current main.

MessageBroker.js
```diff
@@ -1842,47 +1842,26 @@ class MessageBroker {
...
-			}
-
-		
-			window.reconnectAttemptAbovews++;
-			if(window.reconnectAttemptAbovews > 5)
-				return;
-
-			if(window.reconnectAttemptAbovews < 5){
-				let msgdata = {
+			let msgdata = {
```
Azmoria, line 1845: conflict as above

MessageBroker.js
```diff
@@ -1842,47 +1842,26 @@ class MessageBroker {
...
-				let msgdata = {
-					player: window.PLAYER_NAME,
-					img: window.PLAYER_IMG,
-					text: `<span><p>You have disconnected from the AboveVTT websocket 5 times. It is likely you are experiencing connection issues. </p><p>Reconnect messages/forced reconnect will be disabled until refresh.</p><p>This could be caused by a VPN, anti-tracker, adblocker, firewall, school/work network settings, or other extention/program.</p></span>`,
-					whisper: window.PLAYER_NAME
-				};
-				this.inject_chat(msgdata);
-			}
-					
+			};
```
Azmoria, line 1850: conflict as above

Settings.js
```diff
@@ -306,6 +306,17 @@ function avtt_settings() {
 			],
 			defaultValue: false
 		}
+		,
+		{
+			name: "highlightScrolling",
```
Azmoria, line 311: I think we should note it only affects the combat tracker and token pings. 

Find token buttons for example should always center (you haven't changed those I don't think which is good). 

Maybe we consider having default value false based on feedback in beta. Either way is good with me though. We can consider moving the setting to a toggle in UI after too.

CombatTracker.js
```diff
@@ -261,6 +261,9 @@ function init_combat_tracker(){
 				$("[data-current] button.openSheetCombatButton").click();
 			}
 			let newTarget=$("#combat_area tr[data-current=1]").attr('data-target');
+			window.TOKEN_OBJECTS[newTarget].highlight(!get_avtt_setting_value("highlightScrolling"));
+			window.MB.sendMessage("custom/myVTT/highlight", {id: newTarget});
```
Azmoria, line 265: I think we should consider putting the first .highlight() in ct_load further down in this file.

Can remove the sendmessage if we do that just save on server messaging really.

Azmoria (commented):
Looks good mostly just a couple conflicts to sync to main and maybe move the highlight down as per comment. 

👍

Azmoria:
I used this to help create #2018 

I've also added an option for tie breaker token initiative for DMs.

Player and DM setting will save seperately for highlighting token. This allows for it to be set for DM but not player view when playing in person. I moved the setting to a combat tracker settings menu which can be found with the cog in the combat tracker window titlebar.

image

Also added you as a contributor on the splash screen as TachyonicSpace - if you would prefer another name let me know.

image

PR #2453, opened 2024-09-05
Demo of possible feature for token duration/age

jeffsenn:
Adds agedToken/age/maxAge to token options.  
Displays age on tokens that have it (flag with red if age >= maxage).
Modifies age at end of token's turn in combat tracker.
Screenshot 2024-09-05 at 11 20 58 AM

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -18,6 +18,27 @@ function combatmydebounce(func, timeout = 800){ // we need to figure out where t
   };
 }
 
+
+function adjust_age(token, amt) {
+	if(!token.options.agedToken) return false;
+	token.options.age = (parseInt(token.options.age) || 0) + 1;
+	return true;
+}
+
+function adjust_condition_duration(token, amt) {
+	function anycond(v, changed) {
+		for(let i in v) {
+			if(!isNaN(parseInt(v[i].duration))) {
+				console.log("YEP",v[i]);
+				v[i].duration += amt;
+				changed = true;
+			}
+		}
+		return changed;
+	}
+	return anycond(token.conditions, anycond(token.options.custom_conditions, false));
+}
+
 function init_combat_tracker(){
 	window.ROUND_NUMBER =1;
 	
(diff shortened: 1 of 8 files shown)


Azmoria:
Tracking durations of things is a good idea/been requested a bunch. 

Couple comments

• There doesn't seem to be a way to just reset the tracker here. I do see hitting previous counts back. I think I may have got it stuck by hitting 'reset' in the tracker though. Part of the issue with this is player tokens will keep the data between combats without a way to reset them. 

• It might be better to have a tracking option for any given condition - rather than just 1 tracker for the token. I'm not sure the ideal interface option for this but a simple number input beside conditions might work. 

This way you could have multiple trackers for different status effects. Custom conditions for straight tracking a number. For this it might be better to count down and remove the condition at the end though - as that's how most conditions would operate.  Moving the timer icon or number to the condition data itself.

We could just add this to the condition data instead of adding a new age option. This eliminates needing the options in the menu all together. 

Ie. 
```
token.options.conditions = [
  {
    name: "Blinded"
  }, 
  {
    name: "Rage",
    duration: "10"
   }
] 

token.options.custom_conditions = [ 
  {
    name: '#1A6AFF', 
    text: 'test', 
    duration: '1'
  '},
]
```

Happy to hear your thoughts on why you might prefer it as a token timer rather then condition though.

jeffsenn:
> • There doesn't seem to be a way to just reset the tracker here. I do see hitting previous counts back. I think I may have got it stuck by hitting 'reset' in the tracker though. Part of the issue with this is player tokens will keep the data between combats without a way to reset them.

I added age/maxAge to the stats on the token edit (so you can reset or adjust).


> • It might be better to have a tracking option for any given condition - rather than just 1 tracker for the token. I'm not sure the ideal interface option for this but a simple number input beside each condition might work. Maybe with a stopwatch icon beside it to indicate it's for timing durations....

Completely agree that I want that TOO.... but->

> Happy to hear your thoughts on why you might prefer it as a token timer rather then condition though if that's the case.

My main use of this (b/c of the players I happened to have now) is for various summoned things that have a duration (animated objects, spiritual weapon, flaming sphere, summoned animals ... ) that tend to have a token that I DO want to have a sticky age.  Yes I suppose one could ALSO give them some sort of  "summoned" condition with a duration.... but it seems super easy to just have the tokens set up with an age and pop them onto the map.

This was a faster (seems like low impl add) to get to my first needs.  I'll try to take a look at the conditions idea too....

Azmoria:
I can always build off this too if you don't have the time to do the condition stuff - I'd probably use a custom condition for summons but I see your point.

jeffsenn:
I started the implementation of condition duration as well (combat tracker side and badging)
todo:
  - figure out how to put UI on token menu (and how it works for non-custom conditions on player tokens); should set "duration" on token.conditions and/or token.options.custom_conditions
  - tweak badge visuals

jeffsenn:
Todo (in case someone picks up): 
- fix layout of token menu condition popout (duration input field is not right)
- tweak badging on conditions for readability

Azmoria:
I'll pick it up after this release - thanks for doing what you've done already, do you have a preferred name for the contribution list? (this name, discord name, other?)

jeffsenn:
Thanks!!

> ...do you have a preferred name for the contribution list? (this name, discord name, other?)

EIther - I'm not particular.

Azmoria:
This is in #2459 now. I did make it so token age is only available prior to dropping a token similar to roll hp. I'll think on how to include it in the token menu in the future. Options are already pretty cluttered there atm.

PR #2530, opened 2024-10-02
WIP: initial try for Safari Web Extension

jeffsenn:
Don't merge this yet!!

But: just in case anyone wants to try this: the key difference in Safari is that the page URL pattern for an extension can NOT match query params.  The only change here is a re-org to combine Load.js and LoadCharacter.js to dynamically determine what to load on each type of page.  With that change and a few small adjustments to browser extension calls, many things DO now work on Safari.


jeffsenn:
Closing for awhile while I test - maybe it will reduce churn here

PR #2554, opened 2024-10-13
fix: fix alpha argument being ignored in WaypointManagerClass.draw()

MonstraG:
This was clearly an incorrect placement of `)`


Azmoria:
Thanks - good catch, I think it was defaulting to 1 anyway in css which is probably why I never noticed it.

PR #2555, opened 2024-10-13
refactor: replace setInterval with replaceAnimationFrame in fadeout

MonstraG:
This should give nice fadeout animations, whilst allowing for frameskips if overloaded


Azmoria:
Thanks for this. Do you have a prefered name for the contributors list? Also if you're on discord what's your @ so we can give you dev role when this goes live.

Azmoria:
Awesome - thanks you'll get the role when the release goes live. I put up a PR to add you to the contributors list I'll merge it later today probably and push a beta at the same time.

PR #2561, opened 2024-10-14
fix: remember labelX and labelY to avoid label jump on mouse up

MonstraG:
* refactor: optimize jquery calls in drawWaypointSegment
* refactor: remove ruler-svg-bobbles
* refactor: optimize draw loop by moving all dom operations out
* refactor: remove midlineLabels as unused

I've realized it would take ages merging these one by one.

You can review each commit one by one, each is self contained.

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -251,15 +251,13 @@ class WaypointManagerClass {
 	}
 
 	/**
-	* Draw the waypoints, note that we sum up the cumulative distance, midlineLabels is true for token drag
-	* as otherwise the token sits on the measurement label
-	* @param midlineLabels {boolean} always false
+	* Draw the waypoints, note that we sum up the cumulative distance
 	* @param labelX {number | undefined} if provided, move last text of last waypoint there
 	* @param labelY {number | undefined} if provided, move last text of last waypoint there
 	* @param alpha {number | undefined} set alpha of drawings to this, 1 if unset
 	* @param playerId {string | false | undefined} `window.PLAYER_ID` if unset
 	*/
-	draw(midlineLabels, labelX, labelY, alpha = 1, playerId=window.PLAYER_ID) {
+	draw(labelX = undefined, labelY = undefined, alpha = 1, playerId=window.PLAYER_ID) {
 		const rulerContainer = this.getOrCreateDrawingContainer(playerId);
 
 		// update alpha for the entire container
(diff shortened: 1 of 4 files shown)


Azmoria:
Bobbles are the end/mid points of the lines. They do show up on main and should be left in. 

I also only see the jump up when dragging a token. This is due to it keeping it outside the token while dragging then moving it to the center spot on release. This is on purpose as we want the text off the token as we drag it and the text close to the vertex when we drag the token again. With this PR you currently get a very large gap when dragging a gargantuan creature.

For example this PR:

image

vs main:

image

Mostly the performance issue for the ruler is when dragging tokens with the console open - not the ruler function itself. Since people don't play with the console open I've ignored it. I agree it'd be nice if it was better but I think this has to do with the token dragging function more then the ruler function. 

This is originally why it had to be specifically enabled when dragging tokens - although there's been improvements since then that mostly eliminated that issue and contained it to having the console open.

Happy to have some of these other changes if they improve performance while maintaining the other functionality above.

As a note in testing this PR I have noticed sometimes a multi segmented ruler (ruler tool + right click) fails to keep it's first segment positioned correctly but I don't think it's related to this PR.

MonstraG:
Thanks, I'm gonna revert the bubbles and labelX/labelY changes.

I drag and measure with tokens much more than with standalone ruler, so both performance and the jump seem to bug me more than you)

Azmoria:
Fair - I also have token drag measurement on all the time I just don't see performance issues on my end except with the dev tools open. 

As far as I can tell the dev tools don't like removing/adding the ruler dom elements repeatedly while dragging a token. Mind you I do play with a bunch of people who have higher end gaming PCs so it's pretty easy for us to miss those things. 

I think for the jumping the only other option I can think of is putting the text in the center regardless of token but I think that looks worse then it jumping when you drop the token and pick it up again. This might just be preference.

MonstraG:
Here's how these bobbles look like on my screen:
image

That's why I couldn't find them.

MonstraG:
Figured out why they did not show up on firefox. 

Firefox thinks that `calc(2 / 1.5)` is not a valid value, `calc(2px / 1.5)` is.

Azmoria:
I hadn't noticed that in firefox - I merged a PR that should fix that. There was a couple other places where the units were missing in the css. I applied the same changes manually to this PR and it seems to be fine. I'll do some more testing and merge in a bit if you think it's ready to merge.

Looks like this in firefox for me now.

image

MonstraG:
Merged main in, checked it out in firefox, bobbles show up now.

lgtm)

PR #2575, opened 2024-10-18
refactor: small cleanup for pc-does-not-understand-language feature

MonstraG:
I noticed feature request in discord, and wanted to do it. Turns out @Azmoria beat me to it by 12 hours or something.

I noticed that it sorta may crash if pc happens to not have any languages profficiency (`knownLanguages.includes`), and while there I've decided to cleanup tabulation a bit, hope that's okay.


Azmoria:
good call on undefined language to set an empty array.

PR #2583, opened 2024-10-21
refactor: provide bg color for splash screen so it's never transparent

MonstraG:
I've reinstalled os and was reminded that splashscreen is transparent on initially before image is fetched for the first time becuase there is no background color.

Maybe this could be classified as a fix tho


Azmoria:
I'm unable to notice the background not loading in time. I take you're word for it though that it's loading in without a background. Merged.

Azmoria:
yeah, it just loads the image before the splash is visible on my end is what I meant. No big deal though.

PR #2585, opened 2024-10-21
refactor: improve language selector design a bit

MonstraG:
Previous implementation used position absolute to place the button, did not reset the background so on firefox it would stand out and the button was very narrow.

Now it's slightly nicer both in code and in visuals, I hope.

I would actually prefer this selector to be the same as the mui one on the top of the game log, but I didn't find a way to do that

I'm also not quite sure that I changed popoutGamelogCleanup correctly
(especially
```
$(childWindows["Gamelog"].document).find("#chat-language").off('change.value').on('change.value', function(){
		$("#chat-language").val($(this).val());
```

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -2086,13 +2086,16 @@ function monitor_character_sidebar_changes() {
  * @returns
  */
 function inject_chat_buttons() {
-	if ($(".glc-game-log").find("#chat-text").length > 0) {
-		// make sure we only ever inject these once. This gets called a lot on the character sheet which is intentional, but just in case we accidentally call it too many times, let's log it, and return
+	const gameLog = $(".glc-game-log");
+
+	if (gameLog.find("#chat-text").length > 0) {
+		// Make sure we only ever inject these once.
+		// This gets called a lot on the character sheet which is intentional.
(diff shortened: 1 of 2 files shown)


Azmoria:
This makes things out of alignment and too much variation between browsers. I think it'd be best to just leave it as is and make minor adjustments if necesarry.

This PR:

image


Main:
image

MonstraG:
> I think it'd be best to just leave it as is and make minor adjustments if necessary.
ok)

> I don't mind the repeat selector cleanup though. 
It would be easier to create a separate PR for that

> screenshots
Hilariaously, none of these doesn't quite match how  it looks on my branch. Anyway..

Azmoria:
Just checked out with this PR again to double check and I see the same.  

That's odd you're seeing something different. Maybe it's just different views or browsers.

The above are a combination of underdark/normal sheets and DM. Some of them chrome some firefox. I do think the dark chat bar in the top image is actually not refreshed on this PR and was main at the time. In underdark mode the text was showing white for the chat bar too.

Anyway since you're seeing different if you see something broken on current main let me know.

PR #2603, opened 2024-10-25
Beta Fix - snap to grid should not be applying to projector mode when zoom lock is enabled.

Azmoria:
This might be a cool feature to have separately later.


Azmoria:
This might be a cool feature to have separately later.

PR #2724, opened 2024-12-09
Snap to Grid

firelightrpg:
Square grid snapping 

Allows user to draw with grid snapping. Intended for walls, but works for any call to Fog.js::get_event_cursor_position like draw, light, etc.

Uses the same snap logic as the token - when snap to grid is enabled, holding down the control key will disable it and vice versa. 

Future updates: 
- add hex (vert and horizontal) snapping
- door and erase snapping is awkward, will address in future PR


Azmoria:
Nice - works well for a first pass.

Few comments:

> • Might be worth it to have a toggle button in the menu and not just the hotkey. Something like the daylight toggle.

> • For snap to grid hotkey consistency I think maybe using our current snap ctrl bind might be better. Then you can just check if `window.toggleSnap` is true for drawings in the same way we check tokens. 
> 
> No need to worry about extra events in that function then. I also find capslock is often set by accident (It happens far to often with password fields for me - the ones that don't tell me I have it on are the worst 🤣)
> 
> Alternatively add the hotkey to `KeypressHandler.js`. Here's what would probably need to be added for the `window.toggleSnap` one to work in all cases.
> 
> ```
> Mousetrap.bind('ctrl+shift', function () {
>     shiftHeld = true;
>     ctrlHeld=true;
>     window.toggleSnap=true;
> }, 'keydown');
> 
> Mousetrap.bind('ctrl+shift', function () {
>     shiftHeld = false;
>     ctrlHeld=false;
>     window.toggleSnap=false;
> }, 'keyup');
> 
> ```


> • `window.CURRENT_SCENE_DATA.gridType === "1" ` should probably be `window.CURRENT_SCENE_DATA.gridType == "1" ` since that will match when it's just the number 1 and not string - which sometimes gets set by the form reader (or default options I forget which).

> • I think it's also missing adding the offset back to the snapping locations after rounding - my drawings were coming close but not exact until I set offset to 0 in manual grid. Offsets can be found in `window.CURRENT_SCENE_DATA.offsetx` and `window.CURRENT_SCENE_DATA.offsety`

> • `Math.floor` seems to land on my grid lines closer to my mouse then `Math.round` does. `Math.round` is skipping down to the next grid line when drawing on one. This might be map dependent though so maybe just more testing to see which feels better. This may also be different due to adding the offset back here - the offset might have been close to a square in height.
>
> Example: https://i.imgur.com/JkwVlrL.mp4

PR #2922, opened 2025-02-21
QoL - Move drawings under fog overlay

Drenam1:
Prior to this PR, drawings would appear above the fog layer (so players could see the drawings outside of their vision)

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -1172,14 +1172,16 @@ function draw_wizarding_box() {
 	}
 
 }
-function ctxScale(canvasid){
+function ctxScale(canvasid, transform){
 	let canvas = document.getElementById(canvasid);
 	canvas.width = $("#scene_map").width();
   	canvas.height = $("#scene_map").height();
-	$(canvas).css({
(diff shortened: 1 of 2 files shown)


Azmoria:
This is not something that should apply to all drawings. This has been requested as an option in the drawing menu though.

Also all canvas' need to be scaled and transformed or they break after a certain size/scaling of the map due to canvas size limits. It also drastically decreases performance if we don't scale them up due to more pixels.

 I think the way you have it here would make it so only part of the map is drawable when scaled though.

Drenam1:
That makes sense! I guess my group primarily uses drawings for different reasons - I use them to draw things on-the-fly on more generic maps (like putting a pool of water into a room or something) and I didn't consider other, probably more common, use cases. I'll close this PR and maybe start looking into a under-darkness drawing layer. Thanks!

Azmoria:
Sounds good - let me know. I can also do some work towards this if you don't have time finally got around to going through some older requests again.

PR #2923, opened 2025-02-21
Fix - DM map is used on DD2VTT map files if available

Drenam1:
If a map used a DD2VTT (or a UVTT) file as its base, and an image file is added as a DM map, the DM map would not appear for the DM (IE  the DM would see the player map)

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -1874,10 +1874,21 @@ class MessageBroker {
 				if(data.UVTTFile == 1){
 					build_import_loading_indicator("Loading UVTT Map");
 					try{
-						data.map = await get_map_from_uvtt_file(data.player_map);
+						if(window.DM && data.dm_map && data.dm_map_usable){
+							data.map = await get_map_from_uvtt_file(data.map)
+						}
+						else{
+							data.map = await get_map_from_uvtt_file(data.player_map);
+						}			
 					}
 					catch{
-						data.UVTTFile = 0;
+						console.log('non-UVTT file found for map')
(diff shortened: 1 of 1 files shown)


Azmoria:
This is a good idea but I think we should do this above this actually - so that either the player or dm map can be an image or UVTT file, while keeping the initial UVTT import file scaling. 

Since when we load the UVTT file initally it scales it to be a certain size we want the image to match that which yours does for the DM map but currently it messes up if we switch the player map. It's probably also better to maintain being able to use a UVTT file for either of the lines incase they export both as UVTT.

At line 1874 I think we replace this block with the below. 

```
if(data.UVTTFile == 1){
	build_import_loading_indicator("Loading UVTT Map");
	try{
		if(window.DM && data.dm_map && data.dm_map_usable){
			data.map = await get_map_from_uvtt_file(data.map)
		}
		else{
			data.map = await get_map_from_uvtt_file(data.player_map);
		}			
	}
	catch{
		console.log('non-UVTT file found for map')
		if(window.DM && data.dm_map && data.dm_map_usable){
			data.map = data.dm_map;
		}
		else{
			data.map = data.player_map;
		}
	}
}				
```

Drenam1:
I've made the change you recommended and reverted my original change. Good shout!

Azmoria:
Do you have a preferred name for the contributors list (Main.js line 1412)? And are you on our discord for dev role when it goes to production?

Drenam1:
Awesome! Just Drenam1 is fine thanks :)

Issue #1847, opened 2024-01-29
Request: Option for drawings below darkness/fog


Azmoria:
Add an option similar to tokens to allow a drawing to be below darkness/fog. Probably requires a second canvas in the map container.

PR #2930, opened 2025-02-23
QoL - Option for drawings below darkness/fog

Drenam1:
Added new drawing layer;
Added location options in the draw dropdown;
Add drawing to layer based on location option selected;

The code on line 1481 should make it so that any pre-existing drawings will default to 'above fog', the way they currently are, so this shouldn't break people's current drawings.
The only things I'm a bit iffy on for these changes are

1. The location I put the new _drawOverlayUnderFogDarkness_ draw layer in the HTML - I'm happy for this to be moved somewhere that makes more sense, this was just a spot that worked.
2. The labelling in the dropdown - I wanted to make sure that the labels were short enough to not needlessly widen that dropdown, but they may not be descriptive enough? Open to change.
3. Clearing the draw layer - currently, if you click 'clear draw' in the draw dropdown, it clears both the above fog and below fog draw layers. Maybe it should only clear the currently-selected layer?

New menu options
image

DM view (yellow text added by me)
image

Player view (with light/darkvision)
image

Player view (without light/darkvision)
image

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -1172,14 +1172,16 @@ function draw_wizarding_box() {
 	}
 
 }
-function ctxScale(canvasid){
+function ctxScale(canvasid, doNotScale=false){
 	let canvas = document.getElementById(canvasid);
 	canvas.width = $("#scene_map").width();
   	canvas.height = $("#scene_map").height();
-	$(canvas).css({
-		'transform-origin': 'top left',
-		'transform': 'scale(var(--scene-scale))'
-	});
+  	if (!doNotScale) {
+		$(canvas).css({
+			'transform-origin': 'top left',
+			'transform': 'scale(var(--scene-scale))'
+		});
+	}
 }
 
 function reset_canvas(apply_zoom=true) {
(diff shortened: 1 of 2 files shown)


Drenam1:
For #1847

Azmoria:
One small change to fix for scaled maps - what you had before in `ctxscale` where it wasn't double scaling the underdarkness/fog drawings.

I can merge and do this later if you'd like though. Looks good otherwise awesome work👍 

I think this is basically what you had and keeps it simple

```
```function ctxScale(canvasid, doNotScale=false){
	let canvas = document.getElementById(canvasid);
	canvas.width = $("#scene_map").width();
  	canvas.height = $("#scene_map").height();
  	if(!doNotScale){
  		$(canvas).css({
  			'transform-origin': 'top left',
  			'transform': 'scale(var(--scene-scale))'
  		});
  	}
}

function reset_canvas(apply_zoom=true) {
	let sceneMapWidth = $("#scene_map").width();
	let sceneMapHeight = $("#scene_map").height();

	$('#darkness_layer').css({"width": sceneMapWidth, "height": sceneMapHeight});
	$("#scene_map_container").css({"width": sceneMapWidth, "height": sceneMapHeight});

	ctxScale('peer_overlay');
	ctxScale('temp_overlay');
	ctxScale('draw_overlay_under_fog_darkness', true);
	ctxScale('fog_overlay');
	ctxScale('grid_overlay');	
	ctxScale('draw_overlay');
	ctxScale('walls_layer');
	ctxScale('elev_overlay');
```

Drenam1:
I've added the doNotScale option to ctxScale. Should be good to merge!
Thanks!

Issue #2411, opened 2024-08-16
Request: A way to close notes from the DM screen.


Azmoria:
Requested on discord

A way to close notes either through projector mode or just force closed for players. 

The primary use case here being for in person games using a tv table/screen. Online games I think players deciding when to close theirs is fine.

PR #2931, opened 2025-02-23
Added DM-only button to force close a note for all players

Drenam1:
#2411 

Using the existing window.MB.sendMessage functionality and sending _popup: false_ instead of true. When popup == false, find the note popup with data-id == note.id and remove it from the HTML. 

Tested with 
1. Notes that were originally forced open by the DM - forces them closed.
2. Notes that were originally shared to the player and opened by them - forces them closed.
3. Multiple notes open - only forces the one note closed.

image

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -791,10 +791,12 @@ class MessageBroker {
 					if(msg.data.id in window.TOKEN_OBJECTS){
 						window.TOKEN_OBJECTS[msg.data.id].place();			
 					}			
-					const openNote = $(`.note[data-id='${msg.data.id}']`);	
-					if(msg.data.popup){
+					const openNote = $(`.note[data-id='${msg.data.id}']`);
+					// If the 'Open' button is clicked OR the note is already opened by a player and it is saved 
+					// by the DM, the note gets refreshed.
+					if (msg.data.popup == true || (msg.data.popup == undefined && openNote.length != 0)){
 						window.JOURNAL.display_note(msg.data.id);
-					}else {
+					} else if (msg.data.popup == false) {
 						openNote.remove();
 					}
 					
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -790,10 +790,13 @@ class MessageBroker {
 					
 					if(msg.data.id in window.TOKEN_OBJECTS){
 						window.TOKEN_OBJECTS[msg.data.id].place();			
-					}				
-					if(msg.data.popup)
+					}			
+					const openNote = $(`.note[data-id='${msg.data.id}']`);	
+					if(msg.data.popup){
 						window.JOURNAL.display_note(msg.data.id);
-					const openNote = $(`.note[data-id='${msg.data.id}']`);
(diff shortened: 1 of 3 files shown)


MessageBroker.js
```diff
@@ -790,10 +790,13 @@ class MessageBroker {
...
 					if(msg.data.id in window.TOKEN_OBJECTS){
 						window.TOKEN_OBJECTS[msg.data.id].place();			
-					}				
-					if(msg.data.popup)
+					}			
+					const openNote = $(`.note[data-id='${msg.data.id}']`);	
+					if(msg.data.popup){
 						window.JOURNAL.display_note(msg.data.id);
-					const openNote = $(`.note[data-id='${msg.data.id}']`);
+					}else {
```
Azmoria, line 797: We use this same message to send updated data for notes. I think we would explicitly only want this to close when `msg.data.popup === false` so  its only when the button is pressed - maybe `else if`. 

If the player gets an update without the button being pressed maybe a refresh of the note would be in order though? So maybe a third else to refresh the note or close and reopen - thoughts? can be a separate PR though

Azmoria (commented):
Looks good 👍 just one line comment for review.

MessageBroker.js
Drenam1: Fair enough, that makes sense, I'll update it to an else-if.

For the second point - do you mean if the 'force open by players' button is clicked and the user has it open already, it closes and reopens the note to force a refresh?

Azmoria: I was thinking more just if they receive the message without the popup data. This means it's just updating the note from the DM's side/edits. Maybe we could refresh the note with new data if it's already openned - instead of the DM telling the player an edit happened. This can be a separate PR though.

one example comes from here 

```
	save_onsavecallback: function(e) {
		.
		.
		.
		if(self.notes[note_id].player){
			window.MB.sendMessage('custom/myVTT/note',{
				id: note_id,
				note:self.notes[note_id]
			});
		}
				
```

Drenam1: I think I understand what you're getting at though - I've made a change that I think is along the lines of what you wanted. If it is, then great; if not, then I can change it so its just the else if and we can make this a later PR.

The first condition of the if statement now checks
```(msg.data.popup == true) || (msg.data.popup == undefined && openNote.length != 0)```
This means that the code in the first block will run if the 'Open' button has been clicked OR if the note has been saved and the player already has the note open refreshes the opened note window. 

Quick demo of it in action below. Only issue I have is that if the player moves the note window around, refreshing it puts it back into the default location on the screen. Maybe not a big deal? 🤷 

https://github.com/user-attachments/assets/af3bc8c0-058c-4b68-beb8-53ea42ca1210

Azmoria: Yeah this is pretty much what I meant, looks good 👍 

In another PR I'll look at updating `display_note()` to update the note rather then close/reopen if it already exists. That was a recent change anyway so we didn't have duplicates of notes opening.

Issue #1849, opened 2024-01-29
Request: Allow the DM to 'speak' as a selected token


Azmoria:
Give the DM the ability able to 'speak' as a monster/token in chat.

PR #2932, opened 2025-02-23
If the DM has one or more tokens selected, speak as those tokens;

Drenam1:
If the DM has one or more tokens selected, speak as those tokens;
Added title attribute to the character name span in the chat bar so that long titles (such as long character names or, more relevant to this PR, multiple character names can all be seen on hover.

#1849 

In ChatObserver, the sendChatMessage function checks whether the current user is the DM and if they have any tokens selected. 
- If the current user is a player, the message sends as that players character.
- If the current user is the DM with no tokens selected, the message sends from 'THE DM'.
- If the current user is the DM with one or more tokens selected (PC or NPC), the message sends as if from those characters.

Considerations
1. Should the DM be able to send chat messages as if from another player character? Currently with these changes, they are able to, but it should be easy to remove this ability.
2. When sending a message on behalf of multiple tokens, they are currently ordered in the chat panel in the order they were added to the scene. Should this be a different order (such as alphabetic)?

image

### DIFF
--- a/ChatObserver.js
+++ b/ChatObserver.js
@@ -74,9 +74,17 @@ class ChatObserver {
     }
 
     async #sendChatMessage(text) {
+        let player = window.PLAYER_NAME;
+        let image = window.PLAYER_IMG;
+        if(window.DM && window.CURRENTLY_SELECTED_TOKENS.length > 0) {
+            let id = window.CURRENTLY_SELECTED_TOKENS[0];
+            let firstToken = window.TOKEN_OBJECTS[id];
+            image = firstToken.options.imgsrc;
+            player = window.CURRENTLY_SELECTED_TOKENS.map(id => window.TOKEN_OBJECTS[id].options.name).join(", ");
+        }
         let data = {
-            player: window.PLAYER_NAME,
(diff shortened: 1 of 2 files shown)


Azmoria:
This can be a separate PR but I think it'd be nice for players who have multiple PCs in a game or sidekick npcs to be able to speak as them as well. So maybe if they have control via `player accessible stats` (for monster tokens) or shared vision (for either monster or pc) allow them to speak as those?

I can merge this now though as it seems ready and we can do that separately if you'd like.  

From considerations

1. I think it's fine for the DM to send as other PCs, some DMs use PC sheets to build NPCs.

2. I think groups speaking isn't going to be super common other than maybe cultist type chants or crowds shouting. Not sure order matters much in those cases. I think it's fine for now - maybe if we make adjustments later we can look at doing something else after some feedback.

Drenam1:
Also happy to merge as-is and do player side of things later 😄 

No worries, half the things I've done the last couple of days were things I wanted changed in the first place, so I'm being selfish and doing them first haha. Plus I've been looking for a good side-project for my off hours, and I really love AboveVTT so it all works out great.

Issue #2936, opened 2025-02-23
Enhancement - journal search


Drenam1:
I think it'd be cool to have a text input up the top of the Journal (either above or below the Add Folder functionality) where, if it has text content, it filters the journal to show only those notes/folders that have that specific string in the title. Maybe even search the content of the note itself for that specific string, although that might be overkill.

This already basically exists in the Scenes tab - chances are it can be replicated relatively easily to the journal tab.

![Image](https://github.com/user-attachments/assets/39696ebe-039a-45c2-8a03-dea5475b74ea)

Azmoria:
Great job with this your flying through these. Thanks again 🥳

PR #2941, opened 2025-02-24
Enhancement - journal search

Drenam1:
#2936 
The idea here was to do the following:
1. Get a list of notes that contain the searchText, save that to array 'relevantNotes'.
2. Get a list of folders that contain the searchText, save that to array 'relevantChapters'.
3. Traverse the structure up to the root to get all parent folders, add those to array 'relevantChapters'.
4. Traverse the structure down as far as possible for all **folders that were directly searched (see Mirkwood example below)** and add those folders and their notes to the relevant arrays.
5. When rendering a folder/note, it first checks if that folder/note is in the relevantNotes/relevantChapters array. If it is, it renders - if it is not, it doesn't.

This was slightly more complex than initially anticipated due to the way the Journal was initially set up being different to the way I thought it would be, but I feel that the changes made provide the desired functionality. Currently the search only searches the note/folder titles, but it could be easily extended to search contents as well.

image

### DIFF
--- a/Journal.js
+++ b/Journal.js
@@ -216,6 +216,7 @@ class JournalManager{
 		console.log('build_journal');
 		let self=this;
 
+		// Clear all elements from journal panel except the searchbar, which needs to stay in place between searches
 		journalPanel.body.children().not('[name="journal-search"]').remove();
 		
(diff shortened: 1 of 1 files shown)


Azmoria:
You blasted through that, all seems to work as expect in testing. Great job 👍 

Closes #2941

Issue #2942, opened 2025-02-25
Enhancement - expand/collapse all journal folders


Drenam1:
Just something I thought while working on journal search, a couple of buttons that expand and collapse all the journal folders. I'll do this tomorrow - thinking I'll make the journal search bar narrower slightly and put the buttons to the right of it 👍

PR #2945, opened 2025-02-25
Enhancement - expand/collapse all journal folders

Drenam1:
Added expand/collapse buttons;
Added new icon assets;
Updated styling so new buttons fit;

#2942 

image


Journal.js
```diff
@@ -235,6 +235,21 @@ class JournalManager{
 			searchText = searchElement?.value || '';
 		}
 
+		let journalControlContainer = $(`<div id="journal-control-container"></div>`);
+		let expandAllButton = $(`<button class="expand-all-button" title="Expand all folders"><img title="Expand all folders" alt="expand" height="20px" src="${window.EXTENSION_PATH + "assets/icons/expand.svg"}"/></button>`);
```
Azmoria, line 239: For consistency sake maybe we could use the same expand/collapse icons as on the token/scenes panels.

Wouldn't need the new svg files this way either since it just uses google icons.

Expand
```
<button class="expand-all-button token-row-button expand-collapse-button" title="Expand All Folders" style=""><span class="material-icons">expand</span></button>
```

Collapse
```
<button class="collapse-all-button token-row-button expand-collapse-button" title="Collapse All Folders" style=""><span class="material-icons">vertical_align_center</span></button>
```

Css adjustment
```
#journal-control-container {
    display: flex;
    gap: 5px;
    button {
        width: 30px;
        height: 30px;
    }
    .material-icons{
        padding: 1px 0px 0px 0px;
    }
}
```

Azmoria (commented):
Works as expected 👍 - just a comment on consistency of icons to match the other panels.

Journal.js
Drenam1: Nice, I didn't even realise that the scenes had expand/collapse buttons 🤣 updated the code with your changes and removed the two SVG assets I added, thanks!

PR #2974, opened 2025-03-06
Feature: numpad keys setting and implementation

jeffsenn:
Setting for adding extra keys for driving UI with num pad and rolling dice

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -23,9 +23,96 @@ function unhide_interface() {
     }
 }
 
+extras_bound = [];
+function bind_extras() {
+    //cycle between modes of showing combat&log
+    function mybind(a,b) {
+        Mousetrap.bind(a,b);
+        extras_bound.push(a);
+    }
+    cyclelog = function () {
+        setTimeout(() => $('#combat_button').click(), 20);
+        if ($('#combat_tracker_inside').is(':visible')) {
+            //make sure gamelog selected
+            change_sidbar_tab($('#switch_gamelog'), true);
+            $('#hide_rightpanel').click();
+        }
+    };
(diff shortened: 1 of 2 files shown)


Azmoria:
I don't think this needs to be a setting, I also realized it conflicts with the menu number keys but I don't think those hotkeys are really necessary or all that useful anymore since many tool menus have too many buttons. 

I find not having any feedback in this when a button is pressed and auto rolling makes this hard to use tbh. It feels a bit like it lags and I end up hitting too many keys often, or sometimes too few if I'm to slow.

I think it should be more along the lines of this 
https://github.com/Azmoria/AboveVTT/commit/543b60be9e01ca068e2aac908de35b88cd354f99

This also allows you to use + or - after having dice added to the pool to add modifiers. It does limit it so there isn't adv or dis at the moment - but could probably be added to 8 & 9. I just ordered it in the order it shows up in the dice pools (both ddb and ours below chat). Enter then rolls when ready.

I also adjusted the center on current token to just click the button in case that ever changes we don't have to modify it in two spots. It also uses ctrl/control + space to center since shift+space already centers on player. I think it falls more in line with left hand on keyboard binds we tend to have and is just a different mod on an already existing key.

Ignore the ctrl-> mod stuff for other binds that's just to try and have windows and mac play well together without seperate binds all the time, only found this now while looking at this.

jeffsenn:
> I think it should be more along the lines of this [Azmoria@543b60b](https://github.com/Azmoria/AboveVTT/commit/543b60be9e01ca068e2aac908de35b88cd354f99)

Agree... I like this direction better than my quick hack.  I can adapt my setup.  It's essentially compatible with my setup if there was an "auto roll after" timeout....I'll think about that, but hitting the enter key is just fine too.

PR #2975, opened 2025-03-06
Option for hiding scrollbar and during observer mode

jeffsenn:
Style scrollbar for hiding when setting is selected and also during shift-H hide screen.
(Note only brief testing... given that scrollbar seems to act differently on different platforms/browsers)


Azmoria:
Seems to work on my end too just fine in chrome and firefox👍.

PR #3003, opened 2025-03-13
Most Mac users would prefer control here

jeffsenn:
Most Mac users have Cmd-Space bound to Spotlight globally, so I think it's preferable to have Control-space be the binding for find token (on both mac and windows)

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -206,7 +206,7 @@ Mousetrap.bind('enter', function () {       //zoom minus
     $('.roll-mod-container>.roll-button').click(); 
 });
 
-Mousetrap.bind('mod+space', function (e) {    
+Mousetrap.bind('ctrl+space', function (e) {    
     e.preventDefault();
     $('#combat_area tr[data-current=1] .findTokenCombatButton').click();
 });


Azmoria:
image

this is what mod is supposed to do already - if you just put ctrl it wont work on mac from my understanding

jeffsenn:
Huh?  It's just that one key (cmd-space) that is problematic...

Azmoria:
all our bindings are supposed to be control from before I removed them

Azmoria:
I was wrong it was command - sorry I don't know mac buttons well at all lol. I assumed ctrl = control = command. I even listed the space+control hotkey

image

jeffsenn:
So "mod" means "control" on windows and "command" on mac - it's a convenience for just this convention.

jeffsenn:
Oh - do you want me to fix the HELP text too?

Azmoria:
No this is fine. The help text is already correct for this. I _thought_ it was already bound to control via mod.

Azmoria:
I only see it wrong on the numbers - mac binds aren't listed for anything else. CTRL doesn't not mean control in that listing.

PR #3006, opened 2025-03-13
help menu improved for mac

jeffsenn:
afaict this will now be correct for Mac if you want menu to be platform specific

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -39,9 +39,9 @@ function isIOS() {
 }
 function getModKey() {
   if (navigator.userAgentData) {
-    return navigator.userAgentData.platform.includes("mac") ? "CMD" : "CTRL";
+    return navigator.userAgentData.platform.includes("mac") ? "&#8984;" : "CTRL";
   } else {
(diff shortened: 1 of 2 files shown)


Azmoria:
looks good 👍  - except there's no mention of Ctrl = control in the case of the ctrl+space. 

Which yes I know it seems redundant to me too, but occasionally people don't get that for whatever reason. Also may be confusing as old binds anything labeled CTRL was cmd/meta.

jeffsenn:
I pushed a small change to use &#8984; instead of "CMD" on Mac.  This will be super expected for mac users (Mac menu system uses this convention exclusively) and reduce any confusion there.

jeffsenn:
Changes intended to only functionally affect Mac - this is sort of what people expect and matches what the native menu system does for modifier keys: 
Screenshot 2025-03-13 at 3 47 35 PM
Screenshot 2025-03-13 at 3 47 03 PM

Azmoria:
Yeah makes sense  - definitely seens those on mac keybind list. Didnt know what the option or control ones meant until now lol but this works. 👍

PR #3012, opened 2025-03-14
AOE origin at single selected token

jeffsenn:
(NO HURRY!)

How do we feel about this feature:
  Originate a new AOE at a token instead of middle of view  (only if a single token is selected).
  This make it easy to select your wizard, cast the spell (add AOE on the caster) and then move the AOE to range while seeing the ruler and area.

PR is just a start - would love it if cone actually worked too... but some fair bit of geometry not worth it if no one likes this.

https://www.dropbox.com/scl/fi/z4gt3pow5jtp7j2eqaa0b/AOE-on-Selected.mov?rlkey=8pg3qyqlbefmjzz8icobbzfbs&st=g2syex0h&dl=0

### DIFF
--- a/AOETemplates.js
+++ b/AOETemplates.js
@@ -116,7 +116,17 @@ function setup_aoe_button(buttons) {
         const style = $("#aoe_styles").val().toLowerCase()
         const options = build_aoe_token_options(style, shape, size)
 
-        place_aoe_token_in_centre(options)
+        //if single token selected, place there:
+        if(window.CURRENTLY_SELECTED_TOKENS.length == 1) {
+            tok = window.TOKEN_OBJECTS[window.CURRENTLY_SELECTED_TOKENS[0]];
+            const half = parseFloat(window.CURRENT_SCENE_DATA.hpps) * tok.options.gridSquares/2;
+            const x = parseInt(tok.options.left.slice(0,-2)) + half;
+            const y = parseInt(tok.options.top.slice(0,-2)) + half;
+            console.log("PLACEAT",tok, half, x, y);
+            place_token_at_map_point(options, x, y);
+        } else {
+            place_aoe_token_in_centre(options);
+        }
         $('#select-button').click();
 
     });
--- a/AOETemplates.js
+++ b/AOETemplates.js
@@ -118,18 +118,23 @@ function setup_aoe_button(buttons) {
 
         //if single token selected, place there:
         if(window.CURRENTLY_SELECTED_TOKENS.length == 1) {
-            tok = window.TOKEN_OBJECTS[window.CURRENTLY_SELECTED_TOKENS[0]];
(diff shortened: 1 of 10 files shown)


jeffsenn:
Pre-existing (small) issues that this PR now makes more obvious:
 - AOE line token: token adjustment -> custom size,  just works incorrectly
 - AOE cone/line token adj->custom size: should scale from top-middle origin
 - AOE circle/square token adj->custom size; should scale from center origin
 - ruler for AOE line token when moving it is ... odd at least, probably broken.
 - ruler for AOE cone token movement is technically correct on measurement, but visually odd.

I suppose these could be fixed by giving these tokens a "visual origin offset" and a lot of little code adjustments (bigger task)

Azmoria:
For line it seems to set correctly on my end for custom size - entered 25 ft and measure the line it's 25 ft long, 50ft -> 50ft etc. It doesn't seem to display the current size correctly in the input when opened but adjusting it works as expected.

Scaling from origin for cone/line might be nice. I'm not sure how often people are adjusting these rather then just dropping the size they need. I wouldn't expect often so probably low priority. 

Rulers are all measuring correct length for all from what I can see. Just it's from the center of token. 

What are you trying to measure when positioning an aoe? 

I would assume just positioning an aoe not measuring anything when dragging it - except for some rare ones that have movement, in which case where the ruler being in the center is fine as long as measurements are correct. 

Ruler being at origin would also be fine but might not be worth figuring out the rotation and position of origin is all. Probably a good chunk of work for minimal benefit that I can see - if you'd like to do it though that's fine.

I'll add it these to the issue list.

jeffsenn:
Yeah - I don't suggest we fix these now (or maybe even soon) - as you say: complicated.  And scaling them... yeah prob never a priority - just make a new one.

I think the main use case for measuring is plopping a spell AOE on the caster and then moving the AOE token to measure how far away (and/or at what angle) it can originate (eg fireball etc).

Oh - also now it occurs to me.... are we "spells originate at centers of squares" or "spells originate on grid crossings" DMs?  I'm not sure the official rule has ever been clarified (I'm the former).

Azmoria:
I guess I don't understand - the 30ft aoe visual shows you where the fireball is. Isn't the the point, not needing to measure?

As for grid snapping it can be either, aoes snap at 1/2 grid so center or lines for these reason.

Azmoria:
Oh nvm you're talking about ranged spells - I guess I measure before placing them lol

Azmoria:
Put up issues #3014 and #3015 for these. The bug one I'll probably look at fixing before release if possible. 

For the point of origin issues while typing this up I realized Cube is going to be difficult or impossible to accommodate fully since it's origin is anywhere on any face.

Azmoria:
I think #3017 will address the drag ruler part to help with measuring cast distances where needed. 

I decided to put square at the edge as if the edge point of origin is range no matter the orientation of the cube it will be in range (or close enough)

PR #3260, opened 2025-04-17
Assign journal notes to a scene

Drenam1:
A little something I decided I wanted :P Attaching a single note to a given scene so that its easily accessed from one location.
I just noticed that I haven't given a way to un-attach a note from a scene without replacing it with another note. I'll do that tomorrow, so please don't approve this PR :)

https://github.com/user-attachments/assets/bc93ab90-95a7-44b8-b9e7-750d06d96a28


Azmoria:
I think this fits more in the right click menu 'Create Scene Note' if no note already -> 'Open scene note', 'Delete Scene Note', 'Edit Scene Note'. 

Keeping it separate from normal journal notes similar to token notes. 

You don't have to create the note then go through a selection process clicking through all the folders to add it to your scene. Also doesn't clutter up the journal folder for scene specific notes, editing it or deleting it I think would be slightly less of a chore this way too. 

Deleting the scene would have to also delete the note in this case.

Not against the option I'm just not sure there's an advantage over a hidden token on the scene for this. As generally scene notes would only be used when on the scene. 

One down side is if you ever closed it and had to reopen it you're still going to another tab vs a token note where it stays available on screen to double click to open at any time.

Azmoria:
I was thinking about it more a couple other things to consider

Exporting scenes/scene folders will have to include the notes of the scene(s)

Duplicating scenes to match expected functionality of being able to edit it without affecting the original will have to create a copy.

Both of these are probably easier if we go the same route as token notes where it has the same id as the scene and isn't in the journal folder structure.

Drenam1:
Yup fair enough. I'll change it to go that route sometime today!

PR #3506, opened 2025-08-19
Create scene notes

Drenam1:
The idea here is to create a note that applies to a given scene.
It does this by 
1. When the 'Open Scene Note' option is selected, create a hidden note within the journals and open it for editing.
2. User can edit the note as normal.
3. When the user closes the note, the note data is saved to the scene and the hidden note is deleted.
4. When the user selects the 'Open Scene Note' option again, the saved note data is used to re-generate the hidden note and open it for editing.

I did it this way to reuse as much of the existing note infrastructure as possible. Also, since the note data is saved to the scene it should be exported with the rest of the scene data if needed.

Possible next step - if a scene has note data, add an icon on the scene that opens the note without having to use the dropdown menu.

https://github.com/user-attachments/assets/435c56df-9099-49b9-b364-e8cd37a2cb22

### DIFF
--- a/Journal.js
+++ b/Journal.js
@@ -2276,8 +2276,6 @@ class JournalManager{
 		$("textarea[data-note-id]").each(function(){
 			let taid=$(this).attr('id')
 			tinyMCE.get(taid)?.execCommand('mceSave');
-			console.log('Closing notes');
-			console.log(window.JOURNAL);
 			const self = window.JOURNAL;
 			if (self.notes && onSaveCallback) {
 				Object.keys(self.notes).forEach(noteId => {
@@ -2287,7 +2285,6 @@ class JournalManager{
 					}
 				});
 			}
-			console.log(window.JOURNAL);
 			$(this).closest(".note")?.dialog("close");
 		});
 	}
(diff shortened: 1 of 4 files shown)


Azmoria:
Awesome, I'm not sure how much data will end up on the server and I don't want to break the bank for cyruzzo so we probably have to slightly modify this but easy enough to do from this point. Mostly an issue of storage space costs I think.

Instead of saving it to the scene, having it save to the journal with the same ID as the scene. Then on openning the scene note just look for the note id that matchs the scene id.

This is how token notes work too - they look to see if a note exists with the same id as the token. These notes aren't displayed in the normal journal sidebar. It also saves us from having to regrab it from scene data each time we open.

Alternatively just save the note id to scene data on the server if you think it's easier with a seperate id (this would be more like how we save custom stat blocks).

Either option we can add a quick line to make sure scene notes are captured the same way custom statblocks/token notes are when exporting a scene too. 
 
If you'd like I can make these adjustments pretty quickly if you think this is ready for merge otherwise or if you'd like to adjust it or tinker more thats good to. I can probably squeeze it in before the upcoming release.

 
 Thanks for doing this 🍻

Drenam1:
Yeah those changes sound good if you want to make them yourself!

PR #3642, opened 2025-10-15
Fix syntax error: remove double comma in quickToggleDefaults options

andrewklingelhofer:
## Summary
Fixed a syntax error in Settings.js at line 645 where a double comma creates an empty array element in the `quickToggleDefaults` options array.

## Issue
The double comma (`], },,`) creates an empty array slot between the `centerPing` and `selectLocked` option objects.

## Changes
- Settings.js:645 - Changed `], },,` to `], },`

## History
This syntax error has existed since the feature was added in March 2025 (commit 8d1933d3).


Azmoria:
Thanks for catching  that too 🍻

PR #3752, opened 2025-11-21
Enhancement: when token added enhance character sheet data if stats has sheet link

jeffsenn:
When a "my token" has a character sheet reference in the stat block, this will use the DDB api to enhance the token with the current character stats/hp/ac.  

There are a couple of todos (but probably could be merged since this is a pretty marginal feature) 

- the logic to extract from the character data (see: extract_character_data in CoreFunctions) is somewhat complicated - but can be incrementally improved.  One know shortcoming is I have not yet figured out how to apply the "choose your ability" modifiers.

- the logic here is to only replace token options if they have not been overridden in the custom stat block itself (which was my prior approach: just cut-n-paste from character to customstat block -- that is not very dynamic however).
BUT: customStats will be replaced if found in the character.... could improve that logic.


Azmoria:
I like this idea for presetting the base stats. 👍 

Edit2: So you're using `fetchCharacter` when I think `fetchCharacterDetails` avoids you having to manually process it so much. Note all the stuff we don't want it auto stripped from this when a token is placed so you should just be ablet merge this with the customization options already set without an issue.

For example:


image

jeffsenn:
There is also a slight corner case concern here: if your character link in the statblock is bad, then this will pop the "report errors" stuff... I will fix that if you want - but it involves rearranging things a bit in DDB api stuff... let me know if you want me to do that.

Azmoria:
I can deal with any error messaging needed that's not a problem. I'll just change the text to alert about an incorrect url in the statblock or something.

jeffsenn:
Oh characterDetails is much better - I'll correct.

PR #3788, opened 2025-12-12
Dm Screen Implementation

Drenam1:
I implemented a basic DM screen. It has the following features:

- A button added to the top right that opens/closes the DM screen. I made this available to the DM only, but this could be changed.
image

- Pressing escape while the DM screen is open will dismiss it.
- I have added 5 'sections' currently - "Conditions", "Improvising Objects and Damage", "Actions", "Skills and Mechanics", and "Travel". These sections are based off the official DnD 5e DM screens (2024 ruleset).

image

- It should be simple to add additional sections as needed, or move shift things.
- Dice rolling is enabled 

image

### DIFF
--- a/DMScreen.js
+++ b/DMScreen.js
@@ -228,6 +228,9 @@ function buildSkillsAndMechanicsBlock() {
                 </thead>
                 <tbody></tbody>
             </table>
+            <div class='dmScreenChunkDefinition'>
+                As an example, say 10 bandits (+3 to hit) are attacking a player character with 18 AC. The bandits need to roll a 15 or higher to hit. According to the table, 4 attackers are needed for one hit. Since there are 10 bandits, you can expect two hits on average. From there, roll damage as normal.
+            </div>
         </div>
     `);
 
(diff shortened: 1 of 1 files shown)


natemoonlife:
This is rad! Great work!

Azmoria:
Very nice. I think it'd be nice to be able to create a custom page - probably easiest way to do this would be creating notes (like token notes) that can just be a custom page. 

Also just for info, a good chunk of rules data is preformatted into html in window.ddbConfigJson - we pull this from DDB directly and store it. Might have to make sure it's actually initialized before setting up the dm screen but could save having to hard code some things or update them if/when DDB makes updates. This is where we pull condition data from for example.

I'll give this a good test/review in the next day or two and merge if all is good.

image

Drenam1:
Good shout @Azmoria - I didn't realize that was already existing functionality! I've changed the implementation to use window.ddbConfigJson where possible. It kind of blew out the the size of some of the pages (like the Conditions page, with all the legacy definitions), but I don't hate that too much - you can still scroll after all 😄

Azmoria:
Looks good - merged and in beta now.

Thanks for doing that. I'll probably look at at minimum a popout option if not also custom pages before a release. Assuming one isn't forced before the holidays by a change on DDB.

PR #3797, opened 2025-12-21
Additional dm screen blocks

Drenam1:
Added additional DM screen blocks and reorganized the dropdown into alphabetical order.

Bastions
image
Name Improv
image
Services
image
Spellcasting
image
Weapons
image

### DIFF
--- a/DMScreen.js
+++ b/DMScreen.js
@@ -1,13 +1,23 @@
-console.log("DMScreen.js is loading");
-
 /**
- * Initialize the DM_SCREEN array for the current campaign
+ * Gets all owned sourcebooks then builds the DM Screen
  * Should be called on startup
  */
 
 function initDmScreen() {
+    window.ScenesHandler.build_adventures(function () {
+        buildDMScreen();
+    });
+}
+
+/**
+ * Builds the DM Screen UI
+ */
+
+function buildDMScreen() {
     console.log("initDmScreen called");
-    console.log(window.ddbConfigJson);
+    console.log(window.ScenesHandler.sources);
+    // Check if the DnD 2024 DMG is owned
+    const dmg2024Owned = window.ScenesHandler.sources["dnd/dmg-2024"] !== undefined;
 
     let cont = $(`<div id='dmScreenContainer'>
             <div class='dmScreenheader'>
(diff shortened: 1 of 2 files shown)


Azmoria:
I'll look to merging this when I'm back from holidays. 🍻 

I was thinking more about this, some of the stuff is from the DMG like bastions in this and I think mob attacks form the previous one. Since not everyone has access to that stuff I'm not sure how much we should include. 

A roll table _seems_ fine in my mind but not sure if wotc would see it giving out content people don't own. Content outside basic rules may belong more in custom pages users can add themselves. 

Maybe we can use the same logic we use for source links people have access to for journals and import that if they do have access. Example video for how we add tooltips to sources in journals - maybe we can try to load a tooltip link from the DMG and if it works then insert all the DMG content into the DM Screen since they have access.

Third option make them importables like journal chapters/scenes that check for access on add.
 

https://github.com/user-attachments/assets/e57b67d4-6a24-4311-bbe4-b552838cd3d8

Drenam1:
@Azmoria Enjoy your holiday - this can wait until you're back! 🍻 

You make an excellent point. I hadn't considered things that should be paywalled behind owning the DMG.

I've gone through each of the pages I've built to see what is in the Free 'Basic Rules' book and what isnt.

In Basic Rules
- All actions (under Rules Glossary)
- All conditions (under Rules Glossary)
- Object Improvisation (under Rules Glossary > Breaking Objects)
- Lifestyle expenses and Services (under Equipment > Services)
- All skills (under Playing the Game > Proficiency > Skill Proficiencies)
- All light/visibility/cover rules (under Rules Glossary)
- Long/high jumping (under Rules Glossary)
- Concentration (under Rules Glossary)
- Spellcasting rules (under Spells)
- Travel Pace (under Playing the Game > Exploration > Travel)
- All weapons, weapon properties and weapon masteries (under Equipment > Weapons)

Not in Basic Rules (or at least, I couldn't find them)
- Bastions
- Damage improvisation
- Rollable name tables
- Mob rules
- Journey stages + travel terrains + weather

Of the items not in Basic Rules, I would think:
- Bastions, mob rules, travel rules - should only be displayed if the DM owns the DMG
- Damage improvisation, rollable names - could probably be public-facing?

What do you think? Would you prefer if all the 'Not in Basic Rules' items were hidden, just in case? I've made a change to the PR to proof-of-concept it, and I tested it with a brand new DDB account that owns none of the sourcebooks - it successfully hid Bastions and Mob Rules.

Azmoria:
Yeah, Names and Damage improvisation tables sound ok. We should probably hide all of the other content outside basic rules. Maybe even consider loading it directly from the books - for now leave it hardcoded in but hiding it is fine.

 I want to look at a few things myself before release anyway so no rush we have time. I'd like to put this in our draggable windows and get a popout setup for it at minimum I think. 
 
 I'll be on the road most of the day today but I should be back to looking at more stuff in the next few.

Azmoria:
Merging this for now - will make adjustments as needed

PR #3799, opened 2025-12-22
allow build_adventures to enumerate all sources

jeffsenn:
build_adventures(...) recreates iframe after source click and query selector ("[class*='SourceCard_sourceTitle']" will only enumerate unfiltered sources. So if user selected something that does not appear in "owned+shared" resources then the link won't be found.  Changing iframe url allows all sources to be enumerated.

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -629,7 +629,6 @@ class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 	}
 
 	build_chapters(keyword, callback) {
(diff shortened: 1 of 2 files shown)


Azmoria:
So this does display unowned sources in the journal importer as I thought it might. This isn't a huge deal as it did this before the change anyway. Once DDB fixes it we can go back to just checking owned/shared so unowned sources just never populate to begin with.

Once fixed I'm eventually going to hide the `all sources` tab/filter as they won't be useful in AboveVTT anyway. For now it's a good work around. 🍻

PR #3852, opened 2026-01-19
Add token random name generator button

Drenam1:
As per this feature request - I thought it was a good idea.
https://discord.com/channels/815028457851191326/815028716040224818/1462091948797198569

If you have a single token selected, then it generates a random first and last name and updates that token
https://github.com/user-attachments/assets/a46f2500-8995-4505-8540-409dec5f6ecc

If you have multiple tokens selected, then it generates a random first and last name for each token, updates each token individually, then puts "Idividual names generated" into the input field
https://github.com/user-attachments/assets/5d525c3a-f5e3-47e9-a402-b4973f7e6b75

The names list is from a personal project of my own. It could be extended to include more exotic names (like stereotypical dwarf/halfling/gnome/etc names) but this is priamrily fantasy-human-based.


Azmoria:
lol was looking at the DDB fantasy name generator calls in the character builder yesterday after seeing that request. I don't think its really worth using DDB name calls this feels easier and honestly better names lol.

Thanks for doing this🍻

PR #3855, opened 2026-01-20
QoL - Player combat tracker setting for next turn (on deck) indicator

Drenam1:
Setting to enable/disable the next turn indicator, disabled by default. Only available to players, not DMs (I figure DMs don't need the reminders its their turn 9 rounds out of 10 🤣)

image

When the next/previous button is pressed, it checks who is up next in the initiative order, checks who the player is and sends a new messagebroker message "nextTurnIndicator", which triggers an audio cue and a visual message if enabled.
Notification sound is [Positive Notification Alert](https://pixabay.com/sound-effects/film-special-effects-positive-notification-alert-351299/) from Pixabay. I think everything on that site is royalty-free. I don't love the sound, but it was the best I could find  😆

https://github.com/user-attachments/assets/3022dba3-79ae-4b01-a440-e428a21c6c90

### DIFF
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -778,11 +778,6 @@ class MessageBroker {
 			if (msg.eventType == "custom/myVTT/pointer") {
 				set_pointer(msg.data,(!msg.data.dm || (msg.data.dm && !msg.data.center_on_ping)));
 			}
-			
-			if (msg.eventType == "custom/myVTT/nextTurnIndicator") {
-				
-				
-			}
 
 			if (msg.eventType == "custom/myVTT/lock") {
 				if (window.DM)


Azmoria:
Looks good 👍 

A couple thoughts,

We already send a message any time the combat tracker is updated, we should probably piggy back on that instead of it's own message. `window.MB.sendMessage("custom/myVTT/CT",data);` in `ct_persist()` is where that happens. 1 to reduce messages and 2 this will catch if the turn order changes via new initiative rolls, if the DM hits prev, or other combat tracker changes. Alternatively just process the next turn on the player end when they receive the combat tracker data.

The setting itself should probably be part of the combat tracker settings rather then main settings tab.
image

I can take a look at moving some of this around if you don't have the time.

Drenam1:
Changes made based on your suggestions

- Removed global setting;
- Added setting to combat tracker settings;
image

- All logic now handled on players end;

PR #3901, opened 2026-02-03
WIP: Canvas opt grid

jeffsenn:
Don't merge!  MANY features not implemented!

Just for demo of CSS only grid.  Square grids only.  CSS won't hide so you can compare to canvas.

See what you think

### DIFF
--- a/Journal.js
+++ b/Journal.js
@@ -1612,6 +1612,74 @@ class JournalManager{
 		    });
 		}
 	}
+
+	addTrackedInputs(target, id = {noteId: undefined, token: undefined}){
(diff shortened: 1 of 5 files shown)


Azmoria:
Looks great, like the scaling you've got with zoom. Could be even slightly less if you prefer a little less chunky lines or cap at a max size. Unless this is what --hide-when is already kinda doing

Looks like some good variable optimizations you got in there too. Something I _should_ be doing when working in an area 😅.

jeffsenn:
closed in favor of "refactor-grid"

PR #3918, opened 2026-02-08
Performance - fix issue with doubling up on offscreen canvas increasing memory usage and reduce draws to light canvas

Azmoria:


Azmoria:
relabeled internal so as not to go in release notes as I had to go back and rework this

PR #3940, opened 2026-02-12
use mod for rotation; include scene scroll change in calculation

jeffsenn:
2 fixes: 
  - mod target degrees rotation so that odd "across the limit" jump does not happen
  - take into account any amount the scene has scrolled during mouse drag in rotation draggable

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -5185,7 +5185,8 @@ async function do_draw_selected_token_bounding_box() {
 
 			let centerPointRotateOrigin = {
 				x: 0,
-				y: 0
+				y: 0,
+				original_scene_bounding: []
 			};
 			
 			let angle;
@@ -5197,6 +5198,9 @@ async function do_draw_selected_token_bounding_box() {
 					self.orig_top = grabberTop;
 					self.orig_left = grabberLeft;
 
+					//in case the scene autoscrolls while rotating
+					centerPointRotateOrigin.original_scene_bounding = document.getElementById('VTT').getBoundingClientRect();
+					
 					let furthest_coord = {}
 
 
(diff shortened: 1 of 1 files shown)


jeffsenn:
Not sure I explained well: original problem happens when you zoom in close enough, grab the rotation handle and then move the mouse such the scene "auto scrolls" around while you are trying to rotate

jeffsenn:
Here (in released) : If I grab the handle - notice ship rotates with starboard side toward cursor - but as soon as screen scolls, its the back of ship or the other side toward cursor.

https://github.com/user-attachments/assets/8408286c-64d3-4524-bda3-1513c315ad22

Does that not happen for you?

Azmoria:
Ignore that last comment about the scroll it does this on main too just seemed faster for some reason - maybe at a differnet zoom level then I'd normally rotate stuff at.

I do feel like this is bugged though. When you move the screen the grabber stops being on your cursor which breaks the 'look here' functionality. This appears to be on main too.

I'm not sure what this is trying to solve exactly as I dont see any different behaviour zoomed in or out.

Edit: Ah yes it's the same bug I'm describing above but that happens in your version too because the grabber isn't on the mouse anymore. 

Edit 2: It may just be because you only changed the 'look here' rotator and I've been trying to group rotator which still has the bug 🤦‍♂️

Azmoria:
Ok yeah, we can apply the same logic to around ~line 5266 and I think it'll fix both. Sorry for being a dunce with that one lol

jeffsenn:
Oh - yes!  Here is the other problem. It has an odd flip over behavior when you cross the x origin line with the mouse (sometimes - depends on rotation).  The degrees % 360

Uploading test2.mov…

 fixes that.
Sorry I missed that other draggable before submitting :-(

jeffsenn:
did that video upload? try again: 

https://github.com/user-attachments/assets/856cb518-910d-4842-838b-7c5e6dd5b318

Azmoria:
Struggling to replicate that second part but I'll assume it's fixed. Not having that jump in the animation is really nice though. I'll merge this in shortly. Thanks for doing those/catching that

It looks like it might just be the animation due to going above a full rotation as it looks like it ends up in the correct spot which makes sense why your % 360 fixes it.

The same side of the ship (down in the original image position) is facing the eye. Generally the look here being used for top down tokens that for most artists they have facing down to start. Example from a discord message of it's original use case and why group rotate is a seperate option.


https://github.com/user-attachments/assets/7202c6ac-754d-4d6e-8e60-6313aafe4c6a

PR #3950, opened 2026-02-15
Add Flip Option for Token Images

jeffsenn:
- Added horizontal/vertical flip options for token images
- Added Key Handler | (pipe) to cycle flip on selected tokens

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -39,9 +39,8 @@ const availableToAoe = [
 ];
 
 //reused transform definition
-const imageTransform = 'scale(var(--token-scale)) rotate(var(--token-rotation)) scaleX(var(--token-flip-x, 1)) scaleY(var(--token-flip-y, 1))';
+const imageTransform = 'scale(var(--token-scale)) rotate(var(--token-rotation)) scaleX(var(--token-flip-x, 1))';
 function tokenFlipX(token) { return ((token.options.tokenFlip || 0) & 1) ? -1 : 1; }
-function tokenFlipY(token) { return ((token.options.tokenFlip || 0) & 2) ? -1 : 1; }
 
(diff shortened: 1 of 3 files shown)


Azmoria:
Looks good and functions as expected.

Question though - do we need multiple flips? Isn't flipping both horizontal and vertical the same as rotating it with no flips or shift+up. 

If there isn't another use case I'm not thinking of could this just be a toggle to mirror the token and just do horizontal? Then rotation would cover the rest?

jeffsenn:
It was just as easy to do both.  I guess if we only want one, horizontal is probably most useful. What do you prefer?

Azmoria:
I would probably just want horizontal if we do one yeh.

I think from a stand point of useablility it makes more intuitive sense to me to just be 1 mirror too. Going through the extra steps with a hotkey is a bit confusing when I don't know what stage of flipped they are tbh. At least when rotation is also involved it gets confusing to keep track of.

A toggle is also a bit faster then 4 options in a drop down if they aren't needed for something else.

jeffsenn:
Yep - I'll change it to horizontal only

Azmoria:
One use case I have found that vertical does do is invert the 'look here' function since we make the bottom of the token face the eye. 

Not sure how useful this would be for top down tokens, I feel _most_ are designed facing down as VTTs started defaulting that (we do to).

This might be better as a seperate feature to choose which side of the token is the direction it's facing but I've not heard the request for this before so it might not really be needed/worth doing.

https://github.com/user-attachments/assets/efa0a39a-0758-4846-a61a-51f3e53433c3

jeffsenn:
I simplified to horizontal - I'm happy with either. As you point out the vertical is useful for the purpose you suggest.... but I think your "declare the side" is a better (if future) feature.

PR #3995, opened 2026-02-25
Fix the mouse drop point for token placement

jeffsenn:
Set ignore offset so mouse drop of token from panel hits correct point.


jeffsenn:
This only is a problem on scenes that have .offsetx != 0 or .offsety != 0

jeffsenn:
It may be that ignoreOffset should just be true in all cases in create_and_place_token - I don't know what the intent is for all of the "place in center or near center of map" cases - so I'll let someone else decide

PR #4002, opened 2026-02-26
Token Drag Select - improved hit targeting for rotated tokens; bottom to top drag only selects fully enclosed tokens

jeffsenn:
A Token Drag Select from bottom to top (rather than top down) only selects tokens fully enclosed by the drag box rather than tokens that intersect the area of the drag box

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -4657,7 +4657,6 @@ function drawing_mouseup(e) {
 			const isCircle = curr.options.tokenStyleSelect == 'circle'
 			const size = curr.options.size * ((isCircle || curr.options.tokenStyleSelect == 'square' || $("#tokens>div[data-id='" + id + "']").hasClass("isAoe")) ? (curr.options.imageSize || 1) : 1);
 			const R = (isCircle ? 0 : curr.options.rotation) || 0;
-			console.log("RRRRRR", curr.options.size, fullyInside, x0, y0, x1, y1, CX, CY, size, R);
 			if(! (fullyInside ? isRotatedSquareInsideRect : intersectsRotatedSquare) (
 				{x:x0, y:y0, width:x1-x0, height: y1-y0}, CX, CY, size, R)) continue;
 			c++;
@@ -4673,8 +4672,6 @@ function drawing_mouseup(e) {
 		$("#temp_overlay").css('z-index', '25');
 		window.MULTIPLE_TOKEN_SELECTED = (c > 1);
 		draw_selected_token_bounding_box();
-		
-
 		console.log("READY");
 	}
 	else if (window.DRAWFUNCTION == "measure") {


jeffsenn:
In case you want to try that feature I mentioned on #feature-requests. Not high high priority - and possibly confusing, but also very useful.

jeffsenn:
no intended functionality change in second commit - just cleaner

jeffsenn:
One thing that this does not take into account is rotated tokens (particularly circular ones rotated by 45-ish) - where the bounding box is made artificially bigger by the rotated square boundary.  It's not visually obvious that you need a bigger selection rectangle to enclose them.   I should fix this if we decide to merge this one.  It would be better to use token size and location rather than bounding box in this case.

Azmoria:
Honestly this feels fine, it definitely has less impact then I expected. Maybe for someone who doesn't know there's a different it might be weird but I didn't even notice it when randomly selecting tokens, I had to be purposeful with the selection boxes to see the difference. Which probably only really matters in clustered situations you're trying to help here so that's good with me.

I like the cell cursor you added to indicate a difference. Tempted to make or use a google icon svg to indicate only tokens in the box and swap that to the main one. This is good for now ofc.


I do think it'd be nice to adjust the 'hitboxs' to match rotated tokens better too yeh. 

Note there is a difference on selectable area based on token style.  

Square/circle take into account the image scale as well as size due to both affecting the tokens roughly the same way.

Where as other styles like virtual mini, no contraints etc. don't take that image scale into account and just use the token 'base' size (medium, large etc) as the selectable target. This allows top down tokens and objects to break out of their base size without distrupting selection of other tokens. Especially images used that have transparent padding - which is a lot of what the other styles end up used for. Often top down tokens or others with transparency have to be scaled up to fit their square anyway which would cause issues of transparent area being selectable like we see with rotated tokens atm.

jeffsenn:
So I fixed circle - so this should be strictly better than existing.  However I did not include the hit select for rotated squares/rectangles.  This turns out to be a fair bit of complex/subtle geometry code (200 loc or so) that is only noticeable in unusual cases - if you want me to add, let me know.

Azmoria:
yeah, that does sound a bit excessive if it's not very noticiable. It's not like this behaviour was introduced with this anyway just kinda stands out more for the whole token needs to be selected part. Maybe we could just for that type of selection calculate the 4 corners x/y and check if those are all in it? Or would that be more complex/the same as what you were thinking?

jeffsenn:
Oh - that covers the "fully inside" case - and is less code -- yeah, I could do that pretty easily - it's about 1/4 of the code (the intersect code involves rotated rect projections).... Let me take a look at doing just the inside case (annoying that it's "architecturally asymmetrical" but ... pragmatics)

jeffsenn:
Ok - I thought better and I hope this covers all the cases - I think doing math on the center of the token (which I think is always square or circular for the purposes of hit target) makes this much easier.  Check it and let me know if I missed any case (maybe offsets or non-square aspect or something I'm not thinking of)

PR #4057, opened 2026-03-09
Refactored Extension Loader

jeffsenn:
- changed some files to ".mjs" (supported for a long time) to indicate module status\
    (I only changed ones I need to... we probably shoul change all that are modules for future sanity)
- removed a bunch of "duplicate" code to inject iframes by including an injection function in window
- Load.js is now loaded by extension and also injected and run in the page context (for above)
- Removed a couple of unnecessary paths from manifest
- Reorganized Load.js so that it was more easily understandable (at least for me) what are the options for page variants

### DIFF
--- a/Startup.mjs
+++ b/Startup.mjs
@@ -173,7 +173,6 @@ $(function() {
               let token = window.TOKEN_OBJECTS[id];
               if(token.isPlayer() || token.isAoe())
                 continue;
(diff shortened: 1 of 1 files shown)


jeffsenn:
Fixed. Sorry for odd commit message - my git client freaked out...

PR #4061, opened 2026-03-10
Token "select all" Key + code cleanup

jeffsenn:
-- added general use iterators for tokens and selected tokens
-- made "select all" keyhandler work for tokens in "select" mode
-- removed some no-longer-useful code


Azmoria (commented):
Going through the code overall looks good 👍 . I have to do in app testing but looks like it _should_  just all work.

Just one minor comment about the hotkey listing.

PR #4068, opened 2026-03-11
WIP: Modernize pointer events and new select box look

jeffsenn:
-- modernization of mouse events to pointer events
--- pointer capture to improve move-off-window cases
-- new SVG select box for cleaner look (and central control of styling)
--- better styling for contrast on mixed color scenes
-- replace jquery draggable with map specific code and streamline
--- install "token dragger/rotator" once (instead of every event time)

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -950,7 +950,7 @@ function rollDiceButton(e, button){
   }
 
 
-  window.diceRoller.roll(new DiceRoll(rollData.expression, rollData.rollTitle, rollData.rollType));
+  window.diceRoller.roll(new DiceRoll(rollData.expression, rollData.rollTitle, rollData.rollType, undefined, undefined, undefined, undefined, undefined, rollData.damageType, rollData.spellSave));
 
 }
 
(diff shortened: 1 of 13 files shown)


Azmoria:
Putting a couple comments here just so they don't get lost in discord chat. Video examples in the discord chat linked.

https://discord.com/channels/815028457851191326/930926587392692335/1481872050069962864
> got around to doing some testing on mobile
> 
> With touch the selection box starts and gets stuck until next tap. Also triggers on zoom, so appears to be any touch.
> 
> On live it just doesnt draw selection box for touches since we use it to move the map. I think we'd need different tool options to get both selection drag and move drag to have the ideal world on mobile.
> 
> Didn't notice any other issues on mobile from what I can tell so far.

https://discord.com/channels/815028457851191326/930926587392692335/1481872362445209681

> I also find the blend  mode of difference for  the selection box pretty wonky when going across multiple colored tokens or  parts  of a scene. Makes a lot of different  colors for the box which I find hard to follow exactly where it is. 
> 
> A drop-shadow might be a better option to give contrast when white is hard to see. This may have been my original  issue if that's been there the whole time tbh. Especially since the diff between white selection and grey map ends up grey and blends with the map.
> 
> `filter: drop-shadow(1px 1px 1px black);` for example below

https://discord.com/channels/815028457851191326/930926587392692335/1481875859387187271

> Another issue, right click when making a selection box draws the old one instead of dragging the map. Should be able to draw the selection box while holding mouse 1 and move the map with mouse 2 still.

Azmoria:
More comments also in discord:

> right click also isn't creating a way  point for the ruler, when drawing walls - so  it looks like it's just not working properly when holding mouse 1 for anything that allows different behaviour for mouse 2 
>
> middle mouse also is triggering things when  mouse1 is held like the old selection box is being drawn when holding left mouse then using middle  click to move the map
> 
> It seems neither 'pointerdown' or 'pointerup' events trigger on right click if another button is already held? That might be a pretty big problem to how things are setup. 
> 
> I dont think we can just put it in the move event as it could trigger a bunch of waypoints/segments etc. Even though it looks like it might detect it even without a move of the cursor when another button is clicked. We may not be able to use these events or at least not alone.
> 
> Doing reading I think we have to still use mousedown/up if we ever want to detect those correctly and seperately, at least without a rewrite/rethinking how things should work. Seems to be the case for both the document and draw listeners at least. 

> More minor but these are visible on initial load until click
image

jeffsenn:
Some updates to visual stuff (if you get a chance: see if you like).  Still working on event stuff....

jeffsenn:
moved visual changes into separate PR: #4107

jeffsenn:
closing for now

PR #4092, opened 2026-03-16
Fix: Add error handling to AboveApi fetch calls

Jumbalicious79:
## Summary

Added error handling to all three unprotected `fetch()` calls in `AboveApi.js`:

- **`fetchJson()`** — the core method used by `getSceneList()`, `getCurrentScene()`, `getCampaignData()`, `getScene()`, and `exportScenes()`
- **`setCampaignData()`** — the PUT request for campaign data updates
- **`migrateScenes()`** — the POST request for scene uploads + localStorage write

## What changed

Each fetch call now has three layers of protection:

1. **Network errors** (offline, DNS failure, timeout): `try/catch` around `fetch()` — logs and re-throws with context
2. **HTTP errors** (4xx, 5xx): `request.ok` check — logs status code and re-throws
3. **JSON parse errors** (non-JSON response): `try/catch` around `.json()` — logs and re-throws

Also wrapped the `localStorage.setItem()` in `migrateScenes()` with `try/catch` for quota error safety.

## Why

Previously, network failures or server errors caused cryptic `TypeError: Cannot read properties of undefined` errors. Now callers get descriptive errors like `"AboveApi HTTP 500: Internal Server Error"` or `"AboveApi network error: Failed to fetch"`.

Existing `.catch()` handlers (e.g., in `MessageBroker.js`) will receive more informative error objects. No caller changes needed — this is backward compatible.

## What's NOT changed

- No retry logic (separate concern)
- No UI error messages (caller's responsibility)  
- No changes to any callers or other files
- No changes to `checkForErrors()` — still handles API-level error messages
- Happy path is completely unchanged

## Files changed

- `AboveApi.js` — 1 file, +55/-6 lines

### DIFF
--- a/AboveApi.js
+++ b/AboveApi.js
@@ -45,20 +45,26 @@ class AboveApi {
     try {
       request = await fetch(url);
     } catch (error) {
-      console.error(`AboveApi.fetchJson network error for action '${action}':`, error);
-      throw new Error(`AboveApi network error: ${error.message}`);
+      const wrappedError = new Error(`AboveApi network error for '${action}': ${error.message}`);
+      console.error(wrappedError.message, error);
+      showError(wrappedError);
+      throw wrappedError;
     }
     if (!request.ok) {
       const errorText = await request.text().catch(() => 'Unknown error');
-      console.error(`AboveApi.fetchJson HTTP ${request.status} for action '${action}':`, errorText);
-      throw new Error(`AboveApi HTTP ${request.status}: ${errorText}`);
+      const httpError = new Error(`AboveApi HTTP ${request.status} for '${action}': ${errorText}`);
+      console.error(httpError.message);
+      showError(httpError);
+      throw httpError;
     }
     let response;
     try {
       response = await request.json();
     } catch (error) {
-      console.error(`AboveApi.fetchJson JSON parse error for action '${action}':`, error);
-      throw new Error(`AboveApi: invalid JSON response for '${action}'`);
+      const parseError = new Error(`AboveApi: invalid JSON response for '${action}'`);
+      console.error(parseError.message, error);
+      showError(parseError);
+      throw parseError;
     }
     this.checkForErrors(response);
     return response;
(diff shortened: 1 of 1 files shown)


Azmoria:
Is this not for the most part just catching the error logging it and rethrowing it? We already do this and display it with `showErrorMessage` / `showError` across the board.

We want this displaying to people to help with support. It's much easier to get reports know what's actually happening and solve them this way.

We can always add better descriptions of the errors in `showErrorMessage` or call it directly if that's the issue you're trying to solve.

Edit: I think this is true for the rest of your PRs as well. All these should already be displaying the error message/report window.

Jumbalicious79:
Good point — I see the global handlers in monitor_console_logs() that catch uncaught errors. 
 A few thoughts:

  ▎ For #4093 (DDBApi.js) — I think this one is still needed. lookForErrors already calls showError to display to
  the user, but without the throw, the .then() chain resolves with undefined and the caller crashes on
  undefined.json() with a secondary TypeError. The throw just prevents that double-error.

  ▎ For #4092 (AboveApi.js) — fair point on the catch/rethrow being redundant. The one piece that isn't covered by
   the existing error path is the request.ok check — right now a 500 response with valid JSON passes through
  fetchJson() without hitting showError unless checkForErrors() catches it by message text. Happy to slim this
  down to just the request.ok check if you'd prefer.

  ▎ For #4094 #4095 #4096 — you're right, these could interfere with the global handler logging. I can close those if you'd rather let errors bubble naturally.

  What's your preference? Happy to adjust.

Azmoria:
> For https://github.com/cyruzzo/AboveVTT/pull/4093 (DDBApi.js) — I think this one is still needed. lookForErrors already calls showError to display to
> the user, but without the throw, the .then() chain resolves with undefined and the caller crashes on
> undefined.json() with a secondary TypeError. The throw just prevents that double-error.

Agreed, may as well show the actual error positions. Can usually find them in the stack but still would be nice.

> For #4092 (AboveApi.js) — fair point on the catch/rethrow being redundant. The one piece that isn't covered by
> the existing error path is the request.ok check — right now a 500 response with valid JSON passes through
> fetchJson() without hitting showError unless checkForErrors() catches it by message text. Happy to slim this
> down to just the request.ok check if you'd prefer.

Also agreed that we should try and catch as much as possible - as long as the report window comes up (and associated logs added) for people I'm ok with catching more.

We'll often use console.warn for non-breaking errors like when we put a scene in the main scene 'folder' because we can't find it's parent folder. These get logged so when people go to help->get help->copy logs we can see them but no popup by default since nothing actually breaks.

Where as things not loading or saving correctly is where I want to make sure errors are displayed. I just dislike silent errors is all, hard to figure out the issues when things go wrong, especially when people aren't good with computers/unfamiliar with dev tools. Getting too many reports is better then trying to walk through a bunch of steps with people to replicate stuff essentially.

> For #4094 #4095 #4096 — you're right, these could interfere with the global handler logging. I can close those if you'd rather let errors bubble naturally.

The global log handling is used for both the error report window and the copy logs. So we definitely want to preserve that.

 However if we want more specific error messaging we can catch stuff and use showError or do something like the `if(stack.includes('Internal Server Error')...` section in `showErrorMessage`. 

I've been wanting to add some description for failed to fetch errors, these come up often enough just due to connection blips that it'd be nice to include the response we generally give in error report window. To let them know to refresh/try again as it's usually a blip in connection/server response if there isn't wide spread issues.

Overall as long as the error is being caught and displayed in the report window we can get more granular with our error catching. So if these are already showing the error window when caught I can merge them as is. I can force some errors and test if you think they are in a good spot already for that.

Jumbalicious79:
Thanks for the detailed feedback! Updated the PR to call `showError()` in all catch blocks before re-throwing.

**What changed:**

Network errors, HTTP errors, and JSON parse errors in `fetchJson()`, `setCampaignData()`, and `migrateScenes()` now all trigger the error report window.

**Pattern used:**

```javascript
const wrappedError = new Error(`AboveApi network error for '${action}': ${error.message}`);
console.error(wrappedError.message, error);
showError(wrappedError);
throw wrappedError;
```

**Result:**
- ✅ Error report window appears for users (supports your support workflow)
- ✅ Error still propagates to callers with their own `.catch()` handlers
- ✅ `console.error` logs preserved for the copy-logs feature
- ✅ `request.ok` check catches HTTP errors that `checkForErrors()` misses

Azmoria:
Looks good 👍 , thanks for making those changes. I'll  probably look at actively testing/merging these in a couple days. I want to get out hotfix I think for the text display issues being reported.

Issue #4009, opened 2026-02-28
can't access property "json", request is undefined


granthouston-glitch:
**Error:**
```
can't access property "json", request is undefined
Failed to start AboveVTT on https://www.dndbeyond.com/characters/144689117?abovevtt=true
fetchJsonWithToken@moz-extension://d32e9fd7-c35e-4f6f-a59e-0742474c4f26/DDBApi.js:56:5

```
**Environment:**
```
{"avttVersion":"1.49.1","browser":{"name":"Firefox","version":"148","mozilla":true,"chrome":false,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2026-02-28T01:55:18.656Z
      showErrorMessage
      lookForErrors@moz-extension://d32e9fd7-c35e-4f6f-a59e-0742474c4f26/DDBApi.js:43:17


ERROR 2026-02-28T01:55:18.655Z
      DDB API Error: Unknown Error 504 

ERROR 2026-02-28T01:55:18.654Z
      DDBApi.lookForErrors Failed to parse json
      {}
      {}

LOG 2026-02-28T01:54:18.208Z
      DDBApi.fetchCampaignInfo

LOG 2026-02-28T01:54:18.202Z
      harvest_campaign_secret found it in localStorage
```

Jumbalicious79:
## This should be fixed by PR #4093

The error `request is undefined` at `fetchJsonWithToken` line 56 was caused by `lookForErrors()` not throwing on error paths — it called `showError()` but returned `undefined`, which then crashed on `.json()`.

PR #4093 (merged 2026-03-20) adds `throw` statements to both error paths in `lookForErrors()`, so the promise chain now properly rejects instead of resolving with `undefined`.

If you're still seeing this on the latest version, please reopen.

Jumbalicious79:
All fixes for this crash path have shipped in v1.52 beta builds (beta6+):

- **#4093** — `lookForErrors` now throws, preventing the `undefined.json()` TypeError
- **#4114** — `fetchCampaignInfo` retry with exponential backoff
- **#4123** — `fetchCampaignCharacterIds` retry with exponential backoff

@Azmoria — this is part of the `lookForErrors` crash cluster (#4009, #4124, #3812, #3396, #3291, #3287, #3231). All share the same root cause and all fixes are merged. Can these be closed?

Issue #3396, opened 2025-07-24
Cannot read properties of undefined (reading 'json')


willsansbury:
**Error:**
```
Cannot read properties of undefined (reading 'json')
Failed to start AboveVTT on https://www.dndbeyond.com/characters/90355607?abovevtt=true
TypeError: Cannot read properties of undefined (reading 'json')
    at DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:56:26)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:193:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:325:26)
    at async start_above_vtt_for_players (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:430:3)
```
**Environment:**
```
{"avttVersion":"1.40","browser":{"name":"Chrome","version":"138","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2025-07-24T22:27:26.364Z
      showErrorMessage
      Error: DDB API Error: Unknown Error 404 
    at lookForErrors (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:43:17)
    at async DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:55:21)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:193:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:325:26)
    at async start_above_vtt_for_players (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:430:3)

ERROR 2025-07-24T22:27:26.362Z
      DDB API Error: Unknown Error 404 

ERROR 2025-07-24T22:27:26.361Z
      DDBApi.lookForErrors Failed to parse json
      {}
      {"message":"Failed to execute 'json' on 'Response': Unexpected end of JSON input"}

ERROR 2025-07-24T22:27:25.654Z
      Failed to parse json
[...]

Issue #3291, opened 2025-05-01
Cannot read properties of undefined (reading 'json') [user note: sorry to bother you guys again, i try refreshing and it doesn't work]


catfries:
**Error:**
```
Cannot read properties of undefined (reading 'json')
Failed to start AboveVTT on https://www.dndbeyond.com/characters/138415002?abovevtt=true
TypeError: Cannot read properties of undefined (reading 'json')
    at DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:56:26)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:187:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:322:26)
    at async start_above_vtt_for_players (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:426:3)
```
**Environment:**
```
{"avttVersion":"1.38","browser":{"name":"Chrome","version":"135","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2025-05-01T18:19:40.673Z
      showErrorMessage
      Error: DDB API Error: Unknown Error 404 
    at lookForErrors (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:43:17)
    at async DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:55:21)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:187:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:322:26)
    at async start_above_vtt_for_players (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:426:3)

ERROR 2025-05-01T18:19:40.672Z
      DDB API Error: Unknown Error 404 

ERROR 2025-05-01T18:19:40.671Z
      DDBApi.lookForErrors Failed to parse json
      {}
      {"message":"Failed to execute 'json' on 'Response': Unexpected end of JSON input"}

LOG 2025-05-01T18:19:40.428Z
      2 aoe spells discovered

[...]

Issue #3287, opened 2025-04-26
DDB API Error: Unknown Error 403


Carpe321:
**Error:**
```
Cannot read properties of undefined (reading 'json')
Failed to start AboveVTT on https://www.dndbeyond.com/encounters/6ca423f7-ac14-4b68-9e52-c1a6c23b02c2?abovevtt=true
TypeError: Cannot read properties of undefined (reading 'json')
    at DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:56:26)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:187:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:322:26)
    at async start_above_vtt_for_dm (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:367:3)
```
**Environment:**
```
{"avttVersion":"1.38","browser":{"name":"Chrome","version":"135","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2025-04-26T23:14:56.337Z
      showErrorMessage
      Error: DDB API Error: Unknown Error 403 
    at lookForErrors (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:43:17)
    at async DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:55:21)
    at async DDBApi.fetchCampaignInfo (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:187:22)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:322:26)
    at async start_above_vtt_for_dm (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:367:3)

ERROR 2025-04-26T23:14:56.336Z
      DDBApi.lookForErrors Failed to parse json
      {}
      {}

ERROR 2025-04-26T23:14:56.336Z
      DDB API Error: Unknown Error 403 

LOG 2025-04-26T23:14:56.143Z
      start_above_vtt_common

LOG 2025-04-26T23:14:56.143Z
[...]

PR #4093, opened 2026-03-16
Fix: DDBApi.lookForErrors now throws on error paths

Jumbalicious79:
## Summary

`DDBApi.lookForErrors()` calls `showError()` on HTTP 4xx/5xx responses but did not return or throw. The function fell through to `undefined`, causing every caller to crash with a `TypeError` when trying to call `.json()` on `undefined`.

## The Bug

```javascript
// Before: lookForErrors shows the error to the user but falls through
static async lookForErrors(response) {
    if (response.status < 400) return response;
    // error paths call showError() but never return or throw
    showError(new Error(`DDB API Error: ...`));
    // falls through to undefined
}

// Callers do this:
const request = await fetch(url).then(DDBApi.lookForErrors);
return await request.json();  // TypeError: Cannot read properties of undefined
```

## The Fix

Store the Error object, pass it to `showError()` (same user-facing behavior), then `throw` it:

```javascript
const error = new Error(`DDB API Error: ...`);
showError(error);  // user still sees the error dialog
throw error;       // .then() chain properly rejects
```

This is applied to both error paths in `lookForErrors`:
- HTTP 410 path (cache clearing guidance)
- General 4xx/5xx path (API error parsing)

## Impact

Affects **every DDB API call** in the application:
- `#refreshToken()` — auth token refresh
- `fetchJsonWithToken()` — all authenticated GET requests
- `fetchJsonWithTokenOmitCred()` — CORS-omit requests
- And transitively: `fetchCharacter`, `fetchMonsters`, `fetchEncounter`, `fetchCampaignInfo`, party inventory, etc.

## What's NOT changed

- `showError()` still displays the error dialog to the user (same UX)
- The error messages are identical
- `deleteWithToken()` explicitly skips `lookForErrors` (by design) — unaffected
- No changes to any callers

## Files changed

- `DDBApi.js` — 1 file, +6/-2 lines

### DIFF
--- a/DDBApi.js
+++ b/DDBApi.js
@@ -27,7 +27,9 @@ class DDBApi {
       return response;
     }
     if(response.status == 410){
-      showError(new Error(`DDB 410 Error`), `<b>Try clearing <div style="backdrop-filter: brightness(0.8);padding: 0px 3px;display: inline-block;border-radius: 5px;">${navigator.userAgent.indexOf("Firefox") != -1 ? `temporary cached files and pages` : `cached images and files`}</div> and restarting the browser.</b>`, `<br/><b>As long as you do <span style='color: #900;'>not</span> clear <div style="backdrop-filter: brightness(0.8);padding: 0px 3px;display: inline-block;border-radius: 5px;">cookies and other site data</div> this should not remove any AboveVTT data.`);
+      const error = new Error(`DDB 410 Error`);
+      showError(error, `<b>Try clearing <div style="backdrop-filter: brightness(0.8);padding: 0px 3px;display: inline-block;border-radius: 5px;">${navigator.userAgent.indexOf("Firefox") != -1 ? `temporary cached files and pages` : `cached images and files`}</div> and restarting the browser.</b>`, `<br/><b>As long as you do <span style='color: #900;'>not</span> clear <div style="backdrop-filter: brightness(0.8);padding: 0px 3px;display: inline-block;border-radius: 5px;">cookies and other site data</div> this should not remove any AboveVTT data.`);
+      throw error;
     }
     else{
       // We have an error so let's try to parse it
@@ -40,7 +42,9 @@ class DDBApi {
       if(type == 'EncounterLimitException'){
         alert("Encounter limit reached. AboveVTT needs 1 encounter slot free to join as DM. If you are on a free DDB account you are limited to 8 encounter slots. Please try deleting an encounter.")
       }
-      showError(new Error(`DDB API Error: ${type} ${messages}`));
+      const error = new Error(`DDB API Error: ${type} ${messages}`);
+      showError(error);
+      throw error;
     }
 
   }


Jumbalicious79:
DDBApi.js) — I think this one is still needed. lookForErrors already calls showError to display to
the user, but without the throw, the .then() chain resolves with undefined and the caller crashes on
undefined.json() with a secondary TypeError. The throw just prevents that double-error.

Jumbalicious79:
Thanks — no changes needed on this one per your feedback on #4092. The `throw` ensures `showError()` (already called by `lookForErrors`) is followed by a proper rejection instead of falling through to `undefined.json()`.

Jumbalicious79:
Post-merge update: traced all call sites and confirmed this fix has complete coverage — all 3 anchor points (lines 18, 59, 117) where `lookForErrors` is used in `.then()` chains are covered. No remaining crash paths.

Also identified 6 open issues that all hit this exact bug (DDB HTTP error → `lookForErrors` doesn't throw → `undefined.json()` crash):

- #4124 (v1.51.1) — DDB 504 at `fetchCampaignCharacters`
- #4009 (v1.49.1) — DDB 504 at `fetchCampaignCharacters`
- #3812 (v1.51.1) — `fetchCampaignInfo` → `undefined.json()`
- #3396 — `fetchCampaignInfo` → `undefined.json()`
- #3291 — `fetchCampaignInfo` → `undefined.json()`
- #3287 — DDB 403 at `fetchCampaignInfo` → `undefined.json()`

All 6 should be resolved when v1.52 ships. Commented on each issue linking back to this PR + #4114 (retry).

Jumbalicious79:
Live reproduction on v1.51.1 (current production) confirms both bugs:

**Test 1 — lookForErrors does NOT throw:**
```javascript
const fakeResponse = new Response('Gateway Timeout', { status: 504 });
const result = await DDBApi.lookForErrors(fakeResponse);
// result is undefined — lookForErrors fell through without throwing
```

**Test 2 — downstream crash:**
```javascript
const request = await Promise.resolve(fakeResponse).then(DDBApi.lookForErrors);
await request.json();
// TypeError: Cannot read properties of undefined (reading 'json')
```

Exact match to all 6 open issues. Will re-run these tests against the v1.52 build (with our fix) to confirm the fix resolves them.

Jumbalicious79:
**Fix verified against unpacked v1.52 fork (2026-03-22):**

| Test | v1.51.1 (before) | v1.52 fork (after) |
|---|---|---|
| lookForErrors return value | `undefined` (falls through) | Throws `"DDB API Error: Unknown Error 504"` |
| Downstream crash | `TypeError: Cannot read properties of undefined (reading 'json')` | Clean error propagation, no secondary crash |

Both tests reproduced the bug on v1.51.1, then confirmed the fix on v1.52. The 6 linked issues (#4009, #4124, #3812, #3396, #3291, #3287) should all be resolved when v1.52 ships.

PR #4095, opened 2026-03-16
Fix: Add .catch() to peerVideo promise chains

Jumbalicious79:
## Summary

Two promise chains in `peerVideo.js` had no `.catch()` handler, causing unhandled rejections when users deny permissions.

## Fixes

**1. `enumerateDevices()` (line 192)** — If media device permissions are denied or unavailable, the promise rejection was unhandled. Now logs a warning.

**2. `getDisplayMedia()` (line 303)** — If the user cancels the screen share dialog, the promise rejection was unhandled. Now logs a warning.

Both use `console.warn` rather than `showError` since these are expected user actions (denying permissions), not application errors.

## Files changed

- `peerVideo.js` — 1 file, +4 lines

PR #4094, opened 2026-03-16
Fix: Add .catch() to CharactersPage init chain

Jumbalicious79:
## Summary

The `harvest_game_id().then().then()...` chain in `CharactersPage.js` had no terminal `.catch()` handler. Any failure in the chain caused an unhandled promise rejection and silently prevented the character page from initializing.

## The Bug

```javascript
harvest_game_id()
  .then(set_game_id)
  .then(harvest_campaign_secret)
  .then(set_campaign_secret)
  .then(store_campaign_info)
  .then(async () => {
    window.CAMPAIGN_INFO = await DDBApi.fetchCampaignInfo(window.gameId);
    // ...
    window.MB = new MessageBroker();
  })
  // No .catch() — any failure silently kills init
```

## The Fix

Added `.catch()` matching the existing pattern in `CampaignPage.js` (line 56-58):

```javascript
  .catch(error => {
    console.error("Failed to initialize character page", error);
    showError(error, "Failed to initialize character page");
  })
```

## Files changed

- `CharactersPage.js` — 1 file, +5/-1 lines

PR #4096, opened 2026-03-16
Fix: Add .catch() to character refresh API call

Jumbalicious79:
## Summary

`debounce_fetch_character_from_api` calls `DDBApi.fetchCharacterDetails().then()` without a terminal `.catch()`. If the API call fails, the promise rejection is unhandled and character data silently stops updating for the session.

## The Fix

```javascript
  }).catch(error => {
    console.error("debounce_fetch_character_from_api: failed to fetch character details", error);
  });
```

Logs the error for debugging. Character updates will retry on the next debounce cycle (5 seconds later) since `PC_NEEDS_API_CALL` is repopulated by incoming WebSocket messages.

## Files changed

- `CoreFunctions.js` — 1 file, +2 lines

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -1969,6 +1969,8 @@ const debounce_fetch_character_from_api = mydebounce(() => {
         console.log(`debounce_fetch_character_from_api is not calling update_pc_with_data for ${characterData.characterId} because ${lastSynchronized} < ${idsAndDates[characterData.characterId]}`);
       }
     });
+  }).catch(error => {
+    console.error("debounce_fetch_character_from_api: failed to fetch character details", error);
   });
 }, 5000); // wait 5 seconds before making API calls. We don't want to make these calls unless we absolutely have to
 


Jumbalicious79:
For https://github.com/cyruzzo/AboveVTT/pull/4094 https://github.com/cyruzzo/AboveVTT/pull/4095 https://github.com/cyruzzo/AboveVTT/pull/4096 — you're right, these could interfere with the global handler logging. I can close these if you'd rather let errors bubble naturally.

Jumbalicious79:
Closing this PR — same reasoning as #4095. The character refresh API failure is already captured by `window.onunhandledrejection` → `addLog()` → `concerningLogs`. The function retries on a 5-second debounce cycle naturally, so the error is transient and self-recovering. Adding a `.catch()` here would prevent the global handler from logging it without adding meaningful value.

Thanks for the feedback on #4092 — better understanding of the project's error handling now.

PR #4104, opened 2026-03-17
Simplify and localize splash screen handling; close also with esc key

jeffsenn:
Keeps code closer to splash screen (out of main mouse handler)
fixes bug with clicking on a link and then back to tab
adds close splash on escape key press
It does change behavior slightly by closing on down rather than up


jeffsenn:
Also - Does splash screen really need a close button when pretty much every other event closes it?

Azmoria:
I don't feel it really need a close button no, probably just a visual to indicate it can be closed or a hold over from before any interaction closed it

PR #4106, opened 2026-03-17
Fix: Add optional chaining for CAMPAIGN_INFO.dmId in Startup.mjs

### LINKED ISSUE #4062: Cannot read properties of undefined (reading 'dmId')
**Error:**
```
Cannot read properties of undefined (reading 'dmId')
Failed to start AboveVTT on https://www.dndbeyond.com/characters/155235144?abovevtt=true
TypeError: Cannot read properties of undefined (reading 'dmId')
    at chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:51:37
```
**Environment:**
```
{"avttVersion":"1.50","browser":{"name":"Chrome","version":"145","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2026-03-10T23:44:16.161Z
      AboveApi.getCampaignData
      {"Item":{"campaignId":"[REDACTED]","data":{"cloud":1},"timestamp":1645811426224,"objectId":"campaigndata"}}
```

Jumbalicious79:
Fixes #4062

## The Bug

`Startup.mjs` line 51 accesses `window.CAMPAIGN_INFO.dmId` without a null check. If `DDBApi.fetchCampaignInfo()` returns `undefined` (network error, API failure, auth issue), the TypeError kills the entire startup chain:

```
TypeError: Cannot read properties of undefined (reading 'dmId')
Failed to start AboveVTT on https://www.dndbeyond.com/characters/...
```

## The Fix

```diff
- return window.CAMPAIGN_INFO.dmId;
+ return window.CAMPAIGN_INFO?.dmId;
```

## Why This Works

The downstream code already handles `undefined` gracefully:
- **Line 57:** `isDmPage && campaignDmId == userId` → `false` when `campaignDmId` is `undefined`
- **Line 64:** Same check → falls through to the player startup path at line 75

**Result:** AboveVTT still loads for the player instead of crashing with "Failed to start AboveVTT."

## Files changed

- `Startup.mjs` — 1 character change (`.` → `?.`)

### DIFF
--- a/Startup.mjs
+++ b/Startup.mjs
@@ -48,7 +48,7 @@ $(function() {
       .then(async () => {
         window.CAMPAIGN_INFO = await DDBApi.fetchCampaignInfo(window.gameId)
         window.AVTT_CAMPAIGN_INFO = await AboveApi.getCampaignData();
-        return window.CAMPAIGN_INFO.dmId;
+        return window.CAMPAIGN_INFO?.dmId;
       })
       .then((campaignDmId) => {
         const isDmPage = is_encounters_page();


Azmoria:
This is something we may want to either prompt users to refresh or implement some expontential back off retries to get the campaign info (or other requests to DDB). We should probably do that as soon as we try and make the request for window.CAMPAIGN_INFO and it fails tbh rather then wait until here.

It's required for quite a few things, dice rolls, ensuring we're saving/loading the proper journal (player vs dm), vision, chat whispers etc. Mostly this is the DM view but things like roll to DM will fail for players too.

It mostly only happens when DDB is having issues but the occasional blip might happen and retry or notice to refresh would be nice.

If we do implement retries with backoff we should probably cap it at 5 or less retries as we have to be careful about spamming DDB's server.

 Since so many things rely on it it likely wouldn't be very useable (at least for DM), we do probably want to stop avtt from loading at that point and tell the person that it's likely temporary and a refresh may resolve it if wider outages aren't happening.

Jumbalicious79:
## Closing in Favor of a New PR

Great feedback — you're right that `CAMPAIGN_INFO` is too critical for just a null guard. The optional chaining was a band-aid; what's really needed is resilience at the fetch level.

I'm closing this PR and opening a new one with the expanded scope:

- **Exponential backoff retry** on `DDBApi.fetchCampaignInfo()` in `Startup.mjs` (capped at 5 retries per your guidance)
- **`showError` + halt startup** if all retries fail — with a message that it's likely temporary and a refresh may resolve it
- **`console.warn`** on each retry attempt for diagnostics

Following the pattern from `avttFetchWithRetry` in `avttS3Upload.js` for the backoff logic.

New PR will reference this one and #4062.

Issue #3292, opened 2025-05-01
Bug - Failing to get campaign data is coming up more often for players.


Azmoria:
Failing to get campaign data is coming up more often in reports. 

- [ ] Check the DDBApi request returns valid info if given the correct campaign id.
- [ ] Check that where we grab it from the sheet, if the campaign page hasn't been visited and data cached, is still accessible/hasn't been changed.
- [ ] We should also check the 'My Characters' page.

---

For those having this issue try the below 

• Try joining from the campaign page if you're trying to join from the sheet itself. Once you've done this it should add the campaign id to the cache and it should fix the sheet button if that's where it's failing.

• Double check the character is complete in the character builder.

PR #4118, opened 2026-03-20
Fix: Guard against undefined playerUsers in fetchCampaignCharacterIds

### LINKED ISSUE #3613: Cannot read properties of undefined (reading 'filter')
**Error:**
```
Cannot read properties of undefined (reading 'filter')
Failed to start AboveVTT on https://www.dndbeyond.com/characters/153316365?abovevtt=true
TypeError: Cannot read properties of undefined (reading 'filter')
    at DDBApi.fetchCampaignCharacterIds (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:281:41)
    at async DDBApi.fetchCampaignCharacterDetails (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:220:26)
    at async rebuild_window_pcs (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/CoreFunctions.js:1668:30)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:355:3)
    at async start_above_vtt_for_players (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:962:3)
```
**Environment:**
```
{"avttVersion":"1.44.1","browser":{"name":"Chrome","version":"140","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

WARN 2025-10-02T01:15:58.558Z
      fetchCampaignCharacterIds caught an error trying to collect ids from fetchActiveCharacters
      {}

LOG 2025-10-02T01:15:58.549Z
      showErrorMessage
      Error: DDB API Error: Unknown Error 403 
    at lookForErrors (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:43:17)
    at async DDBApi.fetchJsonWithToken (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:55:21)
    at async DDBApi.fetchActiveCharacters (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:251:22)
    at async DDBApi.fetchCampaignCharacterIds (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:270:30)
    at async DDBApi.fetchCampaignCharacterDetails (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/DDBApi.js:220:26)
    at async rebuild_window_pcs (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/CoreFunctions.js:1668:30)
    at async start_above_vtt_common (chrome-extension://ipcjcbhpofedihcloggaichibomadlei/Startup.js:355:3)
    at async star
```

Jumbalicious79:
## Summary

When both `fetchCampaignCharacters()` and `fetchActiveCharacters()` fail (e.g., DDB returns HTTP 403 or 504), `window.playerUsers` is never assigned. Line 322 then crashes calling `.filter()` on `undefined`:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at DDBApi.fetchCampaignCharacterIds (DDBApi.js:281:41)
```

## Changes

**File:** `DDBApi.js` (+6/-1)

- Added null check for `window.playerUsers` before the `.filter()` call
- Returns early with `console.warn` for diagnostics
- Falls back to `CAMPAIGN_INFO.dmId` for `window.myUser` (same fallback the existing code uses)
- Added optional chaining on `CAMPAIGN_INFO.dmId` in both the new guard and the existing line for safety

## Before / After

**Before:** Both API calls fail → `window.playerUsers` is `undefined` → `.filter()` crashes → startup fails.

**After:** Both API calls fail → early return with warning → `myUser` set to DM ID as fallback → startup continues gracefully.

## Context

- PR #4093 (merged) fixed the upstream `lookForErrors` issue that caused many of these API failures, but a 403/504 from DDB can still leave `playerUsers` unset
- Related issue #3292 (Azmoria filed) documents this class of problem affecting players

Fixes #3613.

### DIFF
--- a/DDBApi.js
+++ b/DDBApi.js
@@ -319,8 +319,13 @@ class DDBApi {
         console.warn("fetchCampaignCharacterIds caught an error trying to collect ids from fetchActiveCharacters", error);
       } 
     }
+    if (!window.playerUsers) {
+      console.warn("fetchCampaignCharacterIds: failed to fetch campaign characters, falling back to DM");
+      window.myUser = window.CAMPAIGN_INFO?.dmId;
+      return characterIds;
+    }
     let playerUser = window.playerUsers.filter(d=> d.id == window.PLAYER_ID)[0]?.userId;
-    window.myUser = playerUser ? playerUser : window.CAMPAIGN_INFO.dmId; 
+    window.myUser = playerUser ? playerUser : window.CAMPAIGN_INFO?.dmId;
     return characterIds;
   }
 


Azmoria:
We don't want to default to setting the user id to the DM's user. 

In cases like this we expect valid values and we generally aren't wanting to default to something else. This is going to be true of almost all api calls during start up. 

We can implement more retry on some of the requests to avoid the server blips but we'll still get these reports when DDB servers are down.

Jumbalicious79:
Thanks for the feedback @Azmoria — you're right, defaulting to the DM's user ID isn't the right approach.

Reworked this as #4123 using retry with exponential backoff instead (same pattern as the merged #4114 for fetchCampaignInfo). On final failure it calls showError() and returns empty characterIds rather than crashing on undefined.filter().

Closing this in favor of #4123.

PR #4131, opened 2026-03-22
Fix: Guard draggable enable/disable calls in Token.place()

### LINKED ISSUE #3558: cannot call methods on draggable prior to initialization; attempted to call method 'enable'
**Error:**
```
cannot call methods on draggable prior to initialization; attempted to call method 'enable'

error@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:2616
V.widget.bridge/V.fn[o]/&lt;@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-ui.min.js:6:2212
each@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:3003
each@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:1481
V.widget.bridge/V.fn[o]@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-ui.min.js:6:1951
place@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/Token.js:2400:10
handleToken@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/MessageBroker.js:1743:34
MessageBroker/this.onmessage@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/MessageBroker.js:599:10

```
**Environment:**
```
{"avttVersion":"1.40","browser":{"name":"Firefox","version":"141","mozilla":true,"chrome":false,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

LOG 2025-09-07T00:05:45.199Z
      trovato!!

LOG 2025-09-07T00:05:45.198Z
      Setting combat tracker opacity to 1.0

LOG 2025-09-07T00:05:45.198Z
      cerco id/profile/111579999/characters/60978945

LOG 2025-09-07T00:05:45.197Z
      custom/myVTT/token

LOG 2025-09-07T00:05:44.665Z
      showErrorMessage
      error@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:2616
V.widget.bridge/V.fn[o]/<@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-ui.min.js:6:2212
each@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:3003
each@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-3.6.0.min.js:2:1481
V.widget.bridge/V.fn[o]@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/jquery-ui.min.js:6:1951
place@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/Token.js:2400:10
debounce_pc_token_update</<@moz-extension://8d5a5111-c262-4671-8769-4a0d3b70a56f/CoreFunctions.js:1555:15
debounce
```

Jumbalicious79:
## Summary
- Adds `hasClass('ui-draggable')` guard before `draggable("enable")` and `draggable("disable")` calls in the `place()` method
- Prevents `"cannot call methods on draggable prior to initialization"` crash

## Root Cause
When a token WebSocket update arrives via `handleToken()` while a scene is still loading tokens via `handleScene()` (which uses `async_sleep(0.01)` between tokens), the token's DOM element may exist but `draggable()` hasn't been initialized on it yet. The `place()` method then tries to call `old.draggable("enable")` on an uninitialized element.

## Changes
- Token.js line 2560: `old.draggable("disable")` → `if (old.hasClass('ui-draggable')) old.draggable("disable")`
- Token.js line 2565: `old.draggable("enable")` → `if (old.hasClass('ui-draggable')) old.draggable("enable")`
- Token.js line 2568: `old.draggable("enable")` → `if (old.hasClass('ui-draggable')) old.draggable("enable")`

The new token creation path (line 3072+) always initializes draggable before reaching the disable calls at lines 3555/3560, so those are not affected.

## Files changed
- `Token.js` — 3 lines guarded (+4/-4)

Fixes #3558

### DIFF
--- a/Token.js
+++ b/Token.js
@@ -2155,6 +2155,10 @@ class Token {
 			}		
 
 			if (old.length > 0) {
+				const hasDraggable = old.hasClass('ui-draggable');
+				if (!hasDraggable) {
+					console.warn(`Token.place(): draggable not initialized for token ${this.options.id} — skipping draggable enable/disable`);
+				}
 				if(this.options.type == 'door'){
 					this.options.size = 50;
 					setTokenLight(old, this.options);
@@ -2557,15 +2561,15 @@ class Token {
 				}
 				if((!window.DM && this.options.restrictPlayerMove && !this.isCurrentPlayer() && this.options.share_vision != true &&  this.options.share_vision != window.myUser) || this.options.locked){
 					if(!window.DM || (window.DM && !$('#select_locked>div.ddbc-tab-options__header-heading').hasClass('ddbc-tab-options__header-heading--is-active'))){
-						old.draggable("disable");
+						if (hasDraggable) old.draggable("disable");
 						old.removeClass("ui-state-disabled"); // removing this manually.. otherwise it stops right click menu
 					}
 				}
 				else if((window.DM && this.options.restrictPlayerMove && !this.isCurrentPlayer() && this.options.share_vision != true &&  this.options.share_vision != window.myUser) || !this.options.locked || (window.DM && !$('#select_locked>div.ddbc-tab-options__header-heading').hasClass('ddbc-tab-options__header-heading--is-active'))){
-					old.draggable("enable");
-				}	
+					if (hasDraggable) old.draggable("enable");
+				}
 				else if(!window.DM && ((!this.options.restrictPlayerMove  && !this.isCurrentPlayer())) || !this.options.locked){
-					old.draggable("enable");
+					if (hasDraggable) old.draggable("enable");
 				}
 				if(!this.options.id.includes('exampleToken')){
 					old.toggleClass('lockedToken', this.options.locked==true)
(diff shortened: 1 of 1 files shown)


Azmoria:
If we're looking at anything with `old` in front of it in `place()` means it has already gone through the new token creation below.

Maybe we just move this (and the variables just above it) up to ensure it's never added to dom before it's had draggable set?
image

If it still happens and we want to do this, I'd suggest setting a variable `const hasDraggable = old.hasClass('ui-draggable')` near the top of the `if (old.length > 0)` section to reduce rechecks. In general I want to go through this funciton and look at updating a bunch of stuff where we recall the same checks/dom elements.

Jumbalicious79:
Reworked per your feedback. Added `const hasDraggable = old.hasClass('ui-draggable')` at the top of the `if (old.length > 0)` block (line 2158) and use that cached variable in all three guard spots — no repeated DOM checks.

I looked into moving the draggable init above the DOM append as your primary suggestion. The issue is that between `append(tok)` (line 3032) and `tok.draggable({...})` (line 3073), several things run that depend on the element being in the DOM:
- `setTokenAuras(tok)` / `setTokenLight(tok)` position elements relative to the token's DOM location
- The draggable `start` callback reads computed CSS via `el.css("left")` (lines 3245-3266) which needs DOM attachment
- `update_opacity(tok, true)` likely needs computed styles

Moving draggable init above the append would require also moving the aura/light/opacity setup — a bigger refactor with regression risk. The cached variable approach prevents the crash cleanly without that scope.

Azmoria:
Moving the `tok.draggable(...)` function and `let ctx/dragStopTime` above _should_ be fine. 

This seems to work fine on my end at least
image

It's essentially adding a click/drag events to an element before adding it to the DOM. Can't start a click or drag unless it's interactable so we don't need to worry about the functions inside of `draggable()` running until it's in the DOM anyway.

Iirc, I _think_ the error of draggable not being initialized is most common in the past if an error happens between the lines `$("#tokens").append(tok)` and `tok.draggable()` preventing draggable from initializing. 

We can keep the ui-draggable class check just in case ofc - although if this does happen likely something else went wrong and a token has loaded in without being draggable. Maybe a console.error or warn if it does find it doesn't exist just to get it in the logs without being disruptive.

 Other movement like arrow keys and what not might still work if the rest of the place() function completed correctly.

Jumbalicious79:
Good call — reworked to do both:

**Structural fix:** Moved `tok.draggable({...})` and its variable declarations (`click`, `currentTokenPosition`, `ctx`, `dragStopTimer`) above `$("#tokens").append(tok)`. Draggable is now always initialized before the element enters the DOM, so the race window is eliminated. The darkness/opacity/aura setup stays after the append since those may need DOM positioning.

**Safety net:** Kept the `hasDraggable` class check in the existing-token path. Added `console.warn` when it triggers so it shows up in the logs without being disruptive — if it fires, something else went wrong during token creation.

Verified with `node -c Token.js` — no syntax errors.

PR #4132, opened 2026-03-22
Improved whisper menu (including Image Popup feature); Spectator whispering; Stat block cleaner look for Send to Gamelog

jeffsenn:
Adds a "Popup Image to Players" button next to "send to gamelog" for monster images.

This uses magnific to popup the image directly to players for 10 seconds.  (As if it was sent to gamelog and then clicked on in the gamelog).

I added this primarily for use at my game live table - where I just want to quickly pop an image but not have to do a lot of interaction to do so.

Note: also adds else's to message handling to optimize in MessageBroker

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -2835,9 +2835,7 @@ function display_url_embeded(url){
 
 //+++ dialog menus in the style of top level tools
 // todo:
-// -- make menu construction on-demand? (reduce resources?)
-// -- make sure dialogs fit on screen (positioning)
-// -- more general options for dialog menus
+// -- more general options for dialog menus (when used beyond send-player buttons)
 function dialogMenuOption(rootId, container, opt) {
   if(opt.type === 'hr') {
     $(`<span class="js-popup-decoration">${opt.label || ""}</span>`).appendTo(container);      
@@ -2874,6 +2872,8 @@ function createDialogMenuTrigger(iconName, menuId, create, onShow) {
   <span class="material-symbols-outlined">${iconName}</span></button>`);
   button.on('click', (e) => {
     //create the dialog on first use
+    e.preventDefault();
+    e.stopPropagation();
     const dialog = $("#"+menuId, e.target.ownerDocument)?.[0] || create(menuId, e.target);
     if(dialog.open) {
       dialog.close();
@@ -2930,8 +2930,8 @@ function createSendPlayerMenu(menuId, target) {
       //todo: if selected is everyone then use undefined here:
       send_html_to_gamelog(`<p>${targetBlock[0].outerHTML}</p>`, selected);
     } else {
-      const imgSrc = $(".magnify")?.attr("src")
-      //todo: decide to send to everyone or iterate through users....
+      //todo: deal with video
+      const imgSrc = $(".magnify, .monster-image")?.attr("src")
       if(imgSrc) {
         window.MB.sendMessage('custom/myVTT/Popup',  {
           src: imgSrc,
@@ -2939,6 +2939,8 @@ function createSendPlayerMenu(menuId, target) {
           whisper: selected,
           from:window.PLAYER_ID
         });
+      } else {
+        console.error("Could not find image to pop", imgSrc);
       }
     }
     $(e.target).removeClass('js-popup--is-active');
@@ -2974,11 +2976,11 @@ function createSendPlayerButton(parent, icon, hasPopupOption=false ) {
     $($(dialog).find("button")[1]).css("display", hasPopupOption ? "" : "none");
   });
   button.addClass("block-send-to-game-log");
+  button.addClass("send-player-button")
   if(hasPopupOption) button.addClass("has-popup-option");
-  button.css("right", "2px"); //todo: temp so we see both buttons
   return button;
 }
-//--- dialog menus 
+//-end- dialog menus 
 
 function find_or_create_generic_draggable_window(id, titleBarText, addLoadingIndicator = true, addPopoutButton = false, popoutSelector=``, width='80%', height='80%', top='10%', left='10%', showSlow = true, cancelClasses='', hideOnX = false, alwaysDisplayTitle = false) {
   console.log(`find_or_create_generic_draggable_window id: ${id}, titleBarText: ${titleBarText}, addLoadingIndicator: ${addLoadingIndicator}, addPopoutButton: ${addPopoutButton}`);
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -2985,9 +2985,7 @@ function createSendPlayerButton(parent, icon, hasPopupOption=false ) {
     sendClonedElement($(e.target).closest("button").parent());
   });
   if(hasPopupOption) button.attr("data-haspopup", "1");
-  //todo: evaluate stuff in this style:
   button.addClass("block-send-to-game-log");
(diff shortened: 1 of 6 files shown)


Azmoria:
Got to playing with this, I agree that it's nice to be able to quick show. I wonder if there's a case for a button for closing the image too (for people using a tabletop or projector) similar to notes.

Minor issue, needs  some css adjustment to make sure text is always in the buttons on that row. It might only affect official monsters since they have 3 buttons now. Maybe `height: fit-content` or something like that.

We already discussed and agreed it'd be nice to have the button where other send to buttons are, and individual whispers like you mention. We can do some of this in chunks, we don't need all the buttons available right away.

One other comment for consideration, it'd be nice to get it working for video tokens too. It's still magnific, some are auto detected based on file extension other are if the token has the video token flag.

Example of video popup and it not being setup for this yet
https://github.com/user-attachments/assets/cfae7388-d522-4515-9df9-0c1949825960

jeffsenn:
> Example of video popup and it not being setup for this yet
> https://github.com/user-attachments/assets/cfae7388-d522-4515-9df9-0c1949825960

@Azmoria - what was that example?  Is that a video as a link or image embed? (it's not an embedded video, right?)

Azmoria:
On the token? It's just a video file attached to the token like any image link. In this case it was this 

`https://dl.dropboxusercontent.com/scl/fi/cqkr786hmgr8wfc8syt87/Moonbeam.webm?rlkey=kryl7v4uxnwv20r46v3kvrc52&dl=0`

This is the code that does the popout. which does use an iframe, if that was the question. We use an extension html page to embed the video in since DDB made some security changes awhile back breaking magnific's default settings for video.
image

The one in chat and on the token are just `<video>` elements though

jeffsenn:
I added an alternative to the "send to gamelog" tiny button with a whole new dialog menu system (that should be more efficient - and is a little more stylish).  I'm not quite ready to make a full proposal yet, but I think combining the "send to gamelog/popup" big-blue-buttons with the overlay tiny icon on images is worth considering.   That is: don't take this for complete quite yet.

Azmoria:
Nice, took a quick look

Some things to look into/consider: 

- That button is used for all text containers, and for roll tables(to send results). We probably want to disable this menu for those containers/roll tables - can already whisper by right clicking the button with the current menu. Unless you intend to create a popup for individual containers/roll results but that seems like outside the initial scope of this.

- A few containers the buttons currently dont look positioned correctly (overlapping each other). Doesn't appear the new button is available for roll tables atm. It may be you intend to not have both buttons and just have the menu be right click

- You have all player characters being added to the list but showing as the DDB user. We should probably decide if we want to target player characters or DDB user here. DDB user keeps the list smaller in most cases and you don't have to worry about which character a player using 2 pcs logged into. Player character would allow you to select a specific character to display to. 
 
   example hard to tell which character I'm actually whispering:
      image

   Downside of doing per character rather then user (we play multiple games in this one DDB campaign due to sharing limits):
   image
   
   This is generally why we have always chosen user rather then player character in these menus.

- Would be nice to include spectator as an option.

---

Just opinions on current version looks (not something I would need to change for merging just preference/discussion purpose). You may just have some of this as placeholder anyway: 

• I think we should keep the send to gamelog icon consistent across all menus tbh looking at it. At least in the popup menu - should match the token menu icon (or the button you're replacing's icon)

• I don't like the emojis in the menu personally - this is probably platform dependant as they all appear different. If we want consistent looks across devices, sticking with google symbols/material icons or custom ones is probably best to keep a standard look. 

• I think this should take from the share with players menu more then the main menus looks.

   Checkboxes vs highlighting lines I think is the thing that is throwing me off. It's a lot of red highlighted lines which draw my view away from the buttons that have actions (especially in campaigns with a lot of pcs). Maybe it's just my expectation for it to be more similar though - I might get used to this.

jeffsenn:
Agree with most of that (is a work in progress - and a few things not done like popup availability as you mentioned - and yes, should switch to users)

However:

> • I think this should take from the share with players menu more then the main menus looks.
> 
> Checkboxes vs highlighting lines I think is the thing that is throwing me off. It's a lot of red highlighted lines which draw my view away from the buttons that have actions (especially in campaigns with a lot of pcs). Maybe it's just my expectation for it to be more similar though - I might get used to this.

I was trying to get a consistent "look" with our other controls (as I think we should go for that eventually) and I do very much like the top level menus/controls look ... but I also do agree with the confusion of "redness" being a "checkbox" (this is true of the main menus too for me).   So - I'm conflicted... maybe the right thing is to fix both : ie have a differentiation of the look of a toggle/checkbox vs an "action" button - that is adjust them in the top level too (I think we're just used to all of those and know intuitively which are toggles and which are actions).    For example: "Fog->Reveal->All" is very different from clicking "Fog->Reveal->Square" -- which we do mitigate by having a confirm popup.  But it would be nice to know before clicking that one was a tool toggle and the other was an action.

That was why I was trying "the emojis" -- to get them to look more "action-like" (though I'm not wedded to that at all)
Not sure what a better look would be.... maybe just not "super red" for the toggles?  Thoughts?

Azmoria:
Yeh maybe a slightly less super red, more of a tint, border change, opacity  or any of several css options we have.

I think the main menu having sections and descriptions that are more obvious help. I wouldn't expect clear all for drawings to work the same as draw rect for example. I see the issue though for sure, also why I moved some of the toggles down into the controls section of the wall menu. Felt like seperating the toggles from the tools that required clicking on the map made more sense. 

Maybe toggles should all go under a tool settings heading in general?

I'd like to add some more hover descriptions to other menus like I did the wall menu this past update.

If toggle all and all the users was in a seperate section something like that might be enough to differentiate it. Just a little more spacing essentially/grouping a bit different? Like the roll menu maybe (not the checkmarks just the sections)? 

image


**Actions(or any title):**
Popup
Gamelog

**Send To:**
Toggle All button
Spectator
...Users

or flip it if it makes more sense for the 'actions' to be at the bottom. What ever feels right.

jeffsenn:
Yep make sense. I'll play with it.  (I put actions at the top so it's a quick send in most cases - since menu will tend to appear very near mouse and you probably have the users selected from last time)

jeffsenn:
This is now close to ready for review.  The one issue I'd like to improve is the size of the send-to button to make it easier to hit.

jeffsenn:
Improved visibility of the button.
Ready for review

Azmoria:
Looks good for the most part. Left click works as expected. Right click send to gamelog and popup whisper the correct people. Images and note text containers all work well

A few issues:

- Some video tokens aren't having the button added to the statblock. It seems to be working for custom statblocks but not official with a replaced image.
   image
   
-  Looks like the popup for video may still need to be looked at as well. Whispering the correct user but not loading the video version of the popup still I think.

- Roll tables send to gamelog buttons something is going wrong with adding the button element to the rows. You _may_ end up wanting to use a smaller button for roll table rows. Some people might have pretty small row heights.
   image
   
- The hover note previews don't stay open when using the right click menu. The menu probably just needs to have `.prevent-sidebar-modal-close` class added to it. `remove_tooltip()` checks for if current hovered elements have this so hover notes don't close when interacting with other elements. Otherwise send to fails as it goes to look to clone the element after it disappeared.

   https://github.com/user-attachments/assets/766a982b-85d3-4e53-824e-4d7d120d4fb3

PR #4152, opened 2026-03-25
Fix: Projector toggle sets ProjectorEnabled to true when activating

Jumbalicious79:
**The bug:** Main.js lines 2530-2533 — both branches of the projector toggle click handler set `window.ProjectorEnabled = false`. The else branch (activating) should set it to `true`. The CSS class is toggled correctly on the icon, but the JS flag stays `false`, so projector mode never actually activates.

**Fix:** Change `false` to `true` in the else (activation) branch.

**Verified in Chrome:** Clicked the projector toggle, checked `window.ProjectorEnabled` — stayed `false` after both activate and deactivate clicks.

**Files changed:** `Main.js` (+1/-1)

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -2527,16 +2527,13 @@ function init_zoom_buttons() {
 			const iconWrapper = $(event.currentTarget).find(".ddbc-tab-options__header-heading");
 			if (iconWrapper.hasClass('ddbc-tab-options__header-heading--is-active')) {
 				iconWrapper.removeClass('ddbc-tab-options__header-heading--is-active');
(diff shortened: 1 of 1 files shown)


Azmoria:
Looks like we might not use this anywhere. Probably better to just remove `window.ProjectorEnabled` from being set/in memory at all tbh.

Jumbalicious79:
Good catch — you're right, nobody reads `window.ProjectorEnabled` anywhere. Reworked to remove all 3 assignments instead of fixing the value. Force-pushed.

PR #4163, opened 2026-03-26
Fix: Guard scrollIntoView with optional chaining in combat tracker

Jumbalicious79:
**The bug:** CombatTracker.js lines 369-370 (NEXT button) and 433-434 (PREV button) call \`[0].scrollIntoView()\` on jQuery selectors for \`tr[data-current=1]\`. If the selector returns no results (e.g., popout combat tracker window hasn't updated yet, or combat was cleared), \`[0]\` returns \`undefined\` and \`.scrollIntoView()\` throws \`TypeError: Cannot read properties of undefined\`.

**Reproduced in Chrome:** \`\$("tr[data-nonexistent=999]")[0].scrollIntoView()\` throws \`TypeError: Cannot read properties of undefined (reading 'scrollIntoView')\`.

**Fix:** Added \`?.\` optional chaining on all 4 instances (2 in NEXT handler, 2 in PREV handler). The scroll is silently skipped when no current row exists.

**Files changed:** \`CombatTracker.js\` (+4/-4)

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -366,10 +366,10 @@ function init_combat_tracker(){
 		debounceCombatPersist();
 		ct_update_popout();
 		if(window.childWindows['Combat Tracker'] != undefined)
-			$(window.childWindows['Combat Tracker'].document).find("tr[data-current=1]")[0].scrollIntoView({ behavior: 'instant', block: 'center', start: 'inline' });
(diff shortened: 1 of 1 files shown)


Azmoria:
We can guard against this but since we're hitting next/prev I think we can reasonably expect there will be current turn token in the popout window.

PR #4166, opened 2026-03-28
Add a "lock view" button that rubber bands any panning or zooming to spectator; Add store visual grid size based on current view

Azmoria:
If anyone has any opinions on how this should be implemented let me know. I haven't played with a touch screen irl so don't have any strong opinion myself. I believe the original request being to snap back to a saved position for real life minis alignement but allow some scouting around.

Currently this allows scroll/zoom of the map to see nearby things then will snap back 2 seconds after the last scroll event.

### DIFF
--- a/Main.js
+++ b/Main.js
@@ -2853,7 +2853,36 @@ function init_zoom_buttons() {
 			$(window).off('scroll.resetToLockedPos')
 		})
 	}
-	const gridZoomConversion = $(`<div id='grid_zoom_conversion' class='ddbc-tab-options--layout-pill hideable'><div class='ddbc-tab-options__header-heading hasTooltip button-icon' data-name='Grid Zoom Conversion'><span class='material-symbols-outlined md-16 button-icon'>visibility</span></div></div>`);
+	const gridZoomConversion = $(`<div id='grid_zoom_conversion' class='ddbc-tab-options--layout-pill hideable'><div class='ddbc-tab-options__header-heading hasTooltip button-icon' data-name='Grid Zoom Conversion'><svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24.73 24.74">
+  <g>
+    <path d="M1.5,8.5c.15,0-1.39,0-1.24,0h.72" style="fill: none; stroke: #231f20; stroke-linecap: round; stroke-linejoin: round; stroke-width: .5px;"/>
+    <line x1="3.67" y1="8.5" x2="21.14" y2="8.5" style="fill: none; stroke: #231f20; stroke-dasharray: 3.58 2.69; stroke-linecap: round; stroke-linejoin: round; stroke-width: .5px;"/>
+    <line x1="22.48" y1="8.5" x2="24.48" y2="8.5" style="fill: none; stroke: #231f20; stroke-linecap: round; stroke-linejoin: round; stroke-width: .5px;"/>
+  </g>
+  <g>
+    <path d="M8.32.86c0,.15,0-.75,0-.6v1.34" style="fill: none; stroke: #231f20; stroke-linecap: round; stroke-linejoin: round; stroke-width: .5px;"/>
+    <line x1="8.32" y1="4.22" x2="8.32" y2="21.19" style="fill: none; stroke: #231f20; stroke-dasharray: 3.48 2.61; stroke-linecap: round; stroke-linejoin: round; stroke-width: .5px;"/>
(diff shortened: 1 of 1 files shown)


Azmoria:
Examples:

https://github.com/user-attachments/assets/87f12583-1efc-4858-aae9-269bd82a2bfd


https://github.com/user-attachments/assets/69ada25d-9abf-4214-aec7-0e98b59ac2b7

Jumbalicious79:
Love this idea — my group just started a hybrid minis/digital game with a screen table last night. We use the combat tracker for initiative and on large maps the spectator view jumps around a lot. Re-centering on the minis after each combat turn was a chore. A lock view button would be a huge QoL improvement for hybrid games.

One thought on the UX: rather than auto-snapping back after 2 seconds, what about splitting this into two buttons?

1. **Lock/Unlock toggle** — position the spectator view where you want it, click to lock (saves position + zoom). Click again to unlock and clear the saved position.
2. **Return button** — while locked, press this (or Shift+K) to snap back to the saved position + zoom from wherever you are. Move around freely in between — DM scroll, manual scroll, no restrictions.

To lock a different scene, just unlock and re-lock at the new position.

This gives the user full control over when to return and avoids a potential conflict I noticed: on the spectator page, `projectionScroll` messages (Startup.mjs:262-285) call `window.scroll()`, which would fire the scroll event and trigger the auto-debounce snap-back — potentially fighting with the DM's projected scroll.

Two manual buttons would be simpler to implement and more predictable in use. Happy to help with the implementation if you'd like!

Azmoria:
Yeah im not 100% on if I like the implementation yet. I dont play in person with a screen so happy to adjust to what works best.

The snap back when using projector mode was intended. Allowing the dm to scroll to or ping things out of view then snap back shortly after without having to interact with the other screen (or device if using ping to center view on a different device). 

I think having to use the spectator mode window was the issue so auto snapping back was the request. They also use a touch device where players accidentally scroll sometimes when moving. However they want to allow scrolling out side of that. So toggling lock on/off could get annoying everytime they want to move the screen for a second.

We may not end up with this as the final way we do it though. Id prefer to have fewer toggle buttons where possible but could see a saved snap point being useful too.


Im curious why youre having to recenter on minis after every turn. Is the combat area just larger then the screen? So youre having to move spectator view every turn? In that case, Im not sure a single locked position helps.

 Or maybe you have the combat tracker set to center on the token in spectator view? Otherwise tracking with the combat tracker shouldnt affect the  position of the screen at all.

I would expect most combats, that fit on screen, you'd just not touch the zoom/scroll. Very rarely would you want to show something outside the combat area so the screen would be static most of the time.

Jumbalicious79:
Great questions!

1. Yeah, the combat area is larger than the screen. To get the map resolution to match real-life miniatures on a table TV, we have to zoom in quite a bit. Using the Thundertree map as an example — the map covers the whole village, but the Venomfang fight is only in one corner. If the party isn't all together when the encounter triggers, some players are outside the battle zone and you need to scroll over to resolve their turns. But then you want to snap back to the main fight without having to re-drag and re-zoom everything to line up with the physical minis again.

That's why the auto snap-back on a timer wouldn't quite work for us — you might be resolving something outside the battle zone for multiple turns (10-20 minutes in D&D time), and then want to snap back to the main fight instantly. A manual "return to saved position" button would be perfect for that.

2. We originally had center-on-active-token enabled but turned it off. Either way, we still need to scroll outside the main battle area to resolve other things during combat.

I can see how the auto snap-back makes more sense for the projector/touch-device use case though. Maybe both approaches could coexist — the auto-return for projector mode users, and a manual saved-position button for hybrid table setups?

Azmoria:
That all makes sense

To play around with functionality I added some quick key binds of shift+f1 through shift+f4 to save location  and f1 though f4 to go  to the  saved location. Works in all views at the moment.

I know there's  some browser functions  like find in page and help but I'm not sure how important these are or if other browers/OS have hotkeys that are  important on those.

Mostly just to see how it feels, not sure on a UI for it yet since crowding the screen is a bit of a problem for DM in the side toggles atm.

https://github.com/user-attachments/assets/58ab70f9-62a9-4c79-9623-0f5abe3908d1

Jumbalicious79:
I'm hoping to get time to test this tonight. 🤞

Jumbalicious79:
Tested the Shift+F1–F4 save / F1–F4 recall keybinds locally — loaded the branch as an unpacked extension and it worked great! The save/recall feels snappy and intuitive.

I don't think we need a UI confirmation on the page but some sort of confirmation of loaded positions would be nice. Or some sort of text confirmation in the Gamelog? Not sure what would be useful but also out of the way.

Azmoria:
Merged in #4269

PR #4195, opened 2026-04-01
Fix: toggleClass() passed array instead of string for wall/elevation toggles

Jumbalicious79:
## Summary
- Shift+W (toggle walls) and Shift+E (toggle elevation) pass an array to jQuery's `toggleClass()`, which expects a space-separated string
- The array coerces to a comma-joined string, creating a garbled class name that doesn't match anything
- Fix: pass a space-separated string instead of an array

## Test plan
- [ ] Open AboveVTT as DM with a scene that has walls
- [ ] Press Shift+W — walls should toggle on/off visually
- [ ] Press Shift+E — elevation overlay should toggle on/off
- [ ] Verify the button styling toggles correctly (highlighted when active)

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -256,14 +256,14 @@ Mousetrap.bind('w', function () {
 });
 Mousetrap.bind('shift+w', function () {
     if(window.DM){
-        $('#show_walls').toggleClass(['button-enabled', 'ddbc-tab-options__header-heading--is-active']);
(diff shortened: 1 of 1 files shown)


Azmoria:
This functions as expected already and toggles both classes and says it's supported here https://api.jquery.com/toggleClass/

image

Maybe there's a specific version that doesn't support this?

Jumbalicious79:
Closing — you're right, jQuery 3.3+ supports arrays in toggleClass. Should have checked the docs for 3.6.0 before assuming.

PR #4196, opened 2026-04-01
Fix: hasClass() in handle_undo() uses space-separated string that never matches

Jumbalicious79:
## Summary
- `handle_undo()` passes a space-separated two-class string to jQuery's `hasClass()`, which only accepts a single class name
- The string `"button-enabled ddbc-tab-options__header-heading--is-active"` is treated as one class name, which never matches any element
- **Result: Ctrl+Z undo is completely non-functional** for all tools (select, fog, draw, vision, text, wall, elevation)
- Fix: use `hasClass("button-enabled")` which is sufficient to identify the active tool button

## Test plan
- [ ] Open AboveVTT as DM
- [ ] Select the Draw tool, draw something, press Ctrl+Z — should undo
- [ ] Select the Fog tool, draw fog, press Ctrl+Z — should undo
- [ ] Select tokens, delete them, press Ctrl+Z — should undo
- [ ] Repeat for wall, text, elevation, vision tools

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -596,27 +596,26 @@ Mousetrap.bind('[', function(e) {
 });
 
 function handle_undo(){
-    const buttonSelectedClasses = "button-enabled ddbc-tab-options__header-heading--is-active"
 
-    if ($("#select-button").hasClass(buttonSelectedClasses)){
+    if ($("#select-button").hasClass("button-enabled")){
         undo_delete_tokens();
     }
-    else if(($("#fog_button").hasClass(buttonSelectedClasses))){
+    else if(($("#fog_button").hasClass("button-enabled"))){
         $("#fog_undo").click()
     }
-    else if($("#draw_button").hasClass(buttonSelectedClasses)){
+    else if($("#draw_button").hasClass("button-enabled")){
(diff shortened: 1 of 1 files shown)


Azmoria:
These all work as expected currently. 
We probably actually never want arrow keys by themselves moving the screen is why that one works. We probably need to prevent default in left and right as well looking at it. We can probably just always prevent these.

The undo's also all work based on tool selected. I believe it's looking for either both classes beside each other or seperately as `$("#fog_button").hasClass('button-enabled ddbc-tab-options__header-heading--is-active')` returns true but `$("#fog_button").hasClass('button-enabled apple')` doesn't when the fog button is clicked.

I do find it interesting that this does work though.
image

Azmoria:
accidentaly hit close with comment

Jumbalicious79:
Closing — this PR bundled multiple unrelated changes (hasClass, visibleMenuId, 'n' key binding) which should have been separate. The hasClass fix itself was wrong — jQuery's indexOf-based implementation handles space-separated strings when the classes are adjacent. The visibleMenuId and 'n' key changes are in their own PRs (#4199, #4197).

PR #4197, opened 2026-04-01
Fix: merge combat tracker guard into duplicate 'n' key binding

Jumbalicious79:
## Summary
- The `n` key was bound twice by Mousetrap — the second binding silently overwrites the first
- The first binding (line 42) had a guard checking that the combat tracker is visible before advancing initiative
- The second binding (line 104) was unconditional, so pressing `n` anytime as DM would advance combat even when the tracker isn't open
- Fix: merge the visibility guard into the second binding and remove the dead first binding

## Test plan
- [ ] As DM with combat tracker closed, press `n` — should do nothing
- [ ] As DM with combat tracker open, press `n` — should advance initiative
- [ ] As player during your turn, press `n` — should end your turn

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -39,12 +39,6 @@ Mousetrap.bind('f', function () {       //fog menu
 });
 
 
-Mousetrap.bind('n', function () {   //while combat menu is open, press n to cycle next in initiative order
-    if(window.DM && $('#combat_tracker_inside').attr('style') == 'display: block;') {
-        $('#combat_next_button').click()
-    }
-});
-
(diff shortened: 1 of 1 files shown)


Azmoria:
We want to be able to advance the tracker when it's not visible (I assume that is a hold over from an old version). Especially if they have it closed and are only using the carosel. We can remove the top one  though for sure

Jumbalicious79:
Closing to rework — per your feedback, the combat tracker should advance even when not visible (carousel use). Reopening with a cleaner change that just removes the dead first binding.

PR #4199, opened 2026-04-01
Fix: visibleMenuId always truthy when no menu is open

Jumbalicious79:
## Summary
- Arrow key handlers construct `visibleMenuId` by interpolating `.attr("id")` into a template string
- When no menu element matches, `.attr("id")` returns `undefined`, producing `"#undefined"` — always truthy
- This causes `e.preventDefault()` to fire for every arrow keypress, even when no menu is open
- Fix: check jQuery result length before constructing the selector string

## Test plan
- [ ] Open AboveVTT with no menus open
- [ ] Press arrow keys — should scroll/pan the map normally (no blocked behavior)
- [ ] Open a tool menu (e.g., fog submenu)
- [ ] Press up/down arrows — should navigate menu items

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -359,8 +359,9 @@ function moveKeyWatch() {
 
 Mousetrap.bind('up', function (e) {
     arrowKeysHeld[0] = 1;
-    const visibleMenuId = `#${$('[id*="_menu"].visible').attr("id")}`
-    if (visibleMenuId){
+    const $visibleMenu = $('[id*="_menu"].visible');
+    if ($visibleMenu.length > 0){
+        const visibleMenuId = `#${$visibleMenu.attr("id")}`;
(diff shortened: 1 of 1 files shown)


Azmoria:
> We probably actually never want arrow keys by themselves moving the screen is why that one works. We probably need to prevent default in left and right as well looking at it. We can probably just always prevent these.

Copied from other PR that had same lines

Jumbalicious79:
Closing — tested and arrow keys are working as designed. They move selected tokens (via `moveKeyWatch`/`forSelTokens`), not pan the map. The `throttleMoveRequest()` call is outside the `if` block, so token movement is unaffected by the truthy string. The unnecessary `e.preventDefault()` on Up/Down with no menu open has no practical impact in a full-canvas app. Not a user-visible bug.

PR #4200, opened 2026-04-01
Fix: self-closing style tag drops mixer paused CSS rule

Jumbalicious79:
## Summary
- Three `<style>` elements use self-closing `/>` syntax (`<style id="mixer-paused" />`)
- HTML's `<style>` is not a void element — `/>` closes the tag immediately
- The CSS rule text (`#sounds-panel button.pressed.playing{...}`) becomes an orphaned text node, never applied as styling
- Fix: change `/>` to `>` so the CSS rule is inside the style element

## Test plan
- [ ] Open the audio mixer as DM
- [ ] Play a track, then pause the mixer
- [ ] Verify the paused state shows the gold highlight background on the play/pause button
- [ ] Resume playback — gold highlight should disappear

### DIFF
--- a/audio/ui.js
+++ b/audio/ui.js
@@ -425,7 +425,7 @@ function init_mixer() {
                 mixer_pause_svg.css('display', 'none');
                 playPauseMixer.toggleClass('playing', false);
                 playPauseMixer.toggleClass('pressed', false);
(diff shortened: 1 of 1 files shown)


Azmoria:
Interestingly It seems smart enough to correct this itself - I was wondering how this worked to begin with when looking at this. This may not be ithe best implementation anyway. Css vars might be better then adding/removing dom elements but merging this for now.

PR #4205, opened 2026-04-01
Fix: guard TRACK_LIBRARY.find() result in Edit and Tags context menu

Jumbalicious79:
## Summary
- The Edit and Tags context menu callbacks call `TRACK_LIBRARY.find()` which returns `undefined` when no track matches
- Accessing `[1].tags` on `undefined` throws TypeError, crashing the context menu
- This can happen if the track was deleted between opening the context menu and clicking an option
- Fix: add null check before accessing the result

## Test plan
- [ ] Open the track library, right-click a track
- [ ] Click "Edit" — should open edit fields without error
- [ ] Click "Tags" — should open tag editor without error
- [ ] Verify both work correctly for existing tracks

### DIFF
--- a/audio/ui.js
+++ b/audio/ui.js
@@ -935,7 +935,9 @@ function init_trackLibrary() {
                     const trackSrcInput = $(`<input class='trackSrc trackInput' placeholder='https://.../example.mp3'/>`)
                     const okButton = $('<button class="add-track-ok-button">OK</button>');  
                     const cancelButton = $('<button class="add-track-cancel-button">X</button>');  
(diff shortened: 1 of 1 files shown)


Azmoria:
I think this is redundant the track should always exist here since we are already right clicking on it here to edit it's name/url or adjust tags.

Are you able to get an item in the listing that isn't a part of the library somehow? If so that's probably the issue we should be addressing instead.

Jumbalicious79:
Closing — you're right, the track should always exist here since we're acting on the right-clicked element. Unnecessary guard.

PR #4206, opened 2026-04-01
Fix: guard findIndex result before splice in delete_scene

Jumbalicious79:
## Summary
- `delete_scene()` uses `findIndex()` to locate a scene, then passes the result to `splice()`
- If the scene isn't found, `findIndex` returns `-1`, and `splice(-1, 1)` removes the **last** element — silently deleting the wrong scene
- Fix: check for `-1` before splicing

## Test plan
- [ ] Delete a scene from the scenes panel — should remove the correct scene
- [ ] Verify remaining scenes are unaffected after deletion

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -948,7 +948,9 @@ class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 	delete_scene(sceneId, reloadUI = true) { // not the index, but the actual id
 		window.MB.sendMessage("custom/myVTT/delete_scene",{ id: sceneId });
 		let sceneIndex = window.ScenesHandler.scenes.findIndex(s => s.id === sceneId);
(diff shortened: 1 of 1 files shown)


Azmoria:
The scene should always exist if we are calling delete. If you are able to produce a situation where we try to delete a scene that doesn't exist in `window.ScenesHandler.scenes` we should address that rather then skipping it here. It means we failed to add the scene to `window.ScenesHandler.scenes` to begin with which probably indicates a different issue that will cause issues elsewhere.

It's hard to tell what should always exist vs we are ok with not existing without a lot of pre-exiting knowledge of the code base. So don't worry about submitting stuff like this - always happy to review and point out which we need to make sure exists elsewhere rather then guard.

Jumbalicious79:
Closing — agreed, if the scene doesn't exist in ScenesHandler.scenes that's a different problem to address upstream, not guard around here.

PR #4207, opened 2026-04-01
Fix: guard splitPlayerScenes access in persist_current_scene

Jumbalicious79:
## Summary
- `persist_current_scene()` calls `Object.values(window.splitPlayerScenes)` without checking if it exists
- When the split-scene feature is not in use, `splitPlayerScenes` is `undefined` → TypeError crash
- The neighboring methods `persist_scene()` (line 917) and `delete_scene()` (line 959) both guard with `window.splitPlayerScenes?.players != undefined`
- Fix: add the same guard before accessing `splitPlayerScenes`

## Test plan
- [ ] Save a scene without split-scene feature active — should not crash
- [ ] If split-scene is available: save a scene with split players — should still sync correctly

### DIFF
--- a/ScenesHandler.js
+++ b/ScenesHandler.js
@@ -935,11 +935,13 @@ class ScenesHandler { // ONLY THE DM USES THIS OBJECT
 
 		sceneData.isnewscene=false;
 		window.MB.sendMessage("custom/myVTT/update_scene",sceneData,dontswitch);
-		const currentPlayerScenes = Object.values(window.splitPlayerScenes);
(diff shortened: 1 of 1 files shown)


Azmoria:
This should always exist as it's set in several locations including in `AboveApi.getCurrentScene()` which is run in all view start ups (in this case it only really matters for DM).

If you're able to have it not exist we need to be addressing that rather then protecting against it.

Jumbalicious79:
Closing — same reasoning. It's set during startup and should always exist by this point.

PR #4209, opened 2026-04-01
Fix: remove duplicate critRange=3 from double damage roll call

Jumbalicious79:
## Summary
- `damage_dice_context_menu()` passes `3` as both `critRange` (arg 3) and `forceCritType` (arg 7) for double damage rolls
- `critRange=3` causes the visual crit-success highlight to trigger on nearly every roll (threshold of 3+)
- The `doubleDamage` variable already correctly sets `forceCritType=3` for the actual damage formula doubling
- The inline ternary in arg 3 appears to be a leftover from when the `doubleDamage` variable was added
- Fix: pass `undefined` for `critRange` so the default (20) is used

## Test plan
- [ ] Right-click a damage roll, select "Double Damage"
- [ ] Roll — damage should be doubled but the d20 should only show crit highlight on natural 20
- [ ] Try "Crit Damage" — should still double dice count correctly
- [ ] Try "Flat" — should roll normally

### DIFF
--- a/DiceContextMenu/DiceContextMenu.js
+++ b/DiceContextMenu/DiceContextMenu.js
@@ -150,7 +150,7 @@ function damage_dice_context_menu(diceExpression, modifierString = "", action = 
 
             const doubleDamage = rollAsIndex === 2 ? 3 : undefined;
 
-            window.diceRoller.roll(diceRoll, undefined, rollAsIndex == 2 ? 3 : undefined, undefined, spellSave, damageType, doubleDamage);
+            window.diceRoller.roll(diceRoll, undefined, undefined, undefined, spellSave, damageType, doubleDamage);
             
         });
 


Azmoria:
Looking at this again it's actually supposed to be the fourth variable (crit type) instead of crit range.  Since we are forcing crit type it being undefined is probably fine though.
Crit range I believe only affects hit rolls so I don't think it actually does anything here (right click damage rolls).

Also the text should always exist if we are double clicking it to edit it - if text exist on the screen but not in window.DRAWINGS we messed up somewhere else by removing it from window.DRAWINGS after placing it without removing it from the screen. (other dif line)

---

We actually probably need to send forceCrit in tab communication here too looking at it. I don't think this doubles damage auto.

image

Edit: just added this in #4214

Jumbalicious79:
Closing — thanks for identifying the real fix (crit type, not crit range) and handling it in #4214. This PR also had unrelated Text.js changes mixed in from a dirty branch.

PR #4210, opened 2026-04-01
Fix: text shadow regex crash on rgba() color values

Jumbalicious79:
## Summary
- `draw_text()` parses shadow CSS with a regex that only matches `rgb(...)`, not `rgba(...)`
- When the shadow color is in `rgba()` format, `.match()` returns `null` and destructuring crashes with TypeError
- Fix: update regex to `rgba?` and add null guard — fallback to no shadow instead of crashing

## Test plan
- [ ] Add text with shadow enabled to a scene
- [ ] Verify shadow renders correctly with default color
- [ ] Change shadow color to one that produces rgba() format
- [ ] Verify text still renders (with or without shadow) — no crash

### DIFF
--- a/Text.js
+++ b/Text.js
@@ -638,9 +638,12 @@ function draw_text(
 
     let shadowStyle ='';
     if (font.shadow && font.shadow !== "none"){
-        const shadowRegex = /(rgb\(.*\))\s(\d+px)\s(\d+px)\s(\d+px)/
-        const [_, shadowColor, shadowOffsetX, shadowOffsetY, shadowBlur ] = font.shadow.match(shadowRegex)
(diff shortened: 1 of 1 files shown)


Azmoria:
Text shadow may have been setup to eventually accept other colors but currently it is only a toggle on or off. Is a particular browser returning the value as rgba instead of rgb maybe? 

If drop shadow exists it should always be `rgb(0, 0, 0) 5px 5px 5px` at the moment. We _could_ hard code these unless we want to leave them open to future color selection/sizing etc.

Jumbalicious79:
Closing — the shadow is currently toggle-only with a fixed value. The regex mismatch between the code setting "black 5px 5px 5px" and the regex expecting rgb() may depend on browser normalization, but since it's not causing issues in practice, not worth changing right now.

PR #4213, opened 2026-04-01
Fix: guard filter result in text dblclick edit handler

Jumbalicious79:
## Summary
- Double-clicking map text to edit it filters `window.DRAWINGS` for a matching entry
- If the drawing was deleted between render and double-click, `.filter()[0]` returns `undefined`
- Subsequent property access on `undefined` throws TypeError, crashing the edit flow
- Fix: add early return if no matching drawing is found

## Test plan
- [ ] Double-click existing map text — should open the edit box with current text/styles
- [ ] Verify editing and re-submitting text works correctly

### DIFF
--- a/Text.js
+++ b/Text.js
@@ -754,6 +754,7 @@ function draw_text(
         });
         textSVG.on('dblclick', function(){
             let text_data = window.DRAWINGS.filter(d => d[9] == this.id)[0];
+            if (!text_data) return;
             if(window.TEXTDATA == undefined){
                 window.TEXTDATA = {};
             }


Azmoria:
From other PR where you had this line

> Also the text should always exist if we are double clicking it to edit it - if text exist on the screen but not in window.DRAWINGS we messed up somewhere else by removing it from window.DRAWINGS after placing it without removing it from the screen.

Jumbalicious79:
Closing — same as the note on #4209. If text exists on screen but not in DRAWINGS, that's a different issue to address at the source.

I learned a lot from the feedback on this batch as usual — thank you for your patience.

PR #4220, opened 2026-04-02
Fix: remove dead 'n' key binding overridden by later binding

Jumbalicious79:
The first `Mousetrap.bind('n')` at line 42 (with combat tracker visibility guard) is immediately overridden by the second binding at line 104. Only the last binding fires — the first is dead code.

This just removes the dead first binding. The second binding is left as-is per feedback on #4197 — advancing the tracker when it's not visible is intended for carousel use.

Reworked from closed #4197.

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -39,12 +39,6 @@ Mousetrap.bind('f', function () {       //fog menu
 });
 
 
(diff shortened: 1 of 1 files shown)


Jumbalicious79:
Closing — tested on current main and the `n` key correctly advances combat both with the tracker open and closed. The second binding at line 104 handles both cases as intended. This is just dead code removal with no behavior change, not worth the review overhead.

PR #4198, opened 2026-04-01
Fix: missing closing parentheses in CSS blur() and rgba() template strings

Jumbalicious79:
## Summary
- Three template literal CSS values in WeatherOverlay.js are missing closing `)`:
  - Lines 54, 70: `blur(${value}px` → `blur(${value}px)` — fog and light canvas blur filters
  - Line 945: `rgba(120, 120, 120, ${alpha}` → `rgba(120, 120, 120, ${alpha})` — fog particle color
- Invalid CSS is silently ignored by browsers, so fog blur and fog particle rendering fail silently
- Compare with line 59 which correctly has `blur(1px)` with the closing paren

## Test plan
- [ ] Open AboveVTT as DM, set weather to "Fog" on a scene
- [ ] Verify fog particles render with visible color (not invisible/transparent)
- [ ] Verify fog has a blur effect applied (soft edges, not sharp)
- [ ] Test "Lightning" weather — verify light canvas blur works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4221, opened 2026-04-02
Fix: missing closing parentheses in CSS blur() and rgba() template strings

Jumbalicious79:
Three template literals in WeatherOverlay.js are missing closing `)` in CSS function values:

- **Line 54**: `blur(${hpps}px` → `blur(${hpps}px)`
- **Line 70**: `blur(${hpps/scale}px` → `blur(${hpps/scale}px)`
- **Line 950**: `rgba(120, 120, 120, ${alpha}` → `rgba(120, 120, 120, ${alpha})`

All three produce invalid CSS that silently fails — fog blur filter doesn't apply, fog color is ignored.

Reworked from closed #4198 (removed unrelated KeypressHandler.js changes that leaked from another branch).

### DIFF
--- a/WeatherOverlay.js
+++ b/WeatherOverlay.js
@@ -51,7 +51,7 @@ class WeatherOverlay {
             
             if (this.type === 'fog') {
                 $(this.canvas).css({
-                    filter: `blur(${ window.CURRENT_SCENE_DATA.hpps }px`
+                    filter: `blur(${ window.CURRENT_SCENE_DATA.hpps }px)`
                 })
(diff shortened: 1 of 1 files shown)


Jumbalicious79:
Closing — tested in Chrome and the browser's CSS parser auto-corrects the missing closing parentheses. Both `blur(Xpx` and `rgba(120, 120, 120, 0.5` are silently fixed to valid values. Fog blur renders correctly on current main. Not a user-visible bug.

PR #4229, opened 2026-04-03
Fix: guard itemData access in party loot click handler

Jumbalicious79:
## Bug
`itemData[0]` at Journal.js lines 3079 and 3087 crashes when `find_items_in_cache_by_id_and_name()` returns an empty array. This happens when the item name in the party loot table HTML doesn't exactly match the `ITEMS_CACHE` entry name.

The table-build code at line 2973 already guards with `if (itemData.length>0)` — the click handler at line 3067 lacks the same check.

## Fix
Add the same length check used at line 2973 before accessing `itemData[0]`, with `console.warn` for log tracking per Azmoria's feedback.

```diff
  const itemData = find_items_in_cache_by_id_and_name([{id, name}]);
+ if (itemData.length === 0) {
+     console.warn("Item not found", [{id, name}]);
+     return;
+ }
  
  console.log(`[PartyLoot] Adding ${quantity} of item ${name} to queue`);
```

## Chrome Verification

**Test Environment:** Chrome/146.0.0.0, AboveVTT v1.53, macOS

**Reproduction:** `find_items_in_cache_by_id_and_name([{id: 99999999, name: "Nonexistent Item XYZ"}])` returns `[]`, then `itemData[0].quantity = 1` throws `TypeError: Cannot set properties of undefined (setting 'quantity')`

**Pre-fix behavior:** Clicking a party loot item that doesn't match `ITEMS_CACHE` crashes with TypeError.

**Post-fix behavior:** No crash. `console.warn("Item not found", [{id: 99999999, name: "Nonexistent Item XYZ"}])` logged for support tracking. Early return prevents the crash.

**Happy path:** `find_items_in_cache_by_id_and_name([{id: 4, name: "Longsword"}])` → length 1, `quantity` set correctly. 0 concerningLogs.

## Files Changed
- `Journal.js` — party loot click handler (line 3067)

### DIFF
--- a/Journal.js
+++ b/Journal.js
@@ -3064,7 +3064,11 @@ class JournalManager{
 				const name = $(this).closest('tr').find('.item-link-cell a').text();
 				
 				const itemData = find_items_in_cache_by_id_and_name([{id, name}]);
-				
+				if (itemData.length === 0) {
+					console.warn("Item not found", [{id, name}]);
+					return;
+				}
+
 				console.log(`[PartyLoot] Adding ${quantity} of item ${name} to queue`);
 				
 				if (quantity > 10){


Azmoria:
I think we should add a `console.warn("Item not found", [{id, name}])` or something along those lines to help track down issues if people submit logs.

Jumbalicious79:
Added `console.warn("Item not found", [{id, name}])` as suggested. Rebased onto latest main and Chrome-verified:

- Non-matching item: no crash, warn logged with id + name for support tracking ✅
- Matching item (Longsword): works normally, quantity set correctly ✅
- 0 concerningLogs ✅

PR #4257, opened 2026-04-05
Fix - more explicit line of sight cache clearing for tokens, light, drawings, fog, elev (point los tools and token vision; Fix - issue where wall updates were not recalculating los for some point los tools

Valamorde:
## Problem

`window.lightDrawingLosCache` caches line-of-sight polygon calculations keyed by token position. It was never cleared between scenes or when walls changed, so stale entries from old scenes accumulated in memory indefinitely.

Two failure modes:
1. **Scene switch** — old cache entries are never reused (the `sceneId` check at read time rejects them) but stay in memory forever
2. **Mid-scene wall edits** — after a wall is added or removed, `redraw_light_walls` is called but the cache wasn't cleared, meaning a subsequent LOS query could theoretically match a stale entry if wall count happened to be the same

## Changes (`Fog.js` only)

- **`reset_canvas()`** — clear cache at the top; runs on every scene load/resize
- **`redraw_light_walls()`** — clear cache at the top; runs on every wall add/remove/edit mid-scene (this is the critical gap)
- **Cache write site** — evict the whole cache if it grows beyond 500 entries within a single scene (guard against unbounded growth from many unique token positions)

## Test plan
- [ ] Switch between several scenes and confirm memory usage does not grow with each switch
- [ ] Add/remove walls mid-scene and confirm LOS is recalculated correctly (no stale polygons)
- [ ] Move many tokens around a single scene and confirm the 500-entry eviction triggers without visual glitches

### DIFF
--- a/CharactersPage.js
+++ b/CharactersPage.js
@@ -1745,7 +1745,7 @@ function observe_character_sheet_changes(documentToObserve) {
 
 
     if(is_abovevtt_page()){
-      if($('.dice-rolling-panel').length == 0 && window.diceWarning == undefined){
+      if($('.dice-rolling-panel, [data-floating-ui-portal]').length == 0 && window.diceWarning == undefined){
         showDiceDisabledWarning();
       }
       else if($('.dice-rolling-panel').length > 0){
@@ -3006,7 +3006,7 @@ function observe_non_sheet_changes(documentToObserve) {
   window.non_sheet_observer = new MutationObserver(function(mutationList, observer) {
     if(window.DRAGGING || (typeof arrowKeysHeld !== 'undefined' && (arrowKeysHeld[0] || arrowKeysHeld[1] || arrowKeysHeld[2] || arrowKeysHeld[3])))
       return;
-    mutationList.forEach(mutation => {
+    mutationList.every(mutation => {
       try {
         let mutationTarget = $(mutation.target);
         //Remove beyond20 popup and swtich to gamelog
@@ -3022,7 +3022,9 @@ function observe_non_sheet_changes(documentToObserve) {
           }
           
           window.MB.reprocess_chat_message_history();
+          return false;
         }
+        return true;
       }
       catch(error){
         console.warn("non_sheet_observer failed to parse mutation", error, mutation);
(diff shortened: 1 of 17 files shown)


Azmoria:
This already resets in MessageBroker.js line 1741 on scene change - may be redundant in reset_canvas. It's also only for drawn (point line of sight) light/fog/elev/drawings not token vision.

We also call redraw light walls without walls changing, rechecking all the point line of sight drawings _may_ cause performance issues for people that use this to light/fog/draw a bunch of different areas on their map in those cases. Would have to do some more testing with that Eg. openning the walls menu or toggling always show walls.

Probably as good idea to clear for the different types when drawings are cleared. For example for light drawings in the confirmation in Fog.js line 7051 `vision_menu.find("#delete_light")`. Although we'd probably have to store the type of bucket fill  so we only clear the correct ones when clearing different  draw types.

Valamorde:
Good catches — all three addressed:

1. **reset_canvas redundancy** — the `redraw_light_walls` call inside `reset_canvas` now passes `false`, since the cache is already wiped at the top of `reset_canvas`.

2. **Menu open / non-wall-change calls** — gated the cache clear on the existing `clear` parameter. The wall menu toggle (and the 'always show walls' toggle) now pass `false`, so opening the menu no longer trashes the cache. Same for elevation delete/undo — elevation entries don't touch wall geometry so the cached polygons stay valid.

3. **Per-type clearing** — agreed it would be the ideal long-term approach, but it needs the cache key to encode the drawing type and the deletion handlers to match on it. Happy to follow up in a separate PR if that's worth pursuing.

Azmoria:
Edit: see most recent comment below this as most of the changes I suggest here I made before merge.


> So I might have jumped ahead of myself, I'm little confused what you're trying to address now.
> 
> I'm not sure why you're trying to prevent the canvas from being cleared where you've added `false` for clear. (Unless it was just an oversight due to changes from before and left as false)
> 
> Keeping pixel data in the canvas would just increases memory used I believe. Alternatively would end up drawing the walls on the canvas multiple times, essentially drawing multiple frames of an animation (even if the walls are static). There's only a few cases where we actually want to prevent clearing the canvas, that shows the walls to the DM, from happening but doing other wall/door calculations. Drawing polygon stuff might be the only case for walls where we prevent it currently, unless I'm forgetting something.
> 
> `Mid-scene wall edits — after a wall is added or removed, redraw_light_walls is called but the cache wasn't cleared, meaning a subsequent LOS query could theoretically match a stale entry if wall count happened to be the same`
> 
> This doesn't sound possible, if we are adding or removing a wall the wall the count can't be the same? Unless I'm missing some edge case. 
> 
> When just moving walls rather then remove/adding we can force the cache to be ignored (and do in some cases, such as moving darkness aoe) using `darknessMoved` in `redraw_drawn_light`. If there's a specific case this clears up, where forcing ignoring the cache in redraw_drawn_light isn't being done already, then let me know of course could just be something I'm not seeing. 
> 
> We may want to consider if this is needed for other functions like redraw_elev, fog etc. 
> 
> Currently any time a player gets a message updatings drawing of any kind this will trigger the cache clearing.For that our options would be including if walls changed in draw update messages to prevent that or rely on checking if changes have happened like we are doing now (improve that if needed). 
> 
> Recalculating line of sight is one of the biggest hits to performance that happen. So we want to minimize that as much as possible.

Azmoria:
I did find somethings that wern't working as expected when it came to line of sight tools that this helps. I also have it clear token line of sight cache now, and send if walls have changed with drawing updates. 

We may be able to eventually ignore cache checks if we are always sending correct 'wallsChanged' data. For now I've left both in.

I reworked it a bit to make wallsChanged: false by default and made them all options instead of having to include every var in the function. Added bonus of a bit easier to read what true/false means when used too ( `{wallsChanged: true}` instead of `true,false,true` or `undefined, undefined, true`). 

I commented `// could limit this to point line of sight tool drawings` where we can make future improvements to not redraw all  drawings but just ones affected by walls.

PR #4281, opened 2026-04-08
Fix: use options object for redraw_light_walls in KeypressHandler

Jumbalicious79:
Fixes a merge-coordination issue between #4257 and the `redraw_light_walls` options-object refactoring.

## The Bug

`redraw_light_walls` was refactored from positional parameters `(clear, editingWallPoints, wallsChanged)` to an options object `({clearCanvas, editingWallPoints, wallsChanged})`. Two calls in KeypressHandler.js (Shift+W toggle and Esc key) still use the old positional style:

```js
redraw_light_walls(false, false, false);
```

Passing `false` as the `options` parameter means `...false` spreads to `{}`, so all defaults apply and `clearCanvas` resolves to `true` instead of the intended `false`.

## The Fix

```diff
-        redraw_light_walls(false, false, false);
+        redraw_light_walls({clearCanvas: false});
```

Both call sites (lines 286 and 337) updated to use the new options object syntax, matching all other callers in Fog.js, Token.js, and TokenMenu.js.

## Chrome Verification

Tested on unpacked extension in Chrome 146, DM campaign with 45 walls:

| Test | Before Fix | After Fix |
|---|---|---|
| Shift+W (line 286) | `clearCanvas: true` ❌ | `clearCanvas: false` ✅ |
| Esc with wall tool (line 337) | `clearCanvas: true` ❌ | `clearCanvas: false` ✅ |
| Console errors | 0 | 0 |
| Wall display/hide | works | works |

Verified by instrumenting `redraw_light_walls` to log resolved parameter values.

## Files Changed

- `KeypressHandler.js` — 2 lines changed (lines 286, 337)

### DIFF
--- a/KeypressHandler.js
+++ b/KeypressHandler.js
@@ -283,7 +283,7 @@ Mousetrap.bind('w', function () {
 Mousetrap.bind('shift+w', function () {
     if(window.DM){
         $('#show_walls').toggleClass(['button-enabled', 'ddbc-tab-options__header-heading--is-active']);
-        redraw_light_walls(false, false, false);
+        redraw_light_walls();
     }
 
 });
@@ -334,7 +334,7 @@ Mousetrap.bind('esc', function () {     //deselect all buttons
         $('#select-button').click();
     }
     else{
-        redraw_light_walls(false, false, false);
+        redraw_light_walls();
     }
 
     close_token_context_menu();


Azmoria:
Oops, I guess my KeypressHandler file didn't save due to a conflict and I missed that.

These should just use default values though (true, false, false) so just `redraw_light_walls()` rather then passing in `clearCanvas:false`. We want it clearing the canvas before drawing walls, or when hiding them. 

In most cases we will want the canvas cleared. It's just the visual the DM sees as placeholders for the actual walls that get cleared. Think of it as clearing the frame before draing the next frame of animation, so we aren't double drawing everything or having walls stay that no longer exist etc. If it's not displayed it's just less memory having an empty canvas. Even if we are clearing an already empty canvas that's ok. I think only for polygon drawings do we not clear iirc.

Jumbalicious79:
Updated — both call sites now call `redraw_light_walls()` with no arguments so it uses the defaults (clearCanvas: true). Makes sense that we want the canvas cleared before redrawing. Thanks for the explanation!

PR #4284, opened 2026-04-10
Fix: stop scene loading in reset_canvas when scene map has 0 dimensions

### LINKED ISSUE #4264: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is an OffscreenCanvas element with a width or height of 0.
**Error:**
```
Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is an OffscreenCanvas element with a width or height of 0.

InvalidStateError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D': The image argument is an OffscreenCanvas element with a width or height of 0.
    at redraw_drawn_light (chrome-extension://ejmmfaehjmcdbkganppofnhmifkcokoa/Fog.js:2136:11)
    at reset_canvas (chrome-extension://ejmmfaehjmcdbkganppofnhmifkcokoa/Fog.js:1452:2)
    at MessageBroker.handleScene (chrome-extension://ejmmfaehjmcdbkganppofnhmifkcokoa/MessageBroker.js:1725:11)
```
**Environment:**
```
{"avttVersion":"1.53-local","browser":{"name":"Chrome","version":"146","mozilla":false,"chrome":true,"msie":false,"opera":false,"safari":false}}
```
**Other Logs:**
```

WARN 2026-04-07T03:37:22.434Z
      Attempted draw without scene being loaded or missing image
      

LOG 2026-04-07T03:37:22.432Z
      WEATHERSIZE
      0
      0

LOG 2026-04-07T03:37:20.673Z
      edit_scene_dialog

LOG 2026-04-07T03:37:20.672Z
      did_click_row_gear
      {"originalEvent":{"isTrusted":true},"type":"click","target":{},"currentTarget":{"jQuery360057732293085924691":{"events":{"click":[{"type":"click","origType":"click","guid":283,"namespace":""}]}}},"relatedTarget":null,"timeStamp":1292217.5,"jQuery36005773229308592469":true,"delegateTarget":{"jQuery360057732293085924691":{"events":{"click":[{"type":"click","origType":"click","guid":283,"namespace":""}]}}},"handleObj":{"type":"click","origType":"click","guid":283,"namespace":""}}

LOG 2026-04-07T03:37:20.672Z
      find_sidebar_list_item
      {"id":"05ef2991-2e88-49b4-990b-cfab0f51b1f0","name":"New Scene","image":"","type":"scene","folderPath":"/Scenes","parentId":"scenesFolder","isVideo":false,"isUvtt":false}

LOG 2026-04-07T03:37:20.672Z
      did_click_row_gear
      {"id":"05ef2991-2e88-49b4-990b-cfab0f51b1f0","name":"New Scene","image":"","type":"scene","folderPath":"/Scenes","parentId":"scenesFolder
```

Jumbalicious79:
Fixes #4264

## The Bug

When a scene has no image (empty map URL), `reset_canvas` receives 0x0 dimensions from `#scene_map`. This cascades into `redraw_drawn_light()` → `new OffscreenCanvas(0, 0)` → `drawImage` crash:

```
InvalidStateError: Failed to execute 'drawImage' on 'CanvasRenderingContext2D':
The image argument is an OffscreenCanvas element with a width or height of 0.
```

The crash path: DM creates/switches to empty scene → edits settings (e.g. enables Fog of War) → `handleScene` takes the `isSameScaleAndMaps` fast path → `reset_canvas(false)` → crash.

PR #4272 (merged) guards the crash site in `redraw_drawn_light`, but as noted, that leaves the scene in a broken state where drawn light never renders.

## The Fix

Early return in `reset_canvas` when `sceneMapWidth` or `sceneMapHeight` is 0:
- Shows `showErrorMessage('No scene map link found. Please check your scene settings')`
- Cleans up loading state (`set_default_vttwrapper_size`, loading overlays, `window.LOADING`)
- Returns before any canvas operations

## Why This Works

Catching at `reset_canvas` prevents all downstream draw functions (`redraw_light_walls`, `redraw_drawings`, `redraw_drawn_light`, `redraw_fog`, `redraw_elev`) from operating on 0-dimension canvases. The loading cleanup prevents the UI from getting stuck if this triggers during initial scene load.

Follows Azmoria's guidance in #4264: catch in `reset_canvas`, show error, stop loading to prevent downstream vision check errors.

## Chrome Verified

1. Created empty scene (no image), enabled Fog of War, saved → error banner with friendly message, no crash
2. Switched to scene with image → loaded normally, no errors
3. Console: zero `drawImage` errors, only our `showErrorMessage` logged

## Files Changed

- `Fog.js` — 10-line guard in `reset_canvas`

### DIFF
--- a/Fog.js
+++ b/Fog.js
@@ -1426,6 +1426,16 @@ function reset_canvas(apply_zoom=true) {
 	const sceneMapWidth = $("#scene_map").width();
 	const sceneMapHeight = $("#scene_map").height();
 
+	if (sceneMapWidth === 0 || sceneMapHeight === 0) {
+		showErrorMessage('No scene map link found. Please check your scene settings');
+		set_default_vttwrapper_size();
+		$('#loadingStyles').remove();
+		$('.import-loading-indicator').remove();
+		remove_loading_overlay();
+		delete window.LOADING;
+		return;
+	}
+
 	$('#darkness_layer').css({"width": sceneMapWidth, "height": sceneMapHeight});
 	$("#scene_map_container").css({"width": sceneMapWidth, "height": sceneMapHeight});
 	// grid overlay css tiling needs a container to fill that matches map
--- a/Fog.js
+++ b/Fog.js
@@ -1426,14 +1426,14 @@ function reset_canvas(apply_zoom=true) {
 	const sceneMapWidth = $("#scene_map").width();
 	const sceneMapHeight = $("#scene_map").height();
 
-	if (sceneMapWidth === 0 || sceneMapHeight === 0) {
-		showErrorMessage('No scene map link found. Please check your scene settings');
+	if (!sceneMapWidth || !sceneMapHeight) {
+		showErrorMessage("Error loading scene.", `<p>Possible issues:</p>• The scene map link may be blank<br>• If using a video map ensure the "video map" toggle is enabled<br>• The file may not be publicly accessible (check share settings on host)<br>• The host may have rate limits on file access`);
 		set_default_vttwrapper_size();
 		$('#loadingStyles').remove();
 		$('.import-loading-indicator').remove();
 		remove_loading_overlay();
 		delete window.LOADING;
-		return;
+		return false;
 	}
 
 	$('#darkness_layer').css({"width": sceneMapWidth, "height": sceneMapHeight});
@@ -1464,7 +1464,7 @@ function reset_canvas(apply_zoom=true) {
 	if (!window.FOG_OF_WAR) {
 		const canvas = document.getElementById("fog_overlay");
 		canvas.getContext("2d").clearRect(0, 0, canvas.width, canvas.height);
-		return;
+		return true;
 	}
 
 	setVisionLightOffscreenCanvas(); //adding this here as it was previously being done every fill causing webgl issues
@@ -1501,7 +1501,7 @@ function reset_canvas(apply_zoom=true) {
 	if(apply_zoom)
 		apply_zoom_from_storage();
 	redraw_text();
-	return;
+	return true;
 }
 function check_darkness_value(){
 	
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -1716,7 +1716,8 @@ class MessageBroker {
 				$("#VTT").css("--scene-scale", scaleFactor)
 				window.CURRENT_SCENE_DATA.width = mapWidth;
 				window.CURRENT_SCENE_DATA.height = mapHeight;
-				await reset_canvas(false);
+				const continueLoading = await reset_canvas(false);
+				if(!continueLoading) return;
 
 				if(!isSameTokenLight){
 					for(let i in window.TOKEN_OBJECTS){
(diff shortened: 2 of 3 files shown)


Azmoria:
Currently I believe this will still continue running scene loads after `resert_canvas()` is called. I think to stop that you'll have to adjust this to return false (or some other value), then in places where we call `reset_canvas()` to be something like

```
const continueLoading = await reset_canvas();
if(!continueLoading) return;
```

Probably have to return true at the end of reset_canvas to ensure it doesn't stop otherwise.

Also thinking on it more, this doesn't specifically mean there is no map link, so we may want to change the error message a bit to encompass more possible issues/things to check for. Can add something else if you think there's other cases I'm missing.

```
showErrorMessage("Error loading scene.", `<p>Possible issues:</p>• The scene map link may be blank<br>• If using a video map ensure the "video map" toggle is enabled<br>• The file may not be publicly accessible (check share settings on host)<br>• The host may have rate limits on file access`)
```

Gives a bit better info 

image

Jumbalicious79:
Thanks for the review @Azmoria. Pushed [a9ccb39e](https://github.com/cyruzzo/AboveVTT/pull/4284/commits/a9ccb39e) addressing both points:

**1. Bailout now actually halts loading.** `reset_canvas()` returns `false` on the 0-dimension bailout and `true` on the success paths. All three callers check the return value and bail:

- `MessageBroker.handleScene` path 1 — same-scene refresh (line 1719 on main): bare `return` (LOADING already cleared earlier in this branch)
- `MessageBroker.handleScene` path 2 — fresh-load `load_scenemap` callback (line 1911): closes the open `console.group("load_scenemap callback")`, calls `window.MB.loadNextScene()` to drain any queued scenes (matching the success path at line 1999 and the catch path at 2007), then returns
- `ScenesHandler.switch_scene` grid wizard callback: bare `return`

**2. Updated error message** per your exact wording — `showErrorMessage("Error loading scene.", \`<p>Possible issues:</p>...\`)` covering all four causes (blank link / video map toggle / file accessibility / host rate limits).

**Bonus tightening:** I also changed the guard from `=== 0` to `!sceneMapWidth || !sceneMapHeight` so it catches `undefined` too — when `#scene_map` is removed by `load_scenemap` line 481 before the new element is appended, jQuery's `.width()` returns `undefined`, not `0`. End-to-end testing surfaced this case (see below).

**Chrome verification of the original F-1 stack trace:**

Repro setup:
1. Have a scene with empty `player_map` and `dm_map` URLs (created via the scenes panel)
2. DM switch to that scene → triggers the existing `confirm()` dialog from `Main.js:444 map_load_error_cb`, click Cancel
3. State after: `$("#scene_map")` is the new empty ``, width=0, height=0, `CURRENT_SCENE_DATA.id` set to the broken scene
4. Send `switch_scene` for the same scene id again — `handleScene` takes the `isSameScaleAndMaps` path 1 at line 1719

Pre-fix: `Failed to execute 'drawImage' on 'CanvasRenderingContext2D'... InvalidStateError` from `redraw_drawn_light` (originally — `redraw_fog` after [#4272](https://github.com/cyruzzo/AboveVTT/pull/4272) was merged, but the same drawImage-on-zero-canvas root cause).

Post-fix: bailout fires, `showErrorMessage` displays the new banner with all four bullets, `window.LOADING` cleared, no crash, recovery to a working scene unaffected.

Also smoke-tested directly: `reset_canvas()` returns `false` when `#scene_map` is removed and `true` when called against a valid scene map — both via interactive console.

Let me know if you want anything tweaked.

PR #4289, opened 2026-04-11
Update character token when buff/debuff selected from dropdown

lzenner:


CharactersPage.js
```diff
@@ -1510,6 +1520,26 @@ function rebuild_buffs(fullBuild = false){
...
+		if(buffsDebuffs[i].condition != undefined) { // Allow buffsDebuffs with conditions to update player tokens
+			let setOnOff = 'removeCondition';
+			let condition = buffsDebuffs[i].condition;
+			if($(this).val() != '0'){
+				setOnOff = 'addCondition';
+			}
+			if (is_abovevtt_page()) {
+				const pc = find_pc_by_player_id(window.PLAYER_ID, false);
+				const token = window.all_token_objects[pc.sheet];
+				token[setOnOff](condition);
```
Azmoria, line 1534: If the player token hasn't been seen/loaded at least once yet this will throw a cannot find undefined token error when going to run the add/remove condition function. Probably best to do something like `if(!token) return;` just before that. _Possibly_ worth catching if pc doesn't exist either in cases of pcs from outside the campaign as pc.sheet could error too.

Same in the other spots you're looking  for the pc object and token.

Startup.mjs
```diff
@@ -137,6 +137,16 @@ $(function() {
         enable_dice_streaming_feature(window.JOINTHEDICESTREAM);
        
         tabCommunicationChannel.addEventListener ('message', (event) => {
+		  if(event.data.msgType == 'addCondition' || event.data.msgType == 'removeCondition'){ // Sets a player token's condition on and off
```
Azmoria, line 140: These don't need to be dealt with before merge but something I can look into before release to see if they should be caught and accounted for

Edge cases:

1) This will run in all avtt tabs that are open. Probably can limit this just to the `sendTo` id which looks to match `window.PLAYER_ID` in other messages here. This may not be a common issue unless the DM is running a pc and has spectator open or something.

2) For official conditions, any abovevtt window that isn't the player sheet your sending the message from will send a message to chat rather then setting the condition.

   It may be best to deal with official conditions directly on the sheet by openning the tab and adjusting instead of sending a message to other tabs. Conditions will auto send a message to update when toggled.

   This probably affects people playing multiple pcs most. They'll probably have 1 avtt window open and use other tabs for their characters. They'll only be able to update the character condition for the one  logged into avtt the way it currently is. Same for DMs accessing player sheets (although also rare to adjust roll buffs as the DM unless playing that character themselves)

Azmoria (commented):
Looks good seems to function as expected, one minor change to catch undefined `token` variables if they haven't been on a scene with the player token yet. Similar  with the `pc` variable if the character is from outside the campaign. I can merge after that.

Also noted some edge cases I can look into before release.

PR #4299, opened 2026-04-16
QoL - scroll wheel to adjust HP and max HP in combat tracker

### LINKED ISSUE #2026: Request: Add arrows and scroll event listener for hp/max hp inputs.
Request: Add arrows and scroll event listener for hp/max hp inputs.

Changing these to number inputs won't work as it prevents the +/- math from working. So we'll have to add our own buttons. While doing this we may as well add a scroll listener while the input is focused as well.

Valamorde:
Closes #2026

## What

Adds scroll-wheel increment/decrement to the HP and max HP inputs in the combat tracker.

- Scroll up → +1
- Scroll down → −1
- HP clamps at 0; max HP clamps at 1
- Prevents the tracker list from scrolling while the cursor is over an input (via `e.preventDefault()`)

## How

Adds two `wheel` event listeners inside the existing `!token.isPlayer()` block (lines 1355–1368), alongside the existing `change`/`click`/`keydown` handlers. Uses `token.hp` / `token.maxHp` as the source of truth so a partially-typed `+5` expression in the input field doesn't cause an off-by-one, and triggers the existing `change` event so token sync, broadcast, and HP-bar updates all fire unchanged.

Player tokens are excluded — their inputs are already `disabled` and the listeners live in the `!isPlayer()` branch, so no regression there.

No reformatting, no new dependencies, backward compatible.

### DIFF
--- a/CombatTracker.js
+++ b/CombatTracker.js
@@ -1352,6 +1352,20 @@ function ct_add_token(token,persist=true,disablerolling=false, adv=false, dis=fa
 		maxhp_input.click(function(e) {
 			$(e.target).select();
 		});
+		hp_input.on('wheel', function(e) {
+			e.preventDefault();
+			const delta = e.originalEvent.deltaY < 0 ? 1 : -1;
+			const current = parseInt(token.hp) || 0;
+			$(this).val(Math.max(0, current + delta));
+			$(this).trigger('change');
+		});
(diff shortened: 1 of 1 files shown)


Azmoria:
🍻 works well, going to merge as is

We should watch to see if people want to have to focus the input before scrolling. In case they accidentally have it hovered. Not sure this will be an issue.

Issue #4288, opened 2026-04-11
Replace blocking confirm() in map_load_error_cb with non-blocking banner — wanted?


Jumbalicious79:
@Azmoria — found a small UX issue while Chrome-testing #4284 yesterday and wanted to check before sending a PR, since it's not a user-reported bug and there's a trade-off worth your input on.

## What I found

`Main.js:434` `map_load_error_cb` — the callback wired to the `<img>` `error` event for scene maps (lines 559 and 610). When a map URL fails to load, the generic case fires **two nested native `confirm()` dialogs** at lines 444-445:

```js
else if (confirm(`Map could not be loaded!\n${specificMessage}\nYou may also need to disable ad blockers.\nWould you like to try loading the image in a separate tab to verify that it's accessible? If you are currently logged in to google, you will need to log out or open the image in a different browser or an incognito window to truly test it.`)) {
    if (window.DM || confirm(`SPOILER ALERT!!!\nIf you click OK, you might see the entire map without fog of war. However, the map isn't loading at all so you will probably see a broken link. Are you sure you want to test this image?`)) {
        window.open(window.CURRENT_SCENE_DATA.map, '_blank');
    }
}
```

Native `confirm()` blocks the entire JS event loop until dismissed — renderer freezes, WS processing pauses, queued scenes can't drain, and automation/CDP sessions hang. I tripped over this myself when I created an empty-map scene to reproduce #4264.

The Drive path right above it (line 441-442) already uses the non-blocking `showGoogleDriveWarning()`, and your `peerVideo.js` `8d47c07a` did the same thing for missing media devices. So the precedent and the helper are both already in place — just the generic path didn't get migrated.

## Triggering scenarios

Any DM with a broken map URL: pasted link with a typo, Drive share link revoked but without `drive.google` in the rewritten URL, host returns 404/500/rate-limited, ad/content blocker on the host, imported scene file with a dead URL, request timeout, etc. Not common but not edge-case either.

## Proposed fix (already written + Chrome-verified locally)

Replace the nested `confirm()` block with a single `showErrorMessage()` call — same primitive you had me use in #4284. The Drive special case stays untouched, the always-runs cleanup at lines 447-451 (`delete window.LOADING` / `remove_loading_overlay` / `loadNextScene` / `console.groupEnd`) stays untouched. Net diff: **−6/+3 lines, one file**.

```diff
 function map_load_error_cb(e) {
   console.log(e);
[...]

PR #4317, opened 2026-04-23
Fix: replace blocking alert() with non-blocking banner/error helpers

### LINKED ISSUE #4298: Replace blocking alert() calls in avttS3Upload.js with non-blocking alternatives
## Context

Per @Azmoria's feedback on #4288 (comment), the `alert()` calls in `avttS3Upload.js` could use better error messaging. Some should be `showErrorMessage()`, others should be `showTempMessage()` — depending on whether they're errors vs informational messages.

## Scope

**33 `alert()` calls** in `avttS3Upload.js`. They fall into roughly 4 categories:

### 1. Informational messages → `showTempMessage()`
These aren't errors — they're status confirmations that don't need a blocking dialog:
- Line 2008: `"Migration (hopefully) completed. You need to Re-Join AboveVTT"` — Azmoria noted this is no longer for local→server migration, now used for imports. Could improve the copy and potentially avoid requiring a full reload.

### 2. User input validation → `showTempMessage()` or inline
Quick validation feedback that doesn't need to freeze the renderer:
- Line 3574: `"Cannot paste a folder inside itself."`
- Line 3627: `"Names cannot contain slashes."`
- Line 3636: `"Cannot rename a folder into its own sub-path."`
- Line 3658: `"An item with that name already exists..."`
- Line 3814: `"Cannot move a folder inside itself."`
- Line 4763: `"Patreon login is required..."`
- Line 5861/5865/5870: Folder name validation
- Line 6351: Storage limit exceeded

### 3. Error messages → `showErrorMessage()`
Actual errors where users may need to report the issue or see details:
- Line 268: Upload error with `json.error` + `json.message`
- Line 1831/1850/1872/2303/2332/2364: `"Failed to open the file in a new tab"`
- Line 3534: `"Failed to move item(s)."`
- Line 5406/5423: Export failures
- Line 5769: Upload dropped files error
- Line 5890: `"Failed to create folder"`
- Line 6578: Generic error
- Line 7260/7268/7281: Import failures
- Line 7415: `"Error fetching folder listing..."`
- Line 7882: `"Failed to delete file(s)."`

### 4. Unique name generation failures → `showTempMessage()`
- Lines 1456, 1536, 6330: `"Failed to generate a unique name. Skipping this item."`

## Also in other files

Azmoria also mentioned:
- `CoreFunctions.js:689`: `"This campaigns local explored vision data has been cleared."` → `showTempMessage()`
- `Settings.js:2166`: Same migration alert as avttS3Upload.js line 2008
- `MessageBroker.js:2503`: Long `alert()` for max message size — may need the scrollable `showErrorMessage` container from PR #4295

## Approach

Happy to tackle this in batches (one PR per category, or one PR for the whole file) — whatever scope works best. The `showTempMessage()` and `showErrorMessage()` primitives are already available and well-established in the codebase.

Jumbalicious79:
Fixes #4298

## The Bug

35 `alert()` calls across 3 files fire blocking native dialogs that freeze the renderer, both WebSockets, and the scene-queue drain while the user dismisses them. Same class of problem as #4288/#4295 (`confirm()` in `map_load_error_cb`), applied to the rest of the codebase per @Azmoria's guidance in that thread.

## The Fix

Mechanical swap — `alert(X)` → `showTempMessage(X)` or `showErrorMessage(X)`, preserving arg shape. No other code changes, no reformatting.

### Files changed
| File | Count | Helper |
|---|---|---|
| `avttS3Upload.js` | 13 | `showTempMessage` (validation + skip notices) |
| `avttS3Upload.js` | 19 | `showErrorMessage` (catch-block failures) |
| `CoreFunctions.js` | 2 | `showTempMessage` (explored-data cleared) |
| `MessageBroker.js` | 1 | `showErrorMessage` (max message size) |
| **Total** | **35** | |

### One non-mechanical change
`avttS3Upload.js:268` (HTTP 403 handler) is split from `alert(\`${json.error}\n\s${json.message}\`)` into `showErrorMessage(json.error, json.message)`. The original `\n\s` sequence renders as a literal `s` character because `\s` has no special meaning in JS template strings; preserving it in `showErrorMessage` would produce mangled HTML output (`\n` collapses in HTML rendering). Splitting into `(error, extraInfo)` matches the helper's API and naturally resolves the formatting issue. Azmoria's approved plan categorized this as `showErrorMessage`.

### Intentional out-of-scope
- `avttS3Upload.js:2008` and `Settings.js:2166` — migration alert with `location.reload()`. Per #4288 discussion this could be a separate QoL update to improve copy and avoid the full reload.
- `MessageBroker.js:1078, 1257` — AboveVTT ≤ 0.0.51 / ≤ 0.64 version-warning alerts for players joining with ancient versions. Not part of #4298 scope; intentionally disruptive by design.

## Why This Works

Both `showTempMessage` (`CoreFunctions.js:1674`) and `showErrorMessage` (`CoreFunctions.js:1477`) are well-established helpers used throughout the codebase (`Settings.js`, `Startup.mjs`, `Token.js`, `CombatTracker.js`, `KeypressHandler.js`, `MessageBroker.js`, `TokenMenu.js`, `Main.js`) and accept the same arg shapes the old `alert()` calls used. `showErrorMessage` also logs `console.error(error, ...extraInfo)` internally, so errors continue to reach `addLog`/`concerningLogs` via the hooked console.

`showErrorMessage` benefits from the scrollable `max-height: 200px; overflow-y: auto` container added in PR #4295 — handles the long dynamic alert at `MessageBroker.js:2509` (drawings count + advice) without truncation.

## Chrome Verification

Reproduced and fix-verified on the unpacked build of this branch.

**Test Environment**
- Chrome 147.0.0.0 on macOS (MacIntel)
- AboveVTT 1.53 (unpacked, Web Store version disabled)
- DM tab, scene \"The Tavern\", 21 tokens, both WebSockets OPEN

**Pre-fix** (upstream/main)
- Source inspection confirmed `alert('...')` at the 35 target lines.
- `alert()` behavior is a JS primitive that blocks the renderer until dismissed — behavior is well-understood, not triggered via MCP automation because native dialogs freeze MCP commands.

**Post-fix** (this branch, extension reloaded)
1. Called `deleteExploredScene(window.CURRENT_SCENE_DATA.id)` directly, exercising the patched `CoreFunctions.js:710` `showTempMessage` path (bypasses the preceding `confirm()` which blocks MCP). Installed a spy on `window.alert` to catch any remaining blocking dialog.
2. Result: `.abovevttTempMessage` banner appeared with text \"Scene Explore Trail Data Cleared\", visible at t=100ms, 500ms, 1500ms, auto-faded after. `alertsIntercepted=0`. `concerningLogs` stayed at 0. Both WebSockets remained OPEN. Token count unchanged (21). Scene unchanged.
3. Additional positive check on `showErrorMessage` with the three arg shapes this PR uses: plain string, `error?.message || \"fallback\"`, and `(error, extraInfo)` split — all three rendered correctly into the `#above-vtt-error-message` dialog with `h3#error-message` and `#error-message-details` populated as expected.

**Code analysis** for the 32 `avttS3Upload.js` conversions. Those code paths live inside the Patreon-gated file manager UI (not testable without a Patreon account). Every conversion is a 1:1 mechanical swap using the same helpers validated end-to-end via the `CoreFunctions.js` and `showErrorMessage` probe tests above.

**Happy path / regression**
- No lingering UI artifacts after test.
- Token state, scene state, and both WebSockets preserved throughout.
- 0 `concerningLogs` before and after.

## Follow-up

Per @Azmoria's 2026-04-20 comment on #4298, several of the new `showErrorMessage` sites could benefit from \"possible causes\" extra-info text (modeled on the scenes-error pattern from #4295) — planned as a separate follow-up once this lands. Some may be reclassified as informational `showTempMessage` depending on whether the user can resolve them.

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -686,7 +686,7 @@ function deleteDB(){
     objectStoreRequest.onsuccess = function(event) {       
       $('#exploredCanvas').remove();
       redraw_light();
-      alert('This campaigns local explored vision data has been cleared.')
+      showTempMessage('This campaigns local explored vision data has been cleared.')
     };
   }
 }
@@ -707,7 +707,7 @@ function deleteExploredScene(sceneId){
       if(sceneId == window.CURRENT_SCENE_DATA.id){
         $('#exploredCanvas').remove();
         redraw_light();
-        alert('Scene Explore Trail Data Cleared')
+        showTempMessage('Scene Explore Trail Data Cleared')
       }        
     };
 }
--- a/MessageBroker.js
+++ b/MessageBroker.js
@@ -2506,7 +2506,7 @@ class MessageBroker {
 				alertText = 'Check console warnings for more message data.'
 			}
 			
-			alert(`You reached the maximum message size for "${message.eventType.split('/')[message.eventType.split('/').length-1]}".\n\n${alertText}`);
+			showErrorMessage(`You reached the maximum message size for "${message.eventType.split('/')[message.eventType.split('/').length-1]}".\n\n${alertText}`);
 			return;
 		}
 
--- a/avttS3Upload.js
+++ b/avttS3Upload.js
@@ -265,7 +265,7 @@ async function avttFetchWithRetry(input, init = {}, options = {}) {
         if(response.status == 403){
           const json = await response.json();
           if (json.error && json.message)
-            alert(`${json.error}\n\s${json.message}`)
+            showErrorMessage(json.error, json.message)
         }
         
         return response;
@@ -1453,7 +1453,7 @@ async function avttResolveFolderDescendantConflicts(move, targetRootKey, conflic
           targetKey,
           deriveError,
         );
-        alert("Failed to generate a unique name. Skipping this item.");
+        showTempMessage("Failed to generate a unique name. Skipping this item.");
         skipDescendants.add(sourceKey);
         continue;
       }
@@ -1533,7 +1533,7 @@ async function avttResolveMoveConflicts(moves) {
         targetKey = await avttDeriveUniqueKey(targetKey);
       } catch (deriveError) {
         console.error("Failed to generate unique destination during move", deriveError);
-        alert("Failed to generate a unique name. Skipping this item.");
+        showTempMessage("Failed to generate a unique name. Skipping this item.");
         continue;
       }
     }
@@ -1828,7 +1828,7 @@ async function avttHandleToolbarAction(action) {
 
       } catch (error) {
         console.error("Failed to open file in new tab", error);
-        alert(error?.message || "Failed to open the file in a new tab.");
+        showErrorMessage(error?.message || "Failed to open the file in a new tab.");
       }
       handled = true;
       break;
@@ -1847,7 +1847,7 @@ async function avttHandleToolbarAction(action) {
         }
       } catch (error) {
         console.error("Failed to open file in new tab", error);
-        alert(error?.message || "Failed to open the file in a new tab.");
+        showErrorMessage(error?.message || "Failed to open the file in a new tab.");
       }
       handled = true;
       break;
@@ -1869,7 +1869,7 @@ async function avttHandleToolbarAction(action) {
         }
       } catch (error) {
         console.error("Failed to open file in new tab", error);
-        alert(error?.message || "Failed to open the file in a new tab.");
+        showErrorMessage(error?.message || "Failed to open the file in a new tab.");
       }
       handled = true;
       break;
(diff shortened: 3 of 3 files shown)


Azmoria:
Looks good, will merge in the next day or two now that release is out and there hasn't been a flood of reports. 

Theres one or two I want to check, like the proxy one, I _think_ thats the one where I want to add more info on some sites security settings. There might be 2 different error spots for this though - one that goes through the proxy and one that doesn't. Maybe just have it be showTempMessage too since it would be the one we can't control. Have to try with a few sites to see if its the error I'm thinking of anyway.

We can add more details like that after merge though.

PR #4324, opened 2026-04-25
Campaign Export Reminder and previous link Scene Recovery

jeffsenn:
Added Setting for an export reminder.
Current runs when a scene finishes loading.

Todo (if we like this feature): 
-- Visual cleanup of dialog
-- Better occasional triggering of check

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -2202,6 +2202,13 @@ function store_campaign_info() {
   const campaignSecret = window.CAMPAIGN_SECRET;
   if (typeof campaignId !== "string" || campaignId.length <= 0) return;
   if (typeof campaignSecret !== "string" || campaignSecret.length <= 0) return;
+  // save all previous compaign secrets for later restoration
+  const previous = read_campaign_info(campaignId);
+  if(previous) {
+    if(previous === campaignSecret) return; //short circuit - no need to write
+    const previousPrevious = localStorage.getItem(`AVTT-CampaignInfo-${campaignId}-previous`) || "";
+    localStorage.setItem(`AVTT-CampaignInfo-${campaignId}-previous`, previousPrevious ? `${previous},${previousPrevious}` : previous);
+  }
   localStorage.setItem(`AVTT-CampaignInfo-${campaignId}`, campaignSecret);
 }
 
@@ -2217,6 +2224,7 @@ function read_campaign_info(campaignId) {
 /** @param {string} campaignId the DDB id of the campaign */
 function remove_campaign_info(campaignId) {
   localStorage.removeItem(`AVTT-CampaignInfo-${campaignId}`);
+  //todo: remove -previous as well?
 }
 
 // Low res thumbnails have the form https://www.dndbeyond.com/avatars/thumbnails/17/212/60/60/636377840850178381.jpeg
--- a/Main.js
+++ b/Main.js
@@ -463,6 +463,8 @@ function remove_loading_overlay() {
 	$("#loading_overlay").animate({ "opacity": 0 }, 1000, function() {
 		$("#loading_overlay").hide();
 	});
+	//todo: find a better place
+	checkForExportRemind();
 }
 
 /**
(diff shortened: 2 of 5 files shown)


CoreFunctions.js
```diff
@@ -2202,6 +2202,13 @@ function store_campaign_info() {
...
   if (typeof campaignId !== "string" || campaignId.length <= 0) return;
   if (typeof campaignSecret !== "string" || campaignSecret.length <= 0) return;
+  // save all previous compaign secrets for later restoration
+  const previous = read_campaign_info(campaignId);
+  if(previous) {
+    if(previous === campaignSecret) return; //short circuit - no need to write
+    const previousPrevious = localStorage.getItem(`AVTT-CampaignInfo-${campaignId}-previous`) || "";
+    localStorage.setItem(`AVTT-CampaignInfo-${campaignId}-previous`, previousPrevious ? `${previous},${previousPrevious}` : previous);
+  }
   localStorage.setItem(`AVTT-CampaignInfo-${campaignId}`, campaignSecret);
```
Azmoria, line 2212: should this be above the `if(previous)` due to the return in there?

edit: I guess it'll already be saved if the previous = current though. Assuming nothing else cleared it - which I don't know what would clear 1 but not the other.

Main.js
```diff
@@ -463,6 +463,8 @@ function remove_loading_overlay() {
 	$("#loading_overlay").animate({ "opacity": 0 }, 1000, function() {
 		$("#loading_overlay").hide();
 	});
+	//todo: find a better place
+	checkForExportRemind();
```
Azmoria, line 467: maybe at the end of  `start_above_vtt_for_dm()` at `Startup.mjs:469`?

CoreFunctions.js
```diff
@@ -2217,6 +2224,7 @@ function read_campaign_info(campaignId) {
 /** @param {string} campaignId the DDB id of the campaign */
 function remove_campaign_info(campaignId) {
   localStorage.removeItem(`AVTT-CampaignInfo-${campaignId}`);
+  //todo: remove -previous as well?
```
Azmoria, line 2227: Doesn't look  like we currently use  this function. It may be worth a look over to see if we should be clearing local data on campaign deleting etc in another PR though. I _thought_ we already did some of this but we may not be clearing everything.

Settings.js
```diff
@@ -2141,6 +2235,29 @@ function import_process_datafile_text(fileText) {
 	return DataFile;
 }
 
+function attemptRecovery(campaignSecret) {
+	AboveApi.exportScenes(campaignSecret).then((scenes) => {
+		if(scenes && scenes.length > 0) {
+			//do we really want/need migrate here? or is there a better way?
```
Azmoria, line 2241: Yeah, migrateScenes originally was named for migrate from local -> cloud but we used it for all cloud scene creation/importing. So only way to do it - looks correct for export -> import 👍

Azmoria (commented):
Looks good 👍. Recovery should work well, will give it a full test with my test campaign before merge. I'll test the daily/weekly export options in beta after merge tbh (I'll assume monthly works if they do looking at it lol) - looks like they should be fine though.

A couple minor comments/answering comments

CoreFunctions.js
jeffsenn: I guess - just leave comment for now.

Main.js
jeffsenn: Yeah - I'm not sure how often a DM would leave a tab open "forever" without a reload .... ie would want it to check every once in awhile ... I guess could use setTimeout for that?  Any better thoughts?

CoreFunctions.js
jeffsenn: Yes. that last. will leave it this way unless you think otherwise.

Main.js
Azmoria: I think a message broker failure might force a refresh at some point but I see the idea. Could check on the ping/pong message avtt sends every 8 mins or so if we want to attach to an existing interval instead of having multiple going. In `setupMBIntervals` `MessageBroker.js:223`

CoreFunctions.js
Azmoria: sounds good

Main.js
jeffsenn: Just what I was looking for.  Good Idea - I'll implement...

PR #4417, opened 2026-06-21
QoL - allow scroll wheel to adjust hp on tokens; Fix - issue with scroll wheel adjusting hp in combat tracker if the input isn't clicked into

Azmoria:


dbatiste:
🎉  Thanks for fixing this!!!

PR #4488, opened 2026-07-15
Added option to use mouse wheel as zoomin/zoomout

kvegeta:
Added an option to enable zooming with the mouse wheel without holding Ctrl, making one-handed use easier (e.g. when using AboveVTT on a tabletop without a keyboard).
Zoom In/Out buttons are now hidden when the option is enabled.


Azmoria:
This currently prevents scrolling anywhere else there's a scrollable element. When `wheelZoomAsToggle` is true, you'll probably want to make sure the hovered element is `#black_layer` or inside `#VTT`.

I'll have to try some other devices too as this may interfere with touch pad. Often using gestures that send wheel events instead of right click drag. It might be ok anyway since it'd be optional.

kvegeta:
it's maybe a good way to have the option only on desktop view/device? (that don't prevent as touchscreen input anyway)

Azmoria:
Yeah, probably don't have to hide the setting but could ignore the setting if it's a scroll touchpad gesture that triggers a 'wheel' event.

kvegeta:
Hi Azmoria, any news for the release? :-D

Azmoria:
Sorry I was waiting for changes mentioned in my first comment. Since people do stuff in their free time I don't generally push for quick changes.

The scrolling doesn't work for anything except the map with this setting enabled. It's prevented in the wheel event listener if  `wheelZoomAsToggle` is true. Will need to make sure the scene is hovered rather then just `wheelZoomAsToggle` being true. 

There are elements within the scene that also have wheel listeners like hp so it can't just be hovering anything in the scene. _Maybe_ non-focused elements as I believe all elements in the scene itself that have a scroll bar do require focus iirc.

Unless some ddb changes happen or major bug is reported, _probably_ at least two weeks out from another live release if you wanted to get it in before then.

kvegeta:
no problem and thanks for the time ;-)
I can use it as local extension in the meantime ;-)
Anyway, let me know if there’s anything I can do to speed up the process.
Thanks for the time!

Main.js
```diff
@@ -1277,7 +1277,8 @@ function init_mouse_zoom() {
 	if(window.mouseZoomInitialized) return;
 	window.mouseZoomInitialized = true;
 	window.addEventListener('wheel', function (e) {
-		if (e.ctrlKey) {
+		const wheelZoomAsToggle = get_avtt_setting_value?.('wheelZoomAsToggle') === true;
+		if (e.ctrlKey || wheelZoomAsToggle) {
```
Azmoria, line 1281: In order to allow scrolling on other elements when this setting is enabled this will have to be adjusted to be something along the lines of

```js
const sceneHovered = $('#VTT :hover, #black_layer:hover').length > 0 && !(e.target.tagName == 'INPUT' && e.target == document.activeElement)

if (e.ctrlKey || (wheelZoomAsToggle && sceneHovered))
```

Azmoria (changes requested):
With the below change or similar to make sure other elements in the VTT are still scrollable this could probably be merged but I haven't actually fully tested it yet.

It would also be nice if enabled we swapped hold ctrl to scroll the window up/down like shift scrolls side to side. Not required like the above to be functional ofc. Just would be a nice to have for when using the mouse for things like drawing walls/polygons etc. where right click to drag the map isn't really a good option.

PR #4826, opened 2026-09-02
Add option to import mask to create walls

### LINKED ISSUE #4825: Feature Request: Generate light/vision walls from a color-coded map mask
### Problem

Creating light/vision walls manually on large or irregular maps - or in the case of what's happening right now that made me think about this, a large AND irregular map - can require placing hundreds of wall points.  On the other hand it's relatively simple in most graphics programs to create a multi-pixel outline around an object and give it a specific color.

It would be extremely useful if AboveVTT could generate walls by reading a specified color from an imported image or mask.

### Proposed feature

Allow the DM to specify a color, preferably by hex value (for example #FF00FF), and have AboveVTT interpret pixels / objects of that color as light / vision walls.

Ideally this could work with a separate wall-mask image having the same dimensions as the scene image. That would allow the mask to contain ugly / high-contrast machine-readable geometry without modifying the map players actually see.

The workflow could be something like:

- Prepare the finished map normally.
- Create a second image/layer containing the desired walls in a distinctive color such as #FF00FF.
- Export that layer as a PNG with the same pixel dimensions as the map.
- In AboveVTT choose something like Import Wall Mask.
- Select/upload the mask and specify #FF00FF as the wall color.
- AboveVTT detects the colored boundaries, converts them into light/vision wall segments, and simplifies the resulting geometry to a reasonable number of wall points.
- The normal clean map remains the visible scene.

### Example use case

I am currently preparing a very large cave map with a transparent exterior. Manually tracing all of the irregular cave walls in AboveVTT is extremely time-consuming, and may well drive me mad :)

In Paint.NET, however, I can create essentially the entire wall mask in seconds:

- Select the transparent area.
- Invert the selection to select the playable cave area.
- Create a layer beneath it and flood-fill the selection with a unique color.
- Duplicate/offset that layer a few pixels up, down, left, and right to create a colored boundary around the cave system.
- Flatten/export that result as a wall-mask PNG.

I'm sure the process in a "grown-up" graphics package such as The Gimp or Photoshop would be even quicker.

The resulting image does not need to be attractive, it only needs to tell AboveVTT “pixels of this color represent walls.”

For maps without transparency, simply masking out the neutral background and adding the boundary to it on a new layer would take about the same amount of time.  Again, if I can do it in Paint.NET I'm sure it can be done using Photoshop.

### Possible implementation considerations

The generated walls probably should not create one wall vertex for every boundary pixel. Some amount of line/vector simplification or configurable tolerance would be needed to turn a raster boundary into a manageable number of wall segments.

A simple first implementation could potentially require an exact RGB/hex match. A later version could optionally support a small color tolerance for anti-aliased lines.

This would also make it possible for map creators to distribute optional wall-mask PNGs alongside their maps without requiring them to create a full Universal VTT file.

### Why this would help

For simple rectangular rooms, manual wall drawing is already quick. The major benefit would be large natural environments such as caves, coastlines, ruins, irregular buildings, forests, etc. where the visual wall boundary already exists but reproducing it manually as VTT wall geometry can take a very long time, and in my case often leaves tiny gaps.

Thanks for reading an (hopefully) not laughing :)

Azmoria:
This was suggested / requested in #4825. I think it seems pretty useful as image editting software really do have a lot of options for quickly masking out walls. This is something I might use myself but would also be nice for those creating map packages as an alternative to building out a full UVTT file as mentioned in the initial request.

Basic example:

walls mask2

### DIFF
--- a/CoreFunctions.js
+++ b/CoreFunctions.js
@@ -1049,10 +1049,24 @@ function apply_avtt_slash_command_button(sourceElement, targetElement){
   }
   $(targetElement).empty().append(updatedInnerHtml);
 }
+/* The character sheet snippets will format partial dice rolls due to snippet formulas. Eg <strong>1d8</strong>+5
+   This function unwraps those roll formulas to ensure proper dice roll injection. */
+function unwrap_roll_formulas(sheetElement){
+  const diceTerm = String.raw`\d+d\d+(?:(?:min\d+)|(?:ro(?:[<>=]{1,2})?\d+)|(?:k[hl]\d+))*`;
+  const rollFragment = new RegExp(String.raw`^\s*(?:${diceTerm}(?:\s*[+\-−]\s*(?:${diceTerm}|\d+))*|[+\-−]\s*\d+)\s*$`, 'i');
+  const inlineFormattingElements = sheetElement.querySelectorAll('strong, b, em, i, span, u, mark');
+
+  for(const element of inlineFormattingElements){
+    if(rollFragment.test(element.textContent)){
+      element.replaceWith(sheetElement.ownerDocument.createTextNode(element.textContent));
+    }
+  }
+}
 
 /** Injects roll/aoe buttons into a dnd sheet one editable text node at a time so the replacements can't rewrite the sheet's own markup. */
 function add_roll_buttons_to_text(sheetElement){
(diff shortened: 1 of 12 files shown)


firelightrpg:
This is amazing!

